PR #6, opened 2020-03-26
Remove dependency on PolygonLib and GPC

shr-project:
* the triangulation algoritm used by shapely is different
  from GPC, but if we remove the triangles which cover the
  holes, then both tests pass and it might work without
  any extra library needed.

Signed-off-by: Martin Jansa <martin.jansa@lge.com>

### DIFF
--- a/src/scenic/core/geometry.py
+++ b/src/scenic/core/geometry.py
@@ -8,7 +8,7 @@ import itertools
 import numpy as np
 import shapely.geometry
 import shapely.ops
-import Polygon as PolygonLib	# TODO remove dependency (see triangulatePolygon)
+import pypoly2tri
 
 from scenic.core.distributions import (needsSampling, distributionFunction,
                                        monotonicDistributionFunction)
@@ -172,20 +172,20 @@ def cleanChain(chain, tolerance, angleTolerance=0.008):
 	return newChain
 
 def triangulatePolygon(polygon):
-	# TODO replace with another library!
-	# N.B. can't use shapely.ops.triangulate since it doesn't respect edges
-	poly = PolygonLib.Polygon(polygon.exterior.coords)
-	for interior in polygon.interiors:
-		poly.addContour(interior.coords, True)
-	tristrips = poly.triStrip()
-	triangles = []
-	for strip in tristrips:
-		a, b = strip[:2]
-		for c in strip[2:]:
-			tri = shapely.geometry.Polygon((a, b, c))
-			triangles.append(tri)
-			a = b
-			b = c
+	polyline = []
+	for c in polygon.exterior.coords[:-1]:
+		polyline.append(pypoly2tri.shapes.Point(c[0],c[1]))
+	cdt = pypoly2tri.cdt.CDT(polyline)
+	polyline = []
+	for i in polygon.interiors:
+		for c in i.coords[:-1]:
+			polyline.append(pypoly2tri.shapes.Point(c[0],c[1]))
+		cdt.AddHole(polyline)
+	triangles = list()
+	cdt.Triangulate()
+
+	for t in cdt.GetTriangles():
+		triangles.append(shapely.geometry.Polygon([(t.GetPoint(0).toTuple()), (t.GetPoint(1).toTuple()), (t.GetPoint(2).toTuple())]))
 	return triangles
 
 def plotPolygon(polygon, plt, style='r-'):
--- a/src/scenic/core/geometry.py
+++ b/src/scenic/core/geometry.py
@@ -176,16 +176,20 @@ def triangulatePolygon(polygon):
 	for c in polygon.exterior.coords[:-1]:
 		polyline.append(pypoly2tri.shapes.Point(c[0],c[1]))
 	cdt = pypoly2tri.cdt.CDT(polyline)
-	polyline = []
 	for i in polygon.interiors:
+		polyline = []
 		for c in i.coords[:-1]:
 			polyline.append(pypoly2tri.shapes.Point(c[0],c[1]))
 		cdt.AddHole(polyline)
-	triangles = list()
 	cdt.Triangulate()
 
+	triangles = list()
 	for t in cdt.GetTriangles():
-		triangles.append(shapely.geometry.Polygon([(t.GetPoint(0).toTuple()), (t.GetPoint(1).toTuple()), (t.GetPoint(2).toTuple())]))
+		triangles.append(shapely.geometry.Polygon([
+			t.GetPoint(0).toTuple(),
+			t.GetPoint(1).toTuple(),
+			t.GetPoint(2).toTuple()
+		]))
 	return triangles
 
 def plotPolygon(polygon, plt, style='r-'):
--- a/README.md
+++ b/README.md
@@ -78,6 +78,4 @@ The models are:
 
 ## License
 
-Scenic is distributed under the 3-Clause BSD License. However, it currently uses the [GPC library](http://www.cs.man.ac.uk/~toby/alan/software/) (through the [Polygon3](https://pypi.org/project/Polygon3/) package), which is free only for non-commercial use.
-GPC is used only in the `triangulatePolygon` function in `scenic.core.geometry`, and you can alternatively plug in any algorithm of your choice for triangulation of polygons with holes.
-We plan to replace GPC with a BSD-compatible library in the near future.
+Scenic is distributed under the 3-Clause BSD License.
--- a/setup.py
+++ b/setup.py
@@ -15,7 +15,7 @@ setup(name='scenic',
           'opencv-python',
           'pillow',
           'shapely', # had to separately install via conda install shapely
-          'Polygon3',
+          'pypoly2tri',
       ],
       extras_require={
         'pyproj': ['pyproj'], # issue on Windows with Anaconda
--- a/tests/core/test_geometry.py
+++ b/tests/core/test_geometry.py
@@ -14,8 +14,8 @@ def checkTriangulation(poly, prec=1e-6):
     assert all(t.difference(poly).area == pytest.approx(0, abs=prec) for t in tris)
     # check triangles are (nearly) disjoint
     for i, t1 in enumerate(tris[:-1]):
-        t2 = tris[i+1]
-        assert (t1 & t2).area == pytest.approx(0, abs=prec)
+        for t2 in tris[i+1:]:
+            assert (t1 & t2).area == pytest.approx(0, abs=prec)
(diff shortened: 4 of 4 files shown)


shr-project:
The triangles are different as shown in:
https://github.com/shr-project/Scenic/blob/gpc-svg/triangulation-polygonlib-gpc-vatti.svg
https://github.com/shr-project/Scenic/blob/gpc-svg/triangulation-shapely-geos-delaunay.svg
but both test cases work and the results match with the description I got, please check if it is good enough for Scenic or let me know which cases don't work with this (or even better extend tests/core/test_geometry.py to cover cases like that as well)

dfremont:
Thanks for this, Martin! Your approach indeed works for the polygon I used in my test, but unfortunately doesn't work in general. I've added a new test using the following counterexample:

delaunay_cex

As you can see, one of the edges of the Delaunay point set triangulation (as computed by Shapely) passes through the red hole in the center, so you can't just remove some triangles to get a valid triangulation for the original polygon. If you were using a library for *constrained* Delaunay triangulation then you could exclude the hole, as we need. I hadn't really looked into this before since we don't care about the Delaunay property (and it's typically faster to find any old triangulation than a Delaunay one), but looking quickly, there seem to be existing Python libraries for constrained Delaunay triangulations. [Here's one](https://pypi.org/project/tri/) that uses an MIT license, although it's for Python 2.

(By the way, in the process of adding the new test I discovered that I'd forgotten to harden one of the triangulation assertions against roundoff error, so thanks again!)

shr-project:
Hi Daniel, thanks for information, I'll go back to testing these other libraries. Steve Lemke pointed me to poly2tri before, I've found few more versions of the same, here is what I've found:

The linked https://github.com/greenm01/poly2tri (last updated 7 years ago) isn't compatible with python3 (pip3 install poly2tri fails as well).

I've updated it to be compatible with python3 in https://github.com/shr-project/poly2tri/tree/py3 , but there seems to be better starting point in:
https://github.com/davidcarne/poly2tri.python (last updated 5 years ago)
which should be compatible with python3 already, only tests needed to be updated for Python 3.8:
https://github.com/shr-project/poly2tri.python/tree/py3

There is also
https://github.com/popupcad/pypoly2tri (last updated 11 months ago)
which is most recent, but the lack of any description doesn't help much.

I'll check how they work with the new test case and how to integrate them if they work correctly.

shr-project:
I've tested the poly2tri implementation from
https://github.com/shr-project/poly2tri.python/tree/py3
and it works better with the new test case, see:
https://github.com/shr-project/Scenic/blob/gpc-svg/triangulation-test3.svg
https://github.com/shr-project/Scenic/blob/gpc-svg/triangulation-test3-polygonlib-gpc-vatti.svg
https://github.com/shr-project/Scenic/blob/gpc-svg/triangulation-test3-shapely-geos-delaunay.svg
https://github.com/shr-project/Scenic/blob/gpc-svg/triangulation-test3-poly2tri.svg

But the precision of poly2tri isn't good enough for the current test case:
```
        # check union of triangles is (nearly) identical to poly
        union = shapely.ops.unary_union(tris)
>       assert poly.difference(union).area == pytest.approx(0)
E       assert 1.1920928910669204e-08 == 0 ± 1.0e-12
E        +  where 1.1920928910669204e-08 = <shapely.geometry.multipolygon.MultiPolygon object at 0x7f73312706d0>.area
E        +    where <shapely.geometry.multipolygon.MultiPolygon object at 0x7f73312706d0> = <bound method BaseGeometry.difference of <shapely.geometry.polygon.Polygon object at 0x7f7331270850>>(<shapely.geometry.polygon.Polygon object at 0x7f7331270820>)
E        +      where <bound method BaseGeometry.difference of <shapely.geometry.polygon.Polygon object at 0x7f7331270850>> = <shapely.geometry.polygon.Polygon object at 0x7f7331270850>.difference
E        +  and   0 ± 1.0e-12 = <function approx at 0x7f73d753ac10>(0)
E        +    where <function approx at 0x7f73d753ac10> = pytest.approx
```

If I increase the size of the polygon used in 3rd test case 10 times, then it passes the test, is this going to be an issue in typical use from Scenic, we can probably scale the polygons inside the triangulatePolygon call as needed. Should I update PR for poly2try for you to try or look at some other implementations first?

shr-project:
I see https://github.com/BerkeleyLearnVerify/Scenic/commit/237fdeb2be9ac2c25265c9f9aefba4d6b14bcf3d now, that was quick :)

dfremont:
We don't need that much precision - I just left the default `1.0e-12` precision since Shapely achieved it. I've changed the test to only require `1.0e-6` now: if poly2tri then passes the tests (without scaling), I can test it more thoroughly after you update the PR.

One question remaining is about the installation - I'd still like Scenic to be installable automatically using pip, but since you had to update poly2tri a bit, how would that work? Can we contact whoever is in charge of the PyPI package and try to get it updated? I suppose we could change _setup.py_ to refer directly to your GitHub repo, at least as a temporary measure.

shr-project:
Yes it passes with 1.0e-6 without scaling.

I've sent e-mail to the author and pypi maintainer, lets see if they reply. I've added you to Cc as well.

shr-project:
The undocumented pypoly2tri seems actually pretty good, it works with all 3 test cases without any precision issues and it's available OOTB in pypi.

The result is the same set of triangles as with poly3tri:
https://github.com/shr-project/Scenic/blob/gpc-svg/triangulation-test3-pypoly2tri.svg

I've updated this PR for pypoly2tri, let me know if it works for you now.

src/scenic/core/geometry.py
```diff
@@ -172,20 +172,20 @@ def cleanChain(chain, tolerance, angleTolerance=0.008):
...
-		for c in strip[2:]:
-			tri = shapely.geometry.Polygon((a, b, c))
-			triangles.append(tri)
-			a = b
-			b = c
+	polyline = []
+	for c in polygon.exterior.coords[:-1]:
+		polyline.append(pypoly2tri.shapes.Point(c[0],c[1]))
+	cdt = pypoly2tri.cdt.CDT(polyline)
+	polyline = []
```
dfremont: Needs to go inside loop

shr-project:
I've cherry-picked your fixes and new test case to the branch used in PR, let me know if you expected me to squash it instead.

shr-project:
With the new test from the shr-project/gpc branch from where I've cherry-picked your commit it fails for me:
```
============================================================================================================================== FAILURES ===============================================================================================================================
______________________________________________________________________________________________________________________ test_triangulation_holes _______________________________________________________________________________________________________________________

    def test_triangulation_holes():
        """An example with multiple holes."""
        p = shapely.geometry.Polygon(
            [(0,0), (0,3), (5,3), (5,0)],
            holes=[
                [(1,1), (1,2), (2,2), (2,1)],
                [(3,1), (3,2), (4,2), (4,1)],
            ]
        )
>       checkTriangulation(p)

tests/core/test_geometry.py:54: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
tests/core/test_geometry.py:9: in checkTriangulation
    tris = geometry.triangulatePolygon(poly)
/usr/local/lib/python3.8/dist-packages/scenic-1.0.0b2-py3.8.egg/scenic/core/geometry.py:188: in triangulatePolygon
    cdt.Triangulate()
/usr/local/lib/python3.8/dist-packages/pypoly2tri/cdt.py:25: in Triangulate
    self.sweep_.Triangulate(self.sweep_context_)
/usr/local/lib/python3.8/dist-packages/pypoly2tri/sweep.py:22: in Triangulate
    self.SweepPoints(tcx)
/usr/local/lib/python3.8/dist-packages/pypoly2tri/sweep.py:28: in SweepPoints
    node = self.PointEvent(tcx, point)
/usr/local/lib/python3.8/dist-packages/pypoly2tri/sweep.py:41: in PointEvent
    new_node = self.NewFrontTriangle(tcx, point, node)
/usr/local/lib/python3.8/dist-packages/pypoly2tri/sweep.py:101: in NewFrontTriangle
    triangle.MarkNeighbor(node.triangle)
/usr/local/lib/python3.8/dist-packages/pypoly2tri/shapes.py:131: in MarkNeighbor
    self.MarkNeighbor_t(*args)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

self = <pypoly2tri.shapes.Triangle object at 0x7fcb6132b8e0>, t = <pypoly2tri.shapes.Triangle object at 0x7fcb6132b820>

    def MarkNeighbor_t(self, t):
        a = set(self.points_)
        b = set(t.points_)
>       aa = list(a - b)[0]
E       IndexError: list index out of range

/usr/local/lib/python3.8/dist-packages/pypoly2tri/shapes.py:122: IndexError
================================================================================================================= 1 failed, 3 passed in 0.58 seconds ==================================================================================================================
```

Do you have some other changes or was it somehow working for you?

dfremont:
Strange, that's the error I was getting before the fix. Are you sure you're on the right branch? I just extensively tested your updated PR and everything works, so I was about to merge it.

shr-project:
I'm sorry, haven't had enough coffee today... I was on the right branch, but didn't reinstall Scenic after cherry-picking your change, was just rerunning pytest (against older revision of Scenic).

It works for me as well, sorry for noise.

For the list of contributors I don't mind including the @lge.com e-mail. Or you can just mention "LG Electronics, Inc." as this contribution was sponsored by LGE company.

dfremont:
Okay, thanks again!

PR #10, opened 2020-11-11
Switch to LGSVL Dreamview submodule

shalinmehtalgsvl:
Currently the submodule is available in this preview branch https://github.com/lgsvl/PythonAPI/tree/preview

The branch may be deleted after our full public release, but the submodule will still remain a part of LGSVL

### DIFF
--- a/src/scenic/simulators/lgsvl/actions.py
+++ b/src/scenic/simulators/lgsvl/actions.py
@@ -52,9 +52,9 @@ class SetDestinationAction(Action):
 	def applyTo(self, obj, sim):
 		if self.timer == 0:
 			z = sim.groundElevationAt(self.dest)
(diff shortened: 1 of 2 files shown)


hadiTab:
FYI, The submodule is now part of master.

dfremont:
Thanks for this! I'm finally getting around to looking into this, and it does run with LGSVL 2020.06 (after fixing a few issues on the Scenic side), but I haven't been able to test fully because with the new Borregas Ave map Scenic always creates cars off the road. Have you been able to use that map? Is it possible the OpenDRIVE coordinate system no longer lines up with the LGSVL one?

dfremont:
This is now obsolete: Scenic uses the new API.

PR #11, opened 2020-11-19
CARLA scenarios

joel-mb:
This PR is meant to improve different aspects of Scenic regarding its utilization with CARLA. The following changes have been made:

- Sync mode and actor destruction
- Increased **compatibility to CARLA 0.9.10.1**
- Implementation of the 10 **CARLA challenge scenarios**, as well as improvement of the already implemented ones
- Extended the **prop library** and changed them to lists to ensure determinism
- CarlaActors now have **elevation** and **physics** attributes, and vehicles **autopilot**
- Added some utility functions related to **traffic lights**. 
- Changes to pedestrian's **setWalkingDirection**
- Fixed bug at **SetHandBrakeAction**
- Added **missing utils import**

---

### Sync mode and actor destruction

Currently, while the CARLA simulation is set to sync mode at the CarlaSimulator class, this is overwritten when the map is reloaded, causing the simulation to go back to async mode. Therefore, this reloading has been removed.

This pointed to another issue related to the destruction of actors, which is non-existent. Before, the actors where automatically being destroyed by the reloading of the map, but now this is no longer the case (it is also important to note that the way it was done before was wrong, as actors such as sensors weren't being destroyed at all). The pipeline has been changed to accommodate destruction function at both the simulator and simulation. For example, in the case of CARLA, the simulation destruction handles all actor destructions whereas the simulator one resets the simulation back to async mode.

---

### Upgraded compatibility to CARLA 0.9.10

**set_velocity** calls have been replaced with **set_target_velocity** to accommodate the PythonAPI changes in CARLA 0.9.10. This is also the case for its angular counterparts

---

### CARLA challenge scenarios
Added and improved all CARLA challenge scenarios. A complete list of the scenarios can be found [here](https://leaderboard.carla.org/scenarios/)

---

### CARLA blueprints
Added more CARLA blueprints to the scenic library. A complete list of the CARLA blueprints can be found [here](https://carla.readthedocs.io/en/latest/bp_library/)
Props added: `Debris`, `VendingMachine`, `Chair`, `BusStop`, `Advertisement`, `Garbage`, `Container`, `Table`, `Barrier`, `PlantPot`, `Mailbox`, `Gnome`, `CreasedBox`, `Case`, `Box`, `Bench`, `Barrel`, `ATM`, `Kiosk`, `IronPlate`, `TrafficWarning`. 

Additionally, the library has been changed to lists, to ensure the determinism of the chosen blueprints

---

### Added *elevation* and *physics* and *autopilot*

Two new attributes have been added to the CarlaActor. On one hand, *physics* now lets the user spawn an actor with or without physics. On the other hand, the fixed spawning height of 0.5m has been replaced by the *elevation* attribute. As before, by default all vehicles are spawned at 0.5m above the road, and all props at 0m.

Additionally, added the *autopilot* attribute to *Vehicle*, that allows it activate CARLA's autopilot (Traffic Manager). An example of how to create some background activity for a simulation is available at *examples/carla/backgroundActivity.scenic*

---

### Traffic Light Actions

Some utility function have been added to CARLA's *model.scenic* file:
- freezeTrafficLights: freezes all traffic lights in the town
- unfreezeTrafficLights: unfreezes all traffic lights in the town
- withinDistanceToRedYellowTrafficLight: returns a boolean with wheather or not a vehicle is near a red / yellow traffic light
- withinDistanceToTrafficLight: returns a boolean with wheather or not a vehicle is near a traffic light
- getClosestTrafficLightStatus: returns the status of the closest traffic light to a vehicle
- setClosestTrafficLightStatus: changes the status of the closest traffic light to a vehicle to the desired one

Added a scenario example at *examples/carla/trafficLights.scenic*

---

### Walker walking direction action

Pedestrian's *setWalkingDirection* functions has been modifed in order to take a relative direction with respect to the walker, not an absolute one

---

### Fixed handbrake action bug
Corrected parameter name in the *SetHandBrakeAction*.

---

### Added missing utils import
Added missing import in *src/scenic/simulators/carla/model.scenic*

### DIFF
--- a/examples/carla/Carla_Challenge/carlaChallenge1.scenic
+++ b/examples/carla/Carla_Challenge/carlaChallenge1.scenic
@@ -35,5 +35,5 @@ debris1 = Debris following roadDirection for Range(10, 20)
 debris2 = Debris following roadDirection from debris1 for Range(5, 10)
 debris3 = Debris following roadDirection from debris2 for Range(5, 10)
 
-require (distance from ego to intersection) > 50
+require (distance to intersection) > 50
 terminate when (distance from debris3 to ego) > 10 and (distance to start) > 50
--- a/examples/carla/Carla_Challenge/carlaChallenge1.scenic
+++ b/examples/carla/Carla_Challenge/carlaChallenge1.scenic
@@ -8,7 +8,7 @@ its original lane.
 ## SET MAP AND MODEL (i.e. definitions of all referenceable vehicle types, road library, etc)
 param map = localPath('../../../tests/formats/opendrive/maps/CARLA/Town01.xodr')  # or other CARLA map that definitely works
 param carla_map = 'Town01'
-model scenic.domains.driving.model
+model scenic.simulators.carla.model
 
 ## CONSTANTS
 EGO_MODEL = "vehicle.lincoln.mkz2017"
--- a/examples/carla/Carla_Challenge/carlaChallenge1.scenic
+++ b/examples/carla/Carla_Challenge/carlaChallenge1.scenic
@@ -1,0 +1,39 @@
+""" Scenario Description
+Traffic Scenario 01.
+Control loss without previous action.
+The ego-vehicle loses control due to bad conditions on the road and it must recover, coming back to
+its original lane.
+"""
+
+## SET MAP AND MODEL (i.e. definitions of all referenceable vehicle types, road library, etc)
+param map = localPath('../../../tests/formats/opendrive/maps/CARLA/Town01.xodr')  # or other CARLA map that definitely works
+param carla_map = 'Town01'
+model scenic.domains.driving.model
+
+## CONSTANTS
+EGO_MODEL = "vehicle.lincoln.mkz2017"
+EGO_SPEED = 10
+
+## DEFINING BEHAVIORS
+# EGO BEHAVIOR: Follow lane, and brake after passing a threshold distance to the leading car
+behavior EgoBehavior(speed=10):
+    do FollowLaneBehavior(speed)
+
+## DEFINING SPATIAL RELATIONS
+# Please refer to scenic/domains/driving/roads.py how to access detailed road infrastructure
+# 'network' is the 'class Network' object in roads.py
+
+# make sure to put '*' to uniformly randomly select from all elements of the list, 'lanes'
+lane = Uniform(*network.lanes)
+
+start = OrientedPoint on lane.centerline
+ego = Car at start,
+    with blueprint EGO_MODEL,
+    with behavior EgoBehavior(EGO_SPEED)
+
+debris1 = Debris following roadDirection for Range(10, 20)
+debris2 = Debris following roadDirection from debris1 for Range(5, 10)
+debris3 = Debris following roadDirection from debris2 for Range(5, 10)
+
+require (distance from ego to intersection) > 50
+terminate when (distance from debris3 to ego) > 10 and (distance to start) > 50
--- a/examples/carla/Carla_Challenge/carlaChallenge10.scenic
+++ b/examples/carla/Carla_Challenge/carlaChallenge10.scenic
@@ -6,7 +6,7 @@ Ego-vehicle enters an unsignalized intersection and has to adapt to the flow of 
 #SET MAP AND MODEL (i.e. definitions of all referenceable vehicle types, road library, etc)
 param map = localPath('../../../tests/formats/opendrive/maps/CARLA/Town05.xodr')  # or other CARLA map that definitely works
 param carla_map = 'Town05'
-model scenic.simulators.carla.model #located in scenic/simulators/carla/model.scenic
+model scenic.simulators.carla.model
 
 #CONSTANTS
 EGO_DISTANCE_TO_INTERSECTION = Uniform(15, 20) * -1
--- a/examples/carla/Carla_Challenge/carlaChallenge10.scenic
+++ b/examples/carla/Carla_Challenge/carlaChallenge10.scenic
@@ -1,6 +1,6 @@
 """ Scenario Description
 Based on 2019 Carla Challenge Traffic Scenario 09.
-Ego-vehicle is performing a right turn at an intersection, yielding to crossing traffic.
+Ego-vehicle enters an unsignalized intersection and has to adapt to the flow of traffic
 """
 
 #SET MAP AND MODEL (i.e. definitions of all referenceable vehicle types, road library, etc)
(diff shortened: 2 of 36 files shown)


src/scenic/simulators/carla/model.scenic
```diff
@@ -121,7 +130,8 @@ class Pedestrian(Pedestrian, CarlaActor, Walks):
     blueprint: Uniform(*blueprints.walkerModels)
 
     def setWalkingDirection(self, heading):
-        direction = Vector(0, self.speed).rotatedBy(heading)
+        forward = self.carlaActor.get_transform().get_forward_vector()
```
dfremont, line 133: This doesn't look right - the transform of the CARLA actor is irrelevant to what direction the pedestrian should walk (which is being given by `heading` in Scenic's coordinate system), and you're ignoring `self.speed`. What was wrong with the old calculation?

src/scenic/simulators/carla/model.scenic
```diff
@@ -143,3 +154,146 @@ class Trash(Prop):
...
+
+def _getClosestLandmark(vehicle, type, distance=100):
+    if vehicle._intersection is not None:
+        return None
+
+    waypoint = simulation().map.get_waypoint(vehicle.carlaActor.get_transform().location)
+    landmarks = waypoint.get_landmarks_of_type(distance, type)
+
+    if landmarks:
+        return builtins.min(landmarks, key=lambda l: l.distance)
```
dfremont, line 265: This is a flat bug in Scenic - you should be able to just write `min`, but I forgot that it accepts keyword arguments. I've pushed a fix.

src/scenic/simulators/carla/simulator.py
```diff
@@ -41,22 +44,32 @@ def __init__(self, carla_map, address='127.0.0.1', port=2000, timeout=10,
...
 							   verbosity=verbosity)
 
+	def destroy(self):
+		settings = self.world.get_settings()
+		settings.synchronous_mode = False
+		settings.fixed_delta_seconds = None
+		self.world.apply_settings(settings)
+		self.tm.set_synchronous_mode(False)
+
+		super(CarlaSimulator, self).destroy()
```
dfremont, line 58: Any reason not to just use `super().destroy()`?

src/scenic/simulators/carla/simulator.py
```diff
@@ -82,22 +95,27 @@ def __init__(self, scene, client, map, timestep, render, record, verbosity=0):
 			print("blueprint: ", blueprint)
 
 			# Set up transform
-			loc = utils.scenicToCarlaLocation(obj.position, z=obj.elevation, world=self.world)
+			loc = utils.scenicToCarlaLocation(obj.position, world=self.world)
 			rot = utils.scenicToCarlaRotation(obj.heading)
 			transform = carla.Transform(loc, rot)
-			
+			transform.location.z += obj.elevation
```
dfremont, line 101: I would rather have this in `scenicToCarlaLocation` (with the elevation getting passed in as an optional argument as before), since the point of that function is to do coordinate conversion. See also my comment in `model.scenic` about the default elevation being `None` or not.

src/scenic/simulators/carla/simulator.py
```diff
@@ -169,3 +187,12 @@ def getProperties(self, obj, properties):
 			angularSpeed=utils.carlaToScenicAngularSpeed(currAngVel),
 		)
 		return values
+
+	def destroy(self):
+		for obj in self.objects:
+			if obj.carlaActor is not None:
+				obj.carlaActor.destroy()
+		if hasattr(self, "cameraManager"):
```
dfremont, line 195: `self.cameraManager` could be `None` if we get a spawn error while creating the ego object. I'd write `if self.render and self.cameraManager:` instead.

src/scenic/simulators/carla/simulator.py
```diff
@@ -169,3 +187,12 @@ def getProperties(self, obj, properties):
...
+			self.cameraManager.destroy_sensor()
+
+		super(CarlaSimulation, self).destroy()
```
dfremont, line 198: I'd just put `super().destroy()`.

src/scenic/simulators/carla/utils/utils.py
```diff
@@ -1,5 +1,6 @@
 import carla
 import math
+import string
```
dfremont, line 3: This doesn't seem to get used.

dfremont (changes requested):
Looks good overall, thank you! I've pointed out a few places where minor improvements could be made. I haven't been able to test it with CARLA yet since I was having some issues with my server, but I'll try again once you've updated the PR against the recent fixes I've merged into master (which hopefully solve the problems you mentioned on Discord - let me know if not).

examples/carla/Carla_Challenge/carlaChallenge10.scenic
```diff
@@ -0,0 +1,58 @@
...
+Ego-vehicle enters an unsignalized intersection and has to adapt to the flow of traffic
+"""
+
+#SET MAP AND MODEL (i.e. definitions of all referenceable vehicle types, road library, etc)
+param map = localPath('../../../tests/formats/opendrive/maps/CARLA/Town05.xodr')  # or other CARLA map that definitely works
+param carla_map = 'Town05'
+model scenic.simulators.carla.model
+
+#CONSTANTS
+EGO_DISTANCE_TO_INTERSECTION = Uniform(15, 20) * -1
```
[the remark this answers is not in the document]
glopezdiest: Indeed, Range is what we wanted

examples/carla/Carla_Challenge/carlaChallenge3_dynamic.scenic
```diff
@@ -0,0 +1,55 @@
...
+
+# EGO BEHAVIOR: Follow lane and brake when reaches threshold distance to obstacle
+behavior EgoBehavior(speed=10):
+    do FollowLaneBehavior(speed)
+
+behavior PedestrianBehavior(speed=3):
+    while (distance from self to ego) > THRESHOLD:
+        wait
+
+    take SetWalkingDirectionAction(0)
```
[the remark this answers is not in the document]
glopezdiest: Done

examples/carla/Carla_Challenge/carlaChallenge1.scenic
```diff
@@ -0,0 +1,39 @@
...
+start = OrientedPoint on lane.centerline
+ego = Car at start,
+    with blueprint EGO_MODEL,
+    with behavior EgoBehavior(EGO_SPEED)
+
+debris1 = Debris following roadDirection for Range(10, 20)
+debris2 = Debris following roadDirection from debris1 for Range(5, 10)
+debris3 = Debris following roadDirection from debris2 for Range(5, 10)
+
+require (distance from ego to intersection) > 50
```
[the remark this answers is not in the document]
glopezdiest: Removed the ego instances

examples/carla/Carla_Challenge/carlaChallenge5.scenic
```diff
@@ -1,62 +1,58 @@
...
+lead = Car following roadDirection from ego for Range(10, 25),
+    with behavior LeadingCarBehavior(LEAD_CAR_SPEED)
 
-#EXPLICIT HARD CONSTRAINTS
-require (distance from ego to intersection) > 10
-require (distance from cyclist to intersection) > 10
\ No newline at end of file
+require (distance from ego to intersection) > 50
+require (distance from lead to intersection) > 50
+require always (lead.laneSection._laneToLeft is not None) and (lead.laneSection.isForward == lead.laneSection._laneToLeft.isForward)
```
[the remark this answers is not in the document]
glopezdiest: Indeed, the *fasterLane* and *slowerLane* are definitely helpful.

src/scenic/__main__.py
```diff
@@ -137,35 +137,40 @@ def runSimulation(scene):
...
+        while len(its) < args.gather_stats:
+            scene, iterations = generateScene()
+            its.append(iterations)
+        totalTime = time.time() - startTime
+        count = len(its)
+        print(f'Sampled {len(its)} scenes in {totalTime:.2f} seconds.')
+        print(f'Average iterations/scene: {sum(its)/count}')
+        print(f'Average time/scene: {totalTime/count:.2f} seconds.')
+
+except KeyboardInterrupt:
```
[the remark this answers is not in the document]
glopezdiest: Yes, it is indeed better to add a finally (we didn't think about the successCount). However, we think that it is still good to catch the KeyboardInterrupt, as Scenic already detects it with the different try - finally, but never catches it

src/scenic/simulators/carla/simulator.py
glopezdiest: None, updated

src/scenic/simulators/carla/simulator.py
glopezdiest: Check elevation comment

src/scenic/simulators/carla/simulator.py
glopezdiest: Done

src/scenic/simulators/carla/simulator.py
glopezdiest: Done

src/scenic/simulators/carla/utils/utils.py
glopezdiest: True, removed

ek65:
I tested this pull request with multiple carla scenarios on my linux server with a Carla Simulator (version 0.9.10).
All the scenario ran fine. However, the simulation is running roughly at 40Hz. When I tested scenarios with time limit of 
200 simulation time steps, it took about 5 seconds on average until termination per simulation run. Any idea why this may be?
I see that in scenic/simulators/carla/simulator.py in line 24 the default simulation timestep is set to 0.1 second (10Hz).

src/scenic/simulators/carla/model.scenic
```diff
@@ -71,10 +76,12 @@ class CarlaActor(DrivingObject):
         self.carlaActor.set_location(utils.scenicToCarlaLocation(pos, elevation))
 
     def setVelocity(self, vel):
-        self.carlaActor.set_velocity(utils.scenicToCarlaVector3D(*vel))
+        self.carlaActor.set_target_velocity(utils.scenicToCarlaVector3D(*vel))
```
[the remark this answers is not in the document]
glopezdiest: It does apply an instantaneous velocity to the actor, but it will never be exactly that value, as it is set before the forces that affect the actor are calculated, which end up reducing its final value by a small amount.

src/scenic/core/simulators.py
```diff
@@ -242,6 +251,9 @@ def currentState(self):
         """
         return tuple(obj.position for obj in self.objects)
 
+    def destroy(self):
```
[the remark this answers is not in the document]
glopezdiest: Done

src/scenic/simulators/carla/model.scenic
```diff
@@ -90,6 +97,8 @@ class Vehicle(Vehicle, CarlaActor, Steers):
     def setReverse(self, reverse):
         self.control.reverse = reverse
 
+    def _getClosestTrafficLight(self, distance=100):
```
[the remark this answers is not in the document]
glopezdiest: No, the underscore is explicitly there to convey that that function shouldn't be used directly from a scenario because it returns a CARLA actor. Regarding the distance argument, we are using CARLA's *get_landmarks_of_type(self, distance, type, stop_at_junction=False)*, which checks all the landmarks within that distance. Therefore, the distance cannot be infinite, or it will never stop searching.

glopezdiest:
Hey @ek65 .

1) When running asynchronously, the simulation will always run at real time, and therefore you'll always see that it goes at 10Hz. However, this is not the case for synchronous mode. From your numbers, it seems like the simulation is running at 4x real time, which turns those 10Hz to 40Hz. It is important that the simulation still runs at 10Hz, but due to it not being in real time, it seems different.

2) We'll take a look.

3) Yes, the PID are still not working properly. We believe that the lateral controller should be changed from its *distance to lane center*, into other approach that aims to a waypoint in the front, but we haven't implemented that yet

Also, I definitely agree about being able to detect stops and your traffic light suggestions. They are in our to-do list, but yet again, aren't implemented yet.

glopezdiest:
Regarding the elevation, we have gone back on our changes. *elevation* is no longer changed, and the `snapToGround` has been modified, removing the *+0.5* for props. Also, all props are now spawned without physics, as this is also CARLA's default behavior

dfremont (approved):

PR #13, opened 2020-12-15
Fixed a couple utility functions

glopezdiest:
This PR means to fix two utility functions:
- **withinDistanceToObjsInLane:** Outside intersections, `network.intersectionAt(vehicle)` always returns None, which was causing some objects to be taken into account even if they weren't inside the specified lane
- **Maneuver.conflictingManeuver:** This function used the `PoligonalRegions.intersects()` function, creating some false positives between parallel lanes. Therefore, a new `Lane.intersects()` function has been created to overwrite it, which specifically uses their centerLines, not the whole road.

### DIFF
--- a/src/scenic/domains/driving/model.scenic
+++ b/src/scenic/domains/driving/model.scenic
@@ -358,11 +358,11 @@ def withinDistanceToObjsInLane(vehicle, thresholdDistance):
     for obj in objects:
         if not (vehicle can see obj):
             continue
-        if not (network.laneAt(vehicle) == network.laneAt(obj) or network.intersectionAt(vehicle)==network.intersectionAt(obj)):
-            continue
         if (distance from vehicle.position to obj.position) < 0.1:
             # this means obj==vehicle
-            pass
-        elif (distance from vehicle.position to obj.position) < thresholdDistance:
-            return True
+            continue
+        if (network.intersectionAt(vehicle) is not None and network.intersectionAt(vehicle)==network.intersectionAt(obj)) or \
+            network.intersectionAt(vehicle) is None and network.laneAt(vehicle) == network.laneAt(obj):
+            if (distance from vehicle.position to obj.position) < thresholdDistance:
+                return True
     return False
--- a/src/scenic/domains/driving/model.scenic
+++ b/src/scenic/domains/driving/model.scenic
@@ -361,8 +361,11 @@ def withinDistanceToObjsInLane(vehicle, thresholdDistance):
         if (distance from vehicle.position to obj.position) < 0.1:
             # this means obj==vehicle
             continue
-        if (network.intersectionAt(vehicle) is not None and network.intersectionAt(vehicle)==network.intersectionAt(obj)) or \
-            network.intersectionAt(vehicle) is None and network.laneAt(vehicle) == network.laneAt(obj):
-            if (distance from vehicle.position to obj.position) < thresholdDistance:
-                return True
+        inter = network.intersectionAt(vehicle)
+        if inter and inter != network.intersectionAt(obj):    # different intersections
+            continue
+        if not inter and network.laneAt(vehicle) != network.laneAt(obj):    # different lanes
+            continue
+        if (distance from vehicle to obj) < thresholdDistance:
+            return True
     return False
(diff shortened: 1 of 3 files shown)


src/scenic/domains/driving/model.scenic
```diff
@@ -358,11 +358,11 @@ def withinDistanceToObjsInLane(vehicle, thresholdDistance):
...
         if (distance from vehicle.position to obj.position) < 0.1:
             # this means obj==vehicle
-            pass
-        elif (distance from vehicle.position to obj.position) < thresholdDistance:
-            return True
+            continue
+        if (network.intersectionAt(vehicle) is not None and network.intersectionAt(vehicle)==network.intersectionAt(obj)) or \
+            network.intersectionAt(vehicle) is None and network.laneAt(vehicle) == network.laneAt(obj):
+            if (distance from vehicle.position to obj.position) < thresholdDistance:
+                return True
```
dfremont, line 367: I personally find this harder to read than the old linear style. How about the following:
```
inter = network.intersectionAt(vehicle)
if inter and inter != network.intersectionAt(obj):    # different intersections
    continue
if not inter and network.laneAt(vehicle) != network.laneAt(obj):    # different lanes
    continue
if (distance from vehicle to obj) < thresholdDistance:
    return True
```

src/scenic/domains/driving/roads.py
```diff
@@ -478,6 +478,11 @@ def sectionAt(self, point: Vectorlike, reject=False) -> Union[LaneSection, None]
         """Get the LaneSection passing through a given point."""
         return self.network.findPointIn(point, self.sections, reject)
 
+    def intersects(self, other):
+        if isinstance(other, Lane):
```
dfremont, line 482: I wouldn't override the `intersects` method like this, since `Lane` is an instance of `PolygonalRegion` and one would not expect its `intersects` method to behave differently from other polygons. Instead, I would change the test in `Maneuvers.conflictingManeuvers` from `maneuver.connectingLane.intersects(guideway)` to `maneuver.connectingLane.centerline.intersects(guideway.centerline)`. This is much more explicit.

More broadly, I'm wondering if this change is sound. Consider a car turning left into a lane while another car turns right into the same lane. These should be considered conflicting maneuvers, and are indeed so with our current code, but since their centerlines may only intersect at a single point (where they meet at the beginning of the destination lane), your proposed code might fail to call them conflicting maneuvers due to round-off error. Perhaps it would be better to test whether the distance between the centerlines is below some small epsilon? I'm not even sure whether that would be sound - is it really always the case that maneuvers can conflict only if their centerlines do? What if someone making a U-turn for example clipped a lane due to the nonzero width of the car without the centerline of the maneuver actually crossing the centerline of the lane?

In any case, your original point about false positives from parallel lanes is certainly a good one. Perhaps we could consider an alternative approach that would directly eliminate such cases, like contracting the polygons slightly? But if you think the nearby-centerlines heuristic is good enough we could just go with that for now (we can always revise if we hear complaints from people with other maps).

dfremont (changes requested):
See comments below.

src/scenic/domains/driving/model.scenic
glopezdiest: Agreed, done

src/scenic/domains/driving/roads.py
glopezdiest: Okay, changed the *conflictingManeuver*, instead of overwritting the `Lane.intersects`.

Regarding your other issues, it is true that I didn't think about those cases. From a general point of view, using the whole lane results in false positives while using the centerlines creates false negatives. I agree that the best approach would be either of the two you mentioned. Currently, from what we saw at the CARLA maps, I think that using the centerlines is better than using the whole lane, so I'd suggest changing it, and we'll create this more complex checks later on.

src/scenic/domains/driving/model.scenic
glopezdiest: Done

src/scenic/domains/driving/roads.py
dfremont: OK, let's go with this for now, and I'll put in a distance threshold after merging.

PR #14, opened 2020-12-15
Added global parameter parsing for CARLA simulator

glopezdiest:
This PR adds several parameters to the Carla Simulator:
- **address:** to specify the client address
- **port:** to specify the client port
- **timeout:** to specify the client timeout
- **render:** to specify whether or not Scenic creates and shows the camera. Its values are '0' and '1', as all parameters created from the command line are strings
- **record:** removed the previous utility (saving the camera images to disc). It now calls the CARLA recorder and saves all the scenarios inside a folder which has the name of the parameter
- **timestep:** to set the simulation timestep
- **weather:** to change the weather of the simulation. It accepts both a string with a weather preset (HardRainNoon, WetSunset...) or, if inside an scenario, a dictionary specifying each of its values (sun, fog, rain...) . This is not possible from teh command because, again, all parameters created from the command line are strings.


src/scenic/simulators/carla/simulator.py
```diff
@@ -20,16 +21,28 @@
...
 		super().__init__()
 		verbosePrint('Connecting to CARLA...')
 		self.client = carla.Client(address, port)
 		self.client.set_timeout(timeout)  # limits networking operations (seconds)
-		self.world = self.client.load_world(carla_map)
-		self.map = carla_map
+		if carla_map is not None:
+			self.world = self.client.load_world(carla_map)
+		else:
+			with open(map_path) as odr_file:
```
dfremont, line 33: This assumes that the `map` global parameter refers to an OpenDRIVE file, which isn't necessarily the case (we also support Lanelet2 maps, for example). If it isn't, you can raise an exception saying the CARLA interface only supports OpenDRIVE maps.

src/scenic/simulators/carla/model.scenic
```diff
@@ -39,18 +39,36 @@ except ModuleNotFoundError:
...
-precipitation = Options({0: 70, 1: 30}) * Range(0, 100)
-param precipitation = precipitation
-param precipitation_deposits = Range(precipitation, 100)
-param cloudiness = Range(precipitation, 100)
-param wind_intensity = Range(0, 100)
-param sun_azimuth_angle = Range(0, 360)
-param sun_altitude_angle = Range(-90, 90)
-
+    param carla_map = None
+if 'address' not in globalParameters:
```
dfremont, line 43: You don't need any of these `if` statements. Just write `param carla_map = None`, etc.: the command-line option to set global parameters overrides all `param` statements.

src/scenic/simulators/carla/model.scenic
```diff
@@ -39,18 +39,36 @@ except ModuleNotFoundError:
...
+    param address = '127.0.0.1'
+if 'port' not in globalParameters:
+    param port = 2000
+if 'timeout' not in globalParameters:
+    param timeout = 10
+if 'render' not in globalParameters:
+    param render = '1'
```
dfremont, line 50: Having everything be strings is indeed awkward. I could change the `--param` option to first try interpreting the given string as a number and then fall back on strings - what do you think?

src/scenic/simulators/carla/simulator.py
```diff
@@ -172,3 +194,18 @@ def getProperties(self, obj, properties):
 			angularSpeed=utils.carlaToScenicAngularSpeed(currAngVel),
 		)
 		return values
+
+	def destroy(self):
+		for obj in self.objects:
```
dfremont, line 199: This also looks like it got pulled in from #11; we may need to wait until that is merged before fixing this one (I'm not sure how to do dependencies between PRs on GitHub).

src/scenic/simulators/carla/model.scenic
```diff
@@ -39,18 +39,36 @@ except ModuleNotFoundError:
...
+else:
+    if globalParameters.render not in ['0', '1']:
+        raise ValueError('render param must be either 0 or 1')
+if 'record' not in globalParameters:
+    param record = ''
+if 'timestep' not in globalParameters:
+    param timestep = 0.1
```
dfremont, line 57: We used `time_step` for LGSVL, but I'm thinking maybe `timestep` is better. I can change the LGSVL interface for consistency.

dfremont (changes requested):
Looks good - just a few minor comments below.

@ek65 This PR replaces the recording mechanism you were using previously. Can you please double-check that the new system will work for you?

ek65:
Question:
1. Once I record data using your 'record' feature, how can I visualize it?
2. Is it possible to assign a distribution over the weather condition such that if users are running multiple simulation with one scenic script, then various weather conditions are rendered.

Suggestion:
I think it would be better to assign uniform distribution over all possible weather conditions that Carla offers in model.scenic script. As an example, please refer to how weather parameter is defined in scenic/simulators/gta/model.scenic. If desired, users can override this uniform distribution on weather in the scenic script, by stating : param weather = 'SUNNY'

This allows users to be as general as possible without having to specify all weather conditions, and also let them be more stringent by overriding the prior distribution with a more specific condition.

src/scenic/simulators/carla/model.scenic
glopezdiest: If those conditionals aren't there, the parameters defined at the scenario file might be overwritten. Is this the intended behavior or should the parameters at the scenario file have more priority?

src/scenic/simulators/carla/simulator.py
glopezdiest: Indeed, this PR adds the `client.stop_recorder`, which is why the whole destroy function was brought

glopezdiest:
@ek65 

1) Using the carla recorder, a *.log* file is created with all the information about the simulation. In order to visualize it, you can use `client.replay_file(filename)` (client API [here] (https://carla.readthedocs.io/en/latest/python_api/#carlaclient).

2) We've changed how the weather is parsed, and now each scenario inside a simulation has a different one. Please take a look and see if that meets your suggestion

src/scenic/simulators/carla/model.scenic
dfremont: They shouldn't be overwritten if you define them in the scenario after the `model` statement. But I can see how this could be confusing, since the `map` and `carla_map` parameters need to be defined *before* the `model` statement, and it would be natural to have them all in one place. So maybe I should change the semantics a bit and have `param` statements at the top level always take precedence. In the mean time let's leave your code the same - thanks for the explanation.

dfremont:
Hmm, I tested the ability to just set `map` and not `carla_map`, but got an error: if I set `map` to "tests/formats/opendrive/maps/CARLA/Town01.xodr", I get the exception "RuntimeError: opendrive could not be correctly parsed".

dfremont:
OK, the CARLA folks can't reproduce the error I'm getting, so we're going to merge this for now and revisit the problem later.

dfremont (approved):

PR #15, opened 2020-12-15
Fix bug upon CARLA's SimulationCreationError

glopezdiest:
Upon a SimulationCreationError, Scenic reset but didn't destroy the already created actors, which caused the simulation to possibly have remnants of previously tried ones. **This PR requries the Simulation.destroy() function, with is introduced in PR #11**


dfremont (commented):
Looks good - will merge after #11. I'm a little worried that additional cases like this may get missed, or that we might accidentally destroy a simulation twice, so I'll think about whether there's a safer way to do this (maybe we can make `Simulation` a context manager).

PR #16, opened 2020-12-15
Added xodr signals parsing

glopezdiest:
Updated the CARLA xodr maps to include the traffic lights and parsed them into the network. Added **isSignalized** to `Intersection` class


src/scenic/domains/driving/roads.py
```diff
@@ -352,6 +352,8 @@ class Road(LinearElement):
     laneGroups: Tuple[LaneGroup] = None
     sections: Tuple[RoadSection]    # sections in order from start to end
 
+    signals: Tuple[Signal]
```
dfremont, line 355: Where are these signals? Are they all signals encountered at either end of the road, or can they be within the road itself? Our plan had been to only have signals at intersections, but let me know if that is a bad assumption. If the signals are only encountered at the end of the road, then this would be better as an element of `Lane`, like `Lane.maneuvers`. In fact if there can be multiple signals for a given lane, e.g. if a left-turn arrow is represented separately from a green light, it would be good for each `Maneuver` to reference the signal which governs it.

src/scenic/domains/driving/roads.py
```diff
@@ -645,6 +647,8 @@ class Intersection(NetworkElement):
     outgoingLanes: Tuple[Lane]
     maneuvers: Tuple[Maneuver]  # all possible maneuvers through the intersection
 
+    signals: Tuple[Signal]
```
dfremont, line 650: Are these in any particular order? My plan had been for these to align with `Intersection.incomingLanes` (which the corresponding entry for a lane being `None` if there is no signal for it), but I can also see an argument for simply having a list of all signals. If we go with the latter, then we should also allow getting the signal(s) corresponding to a given incoming lane and the signal corresponding to a given maneuver.

dfremont (changes requested):
Many thanks for adding this functionality! This looks great, except it's not clear to me how one would actually get the signal (if any) corresponding to a particular lane or maneuver - see specific comments below.

src/scenic/domains/driving/roads.py
joel-mb: OpenDrive allows you to define signals anywhere in the map (i.e., inside intersections or not). Signals can also be defined in the middle of the road. There aren't restrictions in this sense. However, for the specific case of CARLA maps, traffic lights are always defined inside intersections.

As you say, right now, signals can only be retrieved from roads and intersections. Our plan is to have signals accessible from lanes and maneuvers also, but I think this would require some more work (e.g., check the lane validity of the signal and decide what to do with lane sections). Were were planning to add this functionality for a future PR.

src/scenic/domains/driving/roads.py
joel-mb: No. Right now they aren't in any particular order. I agree that having them align with `Intersection.incomingLanes` could be very useful. I think it could be reasonable to have both approached implemented. As I've said above, we could introduce this features in a future PR.

dfremont:
@joel-mb OK, thanks for the clarifications. I think we can merge this in as is, but since it is a work in progress and the API may change later, I may put a comment to that effect in `roads.py` to prevent people from assuming the API is stable.

dfremont:
Uh oh, Town03, Town06, and Town07 are now breaking the OpenDRIVE parser (you can see this if you run the test suite with `pytest`). I will investigate later today, after I've merged the other PRs.

joel-mb:
> Uh oh, Town03, Town06, and Town07 are now breaking the OpenDRIVE parser (you can see this if you run the test suite with `pytest`). I will investigate later today, after I've merged the other PRs.

@dfremont I've been investigating the issue. As far as I've seen, it seems that the problem is not related to the signals parsing but to the update of the opendrive files. Specifically, for Town03, I've seen that there is a road with a very short section (the second section of the road `575` has a lenght of the order of `10e-7`).
I've been trying to solve the issue without success so far. Let me know if you find any solution. We'll keep trying tomorrow.

dfremont:
@joel-mb Yes, it's the road geometry, not the signals. It's taken me a while to fix this, but it revealed several bugs in the OpenDRIVE parser, so that's good. I had previously put in some heuristics to try to handle tiny roads and road sections like the one you mention (it's rather annoying that RoadRunner seems to frequently generate those!), but they were breaking in this case. I now have all the maps except Town06 working (will push my changes soon), and in fact that map didn't work even before the update: the more complicated configuration of shoulder lanes is causing a problem. I'll look into it a little further, but a fix may have to wait until after your release; hopefully this isn't too urgent since none of the example scenarios use Town06.

dfremont:
@joel-mb OK, parsing of all CARLA maps is now working! The map schematics look OK, but I haven't had time to test in CARLA and need to sleep now, so I've pushed the fixes to a new `opendrive_fixes` branch. If you could merge that into your branch and give it a try, that would be great. If everything looks good I can merge tomorrow morning.

joel-mb:
@dfremont I've merged your changes and all CARLA maps are loading well now! I just had to include a minor modification to the signal parsing to avoid overwriting an already used variable. I've also added a comment to inform the users that the Signals API may change in the future. Please, feel free to modify whatever you think is necessary.

dfremont (approved):

PR #17, opened 2020-12-15
Improved PID values

glopezdiest:
Moved the **hardcoded PID values into functions** (_setLaneFollowingPIDControllers_ was already there, added _setTurnPIDControllers_, _setLaneChangingPIDControllers_). Their **values have also been updated** and some vehicle blueprint have been uncommented as their PID values are now much better. It is important to note that PR #11 has changes regarding the synchronous mode which makes the PID values perform much much better.

Additonally, moved `RegulatedControlAction(throttle, current_steer_angle, past_steer_angle)` at `FollowLaneBehavior` to several lines before, as it was being done after saving the values of that frame into its past variables

### DIFF
--- a/src/scenic/domains/driving/behaviors.scenic
+++ b/src/scenic/domains/driving/behaviors.scenic
@@ -26,6 +26,28 @@ def setLaneFollowingPIDControllers(is_vehicle, dt):
 
     return lon_controller, lat_controller
 
+def setTurnPIDControllers(is_vehicle, dt):
+    if is_vehicle:
+        lon_controller = controllers.PIDLongitudinalController(K_P=0.5, K_D=0.1, K_I=0.7, dt=dt)
+        lat_controller = controllers.PIDLateralController(K_P=0.8, K_D=0.2, K_I=0, dt=dt)
+
+    else:
+        lon_controller = controllers.PIDLongitudinalController(K_P=0.25, K_D=0.025, K_I=0.0, dt=dt)
+        lat_controller = controllers.PIDLateralController(K_P=0.4, K_D=0.1, K_I=0.0, dt=dt)
+
+    return lon_controller, lat_controller
+
+def setLaneChangingPIDControllers(is_vehicle, dt):
+    if is_vehicle:
+        lon_controller = controllers.PIDLongitudinalController(K_P=0.5, K_D=0.1, K_I=0.7, dt=dt)
+        lat_controller = controllers.PIDLateralController(K_P=0.08, K_D=0.3, K_I=0, dt=dt)
+
+    else:
+        lon_controller = controllers.PIDLongitudinalController(K_P=0.25, K_D=0.025, K_I=0.0, dt=dt)
+        lat_controller = controllers.PIDLateralController(K_P=0.1, K_D=0.3, K_I=0.0, dt=dt)
+
+    return lon_controller, lat_controller
+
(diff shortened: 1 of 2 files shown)


src/scenic/simulators/carla/blueprints.scenic
```diff
@@ -7,15 +7,15 @@ carModels = {# for speed = 10
       'vehicle.audi.etron', # K_P=0.3, K_D=0.2, K_I=0
       'vehicle.audi.tt',
       'vehicle.bmw.grandtourer',
-      # 'vehicle.bmw.isetta', # current PID controller not work 
+      'vehicle.bmw.isetta', # current PID controller not work
```
dfremont, line 10: Probably should remove the comment if it's not accurate anymore. Ditto for "wrangler_rubicon" below.

src/scenic/domains/driving/behaviors.scenic
```diff
@@ -26,6 +26,28 @@ def setLaneFollowingPIDControllers(is_vehicle, dt):
 
     return lon_controller, lat_controller
 
+def setTurnPIDControllers(is_vehicle, dt):
```
dfremont, line 29: These new functions don't seem to get used. Is part of the PR missing?

dfremont (changes requested):
Thanks for these improvements and fixes! It looks like the new PID functions didn't get hooked up properly though (maybe a commit is missing?) - see comments below.

ek65:
The same controller issue that I aforementioned in "CARLA scenarios #11" PR persists here.

src/scenic/domains/driving/behaviors.scenic
glopezdiest: Indeed, there was one commit missing...

src/scenic/simulators/carla/blueprints.scenic
glopezdiest: Done

glopezdiest:
Added a missing commit. The PIDs are still flawed, and will overshoot at closed turns, but this should fix the majority of the cases. I think that in order to fix it, the lateral controller should be changed from checking the distance to the center line, to tracking a waypoint in front.

src/scenic/simulators/carla/blueprints.scenic
dfremont: Comments are still there - maybe another commit is missing?

src/scenic/domains/driving/behaviors.scenic
dfremont: Oops, now there are two copies of `setTurnPIDControllers` and `setLaneChangingPIDControllers`, and they have different numerical parameters so I'm not sure which one is correct. Can you please fix?

dfremont (changes requested):
Nearly there, but a couple functions got duplicated and I'm not sure how to fix it - please see the comments.

glopezdiest:
Resolved the conflicts and removed the comments (now for sure)

dfremont:
Hmm, I'm still seeing two copies of `setTurnPIDControllers` and `setLaneChangingPIDControllers`. I also can't see my comment about them, so maybe I submitted the review wrong.

glopezdiest:
Indeed, removed duplicates

dfremont (approved):

PR #18, opened 2020-12-15
Added compatibility with CARLA's manual control

glopezdiest:
This changes are meant to add the capability to **run CARLA's manual control with Scenic scenarios**. Added several new scenarios that use this manual control, as well as the manual control itself.

This PR adds the **rolename attribute**, which is used by the manual control to detect the ego vehicle. This isn't done through a behavior as all attributes are fixed once the CARLA actor is created.


examples/carla/manual_control.py
```diff
@@ -0,0 +1,595 @@
+#!/usr/bin/env python
+
+# Copyright (c) 2020 Computer Vision Center (CVC) at the Universitat Autonoma de
+# Barcelona (UAB).
+#
+# This work is licensed under the terms of the MIT license.
+# For a copy, see <https://opensource.org/licenses/MIT>.
+
+# Allows controlling a vehicle with a keyboard. For a simpler and more
+# documented example, please take a look at tutorial.py.
```
dfremont, line 10: Presumably this is referring to a file in CARLA? Would be good to add a full link so as not to confuse people.

examples/carla/manual_control/carlaChallenge7.scenic
```diff
@@ -0,0 +1,74 @@
+""" Scenario Description
+Based on 2019 Carla Challenge Traffic Scenario 07.
+Ego-vehicle is going straight at an intersection but a crossing vehicle 
+runs a red light, forcing the ego-vehicle to perform a collision avoidance maneuver.
+Note: The traffic light control is not implemented yet, but it will soon be.
```
dfremont, line 5: This note seems to be outdated.

src/scenic/simulators/carla/model.scenic
```diff
@@ -54,6 +54,7 @@ param sun_altitude_angle = Range(-90, 90)
 
 class CarlaActor(DrivingObject):
     carlaActor: None
+    rolename: None
```
dfremont, line 57: Would be good to put a comment here explaining how this property is used.

src/scenic/simulators/carla/simulator.py
```diff
@@ -172,3 +174,13 @@ def getProperties(self, obj, properties):
 			angularSpeed=utils.carlaToScenicAngularSpeed(currAngVel),
 		)
 		return values
+
+	def destroy(self):
```
dfremont, line 178: Looks like this got accidentally pulled in from the other PR.

dfremont (changes requested):
Looks good; just a few tiny points below.

One broader point you could consider for a future PR: if the "manual_control" versions of the scenarios are the same as the normal versions except for the ego's behavior, maybe it would be good to have universal versions which could either use a built-in behavior or manual control based on a command-line argument. You could arrange this by using a global parameter to select the rolename and behavior; global parameters can be overridden on the command line using the `--param` [option](https://scenic-lang.readthedocs.io/en/latest/options.html#cmdoption-p).

ek65:
When I merged your manual_control branch with your master branch, there was a merge conflict in simulator.py. Please check and resolve this difference. I resolved the conflict by using manual_control's simulator.py. 

However, I was not able to manually control the car with my keyboard. None of the keys caused any movement from the car.

examples/carla/manual_control.py
glopezdiest: Removed

examples/carla/manual_control/carlaChallenge7.scenic
glopezdiest: Removed

src/scenic/simulators/carla/model.scenic
glopezdiest: Added a comment

src/scenic/simulators/carla/simulator.py
glopezdiest: Indeed, a `self.world.tick()` was added to that function, so it got pulled. It'd suggest waiting for PR #11 to be merged

dfremont:
I can fix the conflict, but the manual control also didn't work for me: when I run `manual_control.py`, it says "Ego vehicle found", but a new CARLA window opens with an ego car that isn't the same as the one Scenic created. Maybe I'm using the script wrong?

glopezdiest:
Resolved conflicts.

> Maybe I'm using the script wrong?

Just to be sure, you should be running a Scenic scenario like usual and then on another terminal, activate the manual_control.py at *examples/carla/manual_control.py* (this creates a second window, which is where disabling the Scenic rendering is useful). This one just waits for a vehicle with the 'hero' role_name to be spawned and attaches to it. If you are getting another vehicle it might be because a previous attempt failed to create the simulation, but the ego wasn't correctly destroyed. Is this what you are doing?

dfremont:
@glopezdiest Thanks for the clarification - it seems to be working now.

dfremont (approved):

PR #19, opened 2020-12-18
Scenario improvements

glopezdiest:
This PR adds new behaviors and improved the 10 CARLA challenge scenarios, as well as the manual control scenarios, the traffic light and background activity ones. Therefore, it needs PRs (#18, #16, #13, #11) to be merged in order to work. As they are also meant to be many conflicts, it'll start as a draft PR. 

The new behaviors are:
- **WalkBehavior:** uses CARLA AI controller to randomly move a pedestrian around the town
- **CrossingBehavior:** dynamically controls the speed of a crossing object so that it and a reference actor arrive at a spot in the road at the same time

Additionally, fixed bug at `Range.__contains__` class, which had no *self* when calling attribute *low* and *high*


glopezdiest:
Closing this one, will wait until the rest are merged before reopening it

PR #20, opened 2020-12-21
CARLA scenarios improved

joel-mb:
This PR adds new behaviors and improves the 10 CARLA challenge scenarios, as well as the manual control scenarios, the traffic light and background activity ones.

The new behaviors are:
- WalkBehavior: uses CARLA AI controller to randomly move a pedestrian around the town
- CrossingBehavior: dynamically controls the speed of a crossing object so that it and a reference actor arrive at a spot in the road at the same time

Additionally, fixed bug at Range.__contains__ class, which had no self when calling attribute low and high


examples/carla/Carla_Challenge/carlaChallenge10.scenic
```diff
@@ -43,16 +45,16 @@ adv_trajectory = [adv_maneuver.startLane, adv_maneuver.connectingLane, adv_maneu
...
+ego = Car at ego_spawn_pt,
+    with blueprint EGO_MODEL,
+    with behavior EgoBehavior(EGO_SPEED, ego_trajectory)
 
-adversary = Car following roadDirection from adv_spawn_point for ADV_DISTANCE_TO_INTERSECTION,
-	with behavior CrossingCarBehavior(adv_trajectory)
+adversary = Car at adv_spawn_pt,
+    with behavior AdversaryBehavior(adv_trajectory)
 
+require (distance to intersec) in Range(20, 25)
```
dfremont, line 58: This only works due to what is arguably a bug in Scenic - the `Range` is a distribution and should get sampled to a single number, at which point the `in` operator wouldn't work. In case I fix this bug later, it would be safer to simply write `20 <= (distance to intersec) <= 25`.

dfremont (changes requested):
I read and tested this and it looks good, except for one minor issue with `Range` being used in a non-standard way that may not work in the future (see comment below - there are several places where this happens).

examples/carla/Carla_Challenge/carlaChallenge10.scenic
joel-mb: @dfremont Thanks for the quick review! Updated scenarios to avoid using `Range` in those situations.

dfremont (approved):

PR #22, opened 2021-02-03
Add ability to record videos of dynamic scenarios

JayShenoy:


src/scenic/simulators/carla/recording/__main__.py
```diff
@@ -0,0 +1,36 @@
+import os
+import subprocess
+import argparse
+import json
+
+parser = argparse.ArgumentParser(prog='scenic', add_help=False,
+                                 usage='scenic [-h | --help] [options] FILE [options]',
+                                 description='Sample from a Scenic scenario, optionally '
```
dfremont, line 8: This is misleading, since this entry point doesn't support most of Scenic's options. Please change `prog`, `usage`, and `description` to make clear this is a helper script for running multiple scenarios and collecting recordings.

(If you decide to keep this entry point. You could just get rid of it, or add the `--scenarios` functionality to the main Scenic entry point, which would allow you to drop the `subprocess` stuff.)

src/scenic/simulators/carla/simulator.py
```diff
@@ -243,3 +397,120 @@ def destroy(self):
...
+
+	def process_lidar_data(self, lidar_data, buffer):
+		buffer.append(lidar_data)
+
+	def process_radar_data(self, radar_data, buffer):
+		buffer.append(radar_data)
+
+	def save_recordings(self, save_dir):
+		if not self.record_sensors:
+			print('No recordings saved; turn on recordings for simulator to enable')
```
dfremont, line 421: If you want this to not be a fatal error, use `warnings.warn` instead; otherwise raise an exception.

src/scenic/simulators/carla/simulator.py
```diff
@@ -243,3 +397,120 @@ def destroy(self):
...
+			return
+
+		print('Started saving recorded data')
```
dfremont, line 424: Use `verbosePrint` so that the `--verbosity` option can suppress this message.

dfremont (changes requested):
Thanks for preparing this PR! It looks good except for a few minor issues -- see comments below. The main things are to add documentation for the new features into the Sphinx documentation, and to clean up the `scenic.simulators.carla.recording` entry point (I'm not sure if you meant to include it, since it's in the CARLA folder but the `-r` option is set up so that multiple simulators could support it).

dfremont:
This has been superseded by the new sensors system, so I'm closing this. @ek65 if there's any functionality from this PR that you can't replicate with the new system, we can discuss.

PR #25, opened 2021-02-28
Use `allowCollisions` property to check whether intersections should be allowed.

afsafzal:


dfremont:
Many thanks for this catch, @afsafzal! (No idea how I missed this before...) The intended semantics of `allowCollisions` is that when it is set to `True`, any collisions with the object are allowed: that is, it is OK for two objects to intersect if *at least one* of them has `allowCollisions` set. So I've fixed the bug slightly differently in 5eac20e1219 rather than merging your pull request.

PR #33, opened 2021-06-17
Set CARLA actor's color, width and length

abol-karimi:
When creating an actor in the CARLA simulator, set the properties of the corresponding Scenic agent.

### DIFF
--- a/src/scenic/core/simulators.py
+++ b/src/scenic/core/simulators.py
@@ -247,7 +247,7 @@ class Simulation:
 
         The default implementation returns a tuple of the positions of all objects.
         """
-        return tuple(obj.position for obj in self.objects)
+        return {obj.name:[obj.position, obj.heading] for obj in self.objects}
 
     def destroy(self):
         """Perform any cleanup necessary to reset the simulator after a simulation."""
--- a/src/scenic/domains/driving/actions.py
+++ b/src/scenic/domains/driving/actions.py
@@ -93,7 +93,7 @@ class SetSteerAction(SteeringAction):
 	def applyTo(self, obj, sim):
 		obj.setSteering(self.steer)
 
-class SetBrakeAction(SteeringAction):
+class SetBrakeAction(Action):
 	"""Set the amount of brake.
 
 	Arguments:
@@ -135,7 +135,7 @@ class SetReverseAction(SteeringAction):
 	def applyTo(self, obj, sim):
 		obj.setReverse(self.reverse)
 
-class RegulatedControlAction(SteeringAction):
+class RegulatedControlAction(Action):
 	"""Regulated control of throttle, braking, and steering.
 
 	Controls throttle and braking using one signal that may be positive or negative.
--- a/src/scenic/simulators/carla/simulator.py
+++ b/src/scenic/simulators/carla/simulator.py
@@ -160,12 +160,12 @@ class CarlaSimulation(DrivingSimulation):
 		loc = utils.scenicToCarlaLocation(obj.position, world=self.world, blueprint=obj.blueprint)
 		rot = utils.scenicToCarlaRotation(obj.heading)
 		transform = carla.Transform(loc, rot)
-		
-		# Set the color
+
+		# Color
 		c = obj.color
 		c_str = f'{int(c.r*255)},{int(c.g*255)},{int(c.b*255)}'
 		blueprint.set_attribute('color', c_str)
-		
+
 		# Create Carla actor
 		carlaActor = self.world.try_spawn_actor(blueprint, transform)
 		if carlaActor is None:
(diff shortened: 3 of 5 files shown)


src/scenic/core/simulators.py
```diff
@@ -247,7 +247,7 @@ def currentState(self):
 
         The default implementation returns a tuple of the positions of all objects.
         """
-        return tuple(obj.position for obj in self.objects)
+        return {obj.name:[obj.position, obj.heading] for obj in self.objects}
```
dfremont, line 250: Changing this default behavior would break compatibility with people who are assuming it behaves as it did before (and as described in the docstring). It's probably a good idea to save the heading in `trajectory` as well, but I'd like to wait until the next backwards-incompatible release of Scenic.

In the mean time, we now have the `record` statement for saving additional information from the simulation; I'd recommend using that, or implementing your own subclass of `CarlaSimulation` that overrides `currentState`.

src/scenic/domains/driving/actions.py
```diff
@@ -93,7 +93,7 @@ def __init__(self, steer: float):
 	def applyTo(self, obj, sim):
 		obj.setSteering(self.steer)
 
-class SetBrakeAction(SteeringAction):
+class SetBrakeAction(Action):
```
dfremont, line 96: Why make this change? This action is only supported by objects which support the steering protocol.

src/scenic/simulators/carla/simulator.py
```diff
@@ -161,6 +161,11 @@ def createObjectInSimulator(self, obj):
 		rot = utils.scenicToCarlaRotation(obj.heading)
 		transform = carla.Transform(loc, rot)
 
+		# Color
+		c = obj.color
+		c_str = f'{int(c.r*255)},{int(c.g*255)},{int(c.b*255)}'
+		blueprint.set_attribute('color', c_str)
```
dfremont, line 167: Does this change the color of all vehicles using the same blueprint?

src/scenic/simulators/carla/simulator.py
```diff
@@ -171,6 +176,8 @@ def createObjectInSimulator(self, obj):
 		carlaActor.set_simulate_physics(obj.physics)
 
 		if isinstance(carlaActor, carla.Vehicle):
+			obj.width = carlaActor.bounding_box.extent.y * 2
+			obj.length = carlaActor.bounding_box.extent.x * 2
```
dfremont, line 180: Mutating these properties should not be done at simulation time, if possible: Scenic needs to know the dimensions of the objects during scene generation to ensure that no two objects collide, etc. Can the bounding box be extracted earlier? If not, we could leave this in as a hack for now, but having the width and length change between the sampled scene and the generated simulation would be quite surprising for users I think.

dfremont (changes requested):
Hello Abol, thanks for these improvements, and apologies for the delay! A few of your changes don't completely make sense to me, so I've left some comments. If you can reply to those I'd be happy to get at least some of these changes merged in.

src/scenic/core/simulators.py
abol-karimi: I probably included this change accidentally because I wanted to focus only on one feature for this PR. Thanks for pointing out the alternative options, `record` and sub-classing.

src/scenic/domains/driving/actions.py
abol-karimi: If I remember correctly, I used to get a runtime error from Scenic's type-compatibility checker. Perhaps it's not an issue in the latest version of Scenic.

src/scenic/simulators/carla/simulator.py
abol-karimi: It should only affect the actor that is spawned with this blueprint in the next statement (line 170), so there shouldn't be a problem.

src/scenic/simulators/carla/simulator.py
abol-karimi: That's a good point on inconsistency between a scene and its simulation. However, I couldn't find a way to retrieve the size of an actor before it is spawned. I don't think CARLA blueprints have size attributes.

abol-karimi:
Hi Daniel, thanks for considering these changes and for your helpful comments. I addressed your points.

src/scenic/simulators/carla/simulator.py
dfremont: Yes, I can't find anything either. We can leave this for now.

src/scenic/simulators/carla/simulator.py
dfremont: OK, thanks.

src/scenic/domains/driving/actions.py
dfremont: OK, will revert this then, but let me know if you encounter any problems later.

dfremont (approved):

src/scenic/simulators/carla/simulator.py
abol-karimi: One way to handle the dimensions would be the following:
Before simulating the Scenic scenario, the Scenic-Carla interface can spawn all the blueprints available in the blueprint library, then measure the dimensions of each agent and filter out those blueprints that don't match the size of any of the the scenario's actors. If no blueprint remains, an error is reported that the simulator cannot support this scenario.

PR #42, opened 2022-02-18
Introduce black code formatter

shumbo:
## TLDR

This PR sets up [black](https://github.com/psf/black) on the Scenic repo. `make format` formats all python files in the repository so developers don't need to manually format code. It would also make the coding style consistent across files.

I do not intend this PR to be merged anytime soon. It's more of posing a problem on code formatting and this is one possible solution.

## Motivation

I've been working on the Scenic repo for a while, and one thing I noticed is that it doesn't have a code formatter installed. There are two reasons why I think code formatters are beneficial.

First, developers don't need to manually format their code. Due to the nature of the project, I sometimes write functions and methods with many (4+) arguments. When I do, I have to manually add line breaks to maintain readability, and it would be nice to automate that process. While there are some cases where you want to do manual styling, I believe that formatting generated by a formatter is good enough in most cases.

Second, as you can see in the diff of this PR, there are inconsistencies in coding styles between files. For example,

- Some files use `'` while others use `"` for string literals
- Some files use `<tab>` for indent, while others use four spaces

Code formatters can detect those inconsistencies and fix them automatically.

## black code formatter

https://github.com/psf/black

I installed black. It's been used by some of the most popular projects and I personally have a great experience with it.

I executed the command and formatted all Python files. This is why this PR is incredibly long.

## Makefile and GitHub Actions

It is important that we enforce the coding styles. In this regard, in addition to installing black, I created a Makefile and GitHub Actions workflow.

Makefile provides two useful commands: `make format` and `make format-check`.

`make format` would run black and format all files in the repository. Developers are encouraged to run this command before they commit.

`make format-check` would run black with the `--check` flag, so it exits with an error if it finds an unformatted file.

When you push to GitHub, it would run the `check` workflow. It sets up dependencies, including Python and Poetry, and run `make format-check`. This way, developers would notice if they've committed unformatted files.

In the future, this workflow can do more things, like running tests or publishing the package.

## Discussion

As stated earlier, actually merging this PR would involve several concerns. 

- this might cause conflicts when merging branch
	- rebasing this first and running the formatter should minimize conflicts
	- might want to merge this when everyone's work is merged to master
- `git blame` may be useless
	- but this is also a reason to introduce a code formatter as early as possible

I'd appreciate it if you could share your opinion. I know it's not really of priority, but it would be nice if the project can have something similar so developers do not have to worry about formatting much while maintaining the readability!


dfremont:
Superseded by #148; thanks again @shumbo for the proposal and initial implementation!

PR #48, opened 2022-05-16
Implement Temporal Operators

shumbo:
This PR adds the following five temporal operators to the Scenic language.

- eventually
- always
- next
- until
- implies

These operators can be used inside `require` statements as a prefix or infix operator.

### DIFF
--- a/src/scenic/core/propositions.py
+++ b/src/scenic/core/propositions.py
@@ -53,10 +53,9 @@ class PropositionNode:
 		if isinstance(node, Always):
 			node = node.req
 
-		eligible = True
-		for n in node.flatten():
-			# turns false if any one of the node is temporal
-			eligible = eligible and (not n.is_temporal)
+		# eligible if none of the nodes was temporal
+		eligible = all(not n.is_temporal for n in node.flatten())
+
 		return node.syntax_id if eligible else None
 	
 	@property
@@ -90,10 +89,7 @@ class PropositionNode:
 	@property
 	def has_temporal_operator(self):
 		node = self
-		has_temporal_op = False
-		for n in node.flatten():
-			# turns false if any one of the node is temporal
-			has_temporal_op = has_temporal_op or n.is_temporal
+		has_temporal_op = any(n.is_temporal for n in node.flatten())
 		return has_temporal_op
 
 class Atomic(PropositionNode):
--- a/src/scenic/core/propositions.py
+++ b/src/scenic/core/propositions.py
@@ -150,7 +150,7 @@ class Not(UnaryProposition):
 
 class And(PropositionNode):
 	def __init__(self, reqs, syntax_id):
-		ltl_node = rv_ltl.And(reqs)
+		ltl_node = rv_ltl.And(*[req.ltl_node for req in reqs])
 		super().__init__(syntax_id, ltl_node)
 		self.reqs = reqs
 	def __str__(self):
@@ -163,7 +163,7 @@ class And(PropositionNode):
 
 class Or(PropositionNode):
 	def __init__(self, reqs, syntax_id):
-		ltl_node = rv_ltl.Or(reqs)
+		ltl_node = rv_ltl.Or(*[req.ltl_node for req in reqs])
 		super().__init__(syntax_id, ltl_node)
 		self.reqs = reqs
 	def __str__(self):
--- a/src/scenic/core/propositions.py
+++ b/src/scenic/core/propositions.py
@@ -185,3 +185,15 @@ class Until(PropositionNode):
 	@property
 	def children(self):
 		return [self.lhs, self.rhs]
+
+class Implies(PropositionNode):
+	def __init__(self, lhs, rhs, syntax_id) -> None:
+		self.lhs = lhs
+		self.rhs = rhs
+		ltl_node = rv_ltl.Implies(lhs.ltl_node, rhs.ltl_node)
+		super().__init__(syntax_id, ltl_node)
+	def __str__(self):
+		return f"({self.lhs} implies {self.rhs})"
+	@property
+	def children(self):
+		return [self.lhs, self.rhs]
--- a/src/scenic/core/propositions.py
+++ b/src/scenic/core/propositions.py
@@ -98,7 +98,7 @@ class Atomic(PropositionNode):
 		super().__init__(syntax_id, ap)
 		self.closure = closure
 	def __str__(self):
-		return f"(AP)"
+		return f"(AP syntax_id={self.syntax_id})" if self.syntax_id is not None else "(AP)"
 	def evaluate(self):
 		return self.closure()
 
--- a/src/scenic/core/propositions.py
+++ b/src/scenic/core/propositions.py
@@ -2,9 +2,10 @@
 
 from functools import reduce
 import operator
-import rv_ltl
-from scenic.core.errors import InvalidScenarioError
 
+import rv_ltl
+
+from scenic.core.errors import InvalidScenarioError
 from scenic.core.lazy_eval import needsLazyEvaluation
 
 class PropositionMonitor:
--- a/src/scenic/core/propositions.py
+++ b/src/scenic/core/propositions.py
@@ -177,6 +177,7 @@ class Until(PropositionNode):
 		self.rhs = rhs
 		ltl_node = rv_ltl.Until(lhs.ltl_node, rhs.ltl_node)
 		super().__init__(syntax_id, ltl_node)
+		self.is_temporal = True
 	def __str__(self):
 		return f"({self.lhs} until {self.rhs})"
 	@property
--- a/src/scenic/core/propositions.py
+++ b/src/scenic/core/propositions.py
@@ -1,0 +1,187 @@
+"""Objects representing propositions that can be used to specify conditions"""
+
+from functools import reduce
+import operator
+import rv_ltl
+from scenic.core.errors import InvalidScenarioError
+
+from scenic.core.lazy_eval import needsLazyEvaluation
+
+class PropositionMonitor:
+	def __init__(self, proposition: "PropositionNode") -> None:
+		self._proposition = proposition
+		self._monitor = proposition.ltl_node.create_monitor()
+		
+	def update(self):
+		atomic_propositions = self._proposition.atomics()
+		state = {}
+		for ap in atomic_propositions:
+			b = ap.closure()
+			if needsLazyEvaluation(b):
+				raise InvalidScenarioError(f'value undefined outside of object definition')
+			state[str(ap.syntax_id)] = b
+		self._monitor.update(state)
+		return self._monitor.evaluate()
+
+
+class PropositionNode:
+	"""Base class for temporal and non-temporal propositions"""
+	def __init__(self, syntax_id, ltl_node) -> None:
+		self.syntax_id = syntax_id
+		self.ltl_node = ltl_node
+
+		self.is_temporal = False
+		"""tells if the proposition is temporal"""
+
+	def check_constrains_sampling(self):
+		"""checks if the proposition can be used for pruning
+
+		A requirement can be used for pruning if it is evaluated on the scene generation phase before simulation, and
+		violation in that phase immediately results in discarding the scene and regenerating a new one.
+
+		For simplicity, we currently check two special cases:
+		1. requirements with no temporal requirement
+		2. requirements with only one `always` operator on top-level
+
+		Returns:
+			None|int: returns syntax id that can be used for pruning, None if proposition cannot be used for pruning
+		"""
+		node = self
+
+		# if `always` is on top-level, check what's inside
+		if isinstance(node, Always):
+			node = node.req
+
+		eligible = True
+		for n in node.flatten():
+			# turns false if any one of the node is temporal
+			eligible = eligible and (not n.is_temporal)
+		return node.syntax_id if eligible else None
+	
+	@property
+	def children(self) -> list["PropositionNode"]:
+		"""returns all children of proposition tree
+
+		Returns:
+			list: proposition nodes that are directly under this node
+		"""
+		return []
+
+	def flatten(self) -> list["PropositionNode"]:
+		"""flattens the tree and return the list of nodes
+
+		Returns:
+			list: list of all children nodes
+		"""
+		return [self] + reduce(
+			operator.concat, [node.flatten() for node in self.children], []
+		)
+	
+	def atomics(self) -> list["Atomic"]:
+		return list(filter(lambda n: isinstance(n, Atomic), self.flatten()))
+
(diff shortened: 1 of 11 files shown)


src/scenic/core/propositions.py
```diff
@@ -0,0 +1,199 @@
+"""Objects representing propositions that can be used to specify conditions"""
+
+from functools import reduce
+import operator
+import rv_ltl
+from scenic.core.errors import InvalidScenarioError
+
+from scenic.core.lazy_eval import needsLazyEvaluation
```
dfremont, line 8: A very minor stylistic point: I like to group top-level imports into 3 sections, separated by blank lines: Python standard library imports, other external imports, then imports from Scenic itself. This helps make clear which external/internal dependencies the file has.

src/scenic/core/propositions.py
```diff
@@ -0,0 +1,199 @@
...
+		node = self
+
+		# if `always` is on top-level, check what's inside
+		if isinstance(node, Always):
+			node = node.req
+
+		eligible = True
+		for n in node.flatten():
+			# turns false if any one of the node is temporal
+			eligible = eligible and (not n.is_temporal)
```
dfremont, line 58: I find this a little awkward to read; what about `eligible = all(not n.is_temporal for n in node.flatten())`?

src/scenic/core/propositions.py
```diff
@@ -0,0 +1,199 @@
...
+	def evaluate(self):
+		raise RuntimeError("This proposition contains temporal operators and can only be evaluated using monitor")
+
+	@property
+	def has_temporal_operator(self):
+		node = self
+		has_temporal_op = False
+		for n in node.flatten():
+			# turns false if any one of the node is temporal
+			has_temporal_op = has_temporal_op or n.is_temporal
```
dfremont, line 95: Same comment as above.

src/scenic/core/propositions.py
```diff
@@ -0,0 +1,199 @@
...
+		return has_temporal_op
+
+class Atomic(PropositionNode):
+	def __init__(self, closure, syntax_id = None):
+		ap = rv_ltl.Atomic(identifier=str(syntax_id))
+		super().__init__(syntax_id, ap)
+		self.closure = closure
+	def __str__(self):
+		return f"(AP)"
```
dfremont, line 104: Maybe should include `syntax_id` here?

src/scenic/core/requirements.py
```diff
@@ -36,24 +39,31 @@ def __init__(self, ty, condition, line, prob, name, ego):
         self.prob = prob
         self.name = name
 
-        # the translator wrapped the requirement in a lambda to prevent evaluation,
-        # so we need to save the current values of all referenced names; we save
-        # the ego object too since it can be referred to implicitly
```
dfremont, line 41: Let's keep this comment, which explains what the bindings are doing here.

src/scenic/syntax/translator.py
```diff
@@ -985,19 +1029,32 @@ def wrapStatementCall():
...
 					if context in modifierNames:
 						allowedTerminators = modifierNames[context].terminators
 					elif context in terminatorsForStatements:
 						allowedTerminators = terminatorsForStatements[context]
+			# construct allowedInfixOps
+			if context:
+				# all function calls on the stack
+				contexts = set(ctx for ctx, _ in functionStack)
+				for opTokens, op in infixTokens.items():
+					# if not context is specified, the operator can be used
```
dfremont, line 1044: "not" -> "no"

src/scenic/syntax/translator.py
```diff
@@ -985,19 +1029,32 @@ def wrapStatementCall():
...
+					if not op.contexts:
+						allowedInfixOps[opTokens] = op.tokens
+						continue
+					# if operator is start level only and it is start level, cannot be used
```
dfremont, line 1048: "it is" -> "it is not"

src/scenic/syntax/translator.py
```diff
@@ -985,19 +1029,32 @@ def wrapStatementCall():
...
+					if op.startLevelOnly and parenLevel != startLevel:
+						continue
+					# a set of contexts under which this operator is allowed
+					allowedContexts = set(op.contexts)
```
dfremont, line 1052: Can we create these sets once at operator definition time to avoid constantly recreating them? It might be faster to use a `frozenset` too (although not worth spending much time investigating that).

src/scenic/syntax/translator.py
```diff
@@ -1824,6 +2074,40 @@ def visit_Call(self, node):
 		"""
 		func = node.func
 		newArgs = []
+
+		# wrap temporal requirement in lambda
+		if isinstance(func, Name) and func.id in PROPOSITION_FACTORY:
+			if not self.canUseTemporalOps:
+				self.parseError(node, f"Operator {func.id} is not allowed in this statement")
+			checkedArgs = node.args
+			self.validateSimpleCall(node, (1, None), args=checkedArgs)
```
dfremont, line 2083: You can just leave off the `args` keyword argument here and drop `checkedArgs`.

tests/syntax/test_requirements.py
```diff
@@ -86,6 +86,14 @@ def test_runtime_parse_error_in_requirement():
     with pytest.raises(ScenicSyntaxError):
         sampleScene(scenario)
 
+def test_soft_requirement_with_temporal_operators():
+    "Temporal operators are not be allowed if prob != 1"
```
dfremont, line 90: Extra "be"

src/scenic/syntax/veneer.py
```diff
@@ -441,8 +438,6 @@ def makeRequirement(ty, reqID, req, line, name):
 		raise RuntimeParseError(f'tried to use "{ty.value}" inside a requirement')
 	elif currentBehavior is not None:
 		raise RuntimeParseError(f'"{ty.value}" inside a behavior on line {line}')
-	elif currentSimulation is not None:
-		currentScenario._addDynamicRequirement(ty, req, line, name)
```
dfremont, line 445: Removing this case looks suspect to me: since `_addRequirement` just adds the requirement to a list of pending requirements, won't the requirement not get enforced?

dfremont (changes requested):
Looks great overall, but I'm not sure it works inside modular scenarios. We need some more tests along the lines of those testing requirements in "test_modular.py".

src/scenic/core/propositions.py
shumbo: Fixed in fc851b08d425ac04d7bcd85f255be102e59d691a

src/scenic/core/propositions.py
shumbo: Good point. Fixed in 1f01cda92fbea858e0fbb053461a2e49b0231796

src/scenic/core/propositions.py
shumbo: 1f01cda92fbea858e0fbb053461a2e49b0231796

src/scenic/core/propositions.py
shumbo: Changed to include `syntax_id` acbc17b0c8e94cd0e5089ccc1ad59ceeb536ed99

src/scenic/syntax/translator.py
shumbo: Fixed 39474008e0421aa629590c8142ee0be642314e61

src/scenic/syntax/translator.py
shumbo: Fixed in 8da575ad72765f9fb3602857d56148ce4070e755

src/scenic/syntax/translator.py
shumbo: I agree that these sets should be created once and reused. Although `op.contexts` was a tuple, the order should not matter. Therefore, I changed the type of `contexts` field of `InfixOp` to `FrozenSet` so we can just take the intersection.

I found [this StackOverflow post](https://stackoverflow.com/questions/36555214/set-vs-frozenset-performance) which claims that there's no big performance difference between `set` and `frozenset`. However, these sets clearly should be immutable and I believe it makes sense to use `frozenset`.

src/scenic/syntax/translator.py
shumbo: Thanks! Fixed in 93f296c321882bb64f72387900eeaf5029e95ce2

tests/syntax/test_requirements.py
shumbo: Fixed in f30d11e4dcf3bcc6d8a894fc66b753e9c2af6c2d

src/scenic/core/propositions.py
```diff
@@ -0,0 +1,196 @@
...
+	def __init__(self, lhs, rhs, syntax_id) -> None:
+		self.lhs = lhs
+		self.rhs = rhs
+		ltl_node = rv_ltl.Until(lhs.ltl_node, rhs.ltl_node)
+		super().__init__(syntax_id, ltl_node)
+	def __str__(self):
+		return f"({self.lhs} until {self.rhs})"
+	@property
+	def children(self):
+		return [self.lhs, self.rhs]
```
shumbo, line 184: `is_temporal` should be True

src/scenic/syntax/veneer.py
shumbo: You're right that these two lines are required and I've restored the changes in 85be02d90ee932129e77933b329a1bcbf58e676d.

I deleted them because no test case uses this branch, so I added one in 3aaf090547648b25d14884c3833f91bda694cc3b

Also, there was an assertion that was no longer valid in `dynamics.py` so I deleted it. 7e00a84c44b73c3ed6cdb97ea380d145e0b49a79

Before this PR, `ty` of `_addDynamicRequirement` needed to be anything but `require` because plain requirements are not dynamic requirements. 

https://github.com/BerkeleyLearnVerify/Scenic/blob/9215576521c1c533d0f71de17d15e706fe31a991/src/scenic/core/dynamics.py#L490

With temporal requirements, plain requirements can also be passed as a dynamic requirement. Therefore, I'm removing this assertion.

src/scenic/core/propositions.py
shumbo: Fixed in 84c582ef75ed36c844dcc315bffbbfc849c83797

src/scenic/core/requirements.py
```diff
@@ -130,18 +151,23 @@ def __str__(self):
...
-        assert compiledReq.prob == 1
         self.sample = sample
+        self.compiledReq = compiledReq
+        self.proposition = proposition
 
     def isTrue(self):
         return self.value()
 
     def value(self):
+        one_time_monitor = self.proposition.create_monitor()
```
dfremont, line 167: Is this necessary? If you don't pass a monitor at all, won't the closure test the correct proposition anyway? If so, there shouldn't be any need for 3 different methods to check the truth value (even for the requirements with built-in monitors, they're confusing: if we need them, let's put docstrings explaining the difference).

src/scenic/syntax/translator.py
```diff
@@ -269,7 +271,7 @@ def compileStream(stream, namespace, params={}, model=None, filename='<stream>')
 			# Parse the translated source
 			tree = parseTranslatedSource(newSource, filename)
 			# Modify the parse tree to produce the correct semantics
-			newTree, requirements = translateParseTree(tree, allConstructors, filename)
+			newTree, requirements, propositionSyntax = translateParseTree(tree, allConstructors, filename)
```
dfremont, line 274: Looks like `requirements` is no longer used; if so, I'd remove it.

src/scenic/syntax/translator.py
```diff
@@ -635,13 +662,29 @@ class InfixOp(typing.NamedTuple):
...
+PROPOSITION_AND = "PropositionAnd"
+PROPOSITION_OR = "PropositionOr"
+PROPOSITION_NOT = "PropositionNot"
+ATOMIC_PROPOSITION = "AtomicProposition"
+
+# List of functions that returns objects that supports temporal evaluation
+PROPOSITION_FACTORY = (
+    (PROPOSITION_AND, PROPOSITION_OR, PROPOSITION_NOT, ATOMIC_PROPOSITION)
+    + tuple(impl for _, impl in temporalPrefixOperators.items()) # prefix operators
+	+ tuple(op.implementation for op in temporalInfixOperators) # infix operators
```
dfremont, line 684: Extra indent on this line.

src/scenic/syntax/translator.py
```diff
@@ -1824,6 +2073,39 @@ def visit_Call(self, node):
 		"""
 		func = node.func
 		newArgs = []
+
+		# wrap temporal requirement in lambda
```
dfremont, line 2077: Please add an entry at the start of the list in the docstring explaining what this does, since it's a new transformation done by `visit_Call`.

src/scenic/syntax/veneer.py
```diff
@@ -383,15 +387,21 @@ def require(reqID, req, line, name, prob=1):
...
+			if prob >= 1 or random.random() <= prob:
+				result = req.evaluate()
+				assert not needsSampling(result)
+				if needsLazyEvaluation(result):
+					raise RuntimeParseError(f'requirement on line {line} uses value'
+											' undefined outside of object definition')
+				if not result:
+					raise RejectSimulationException(name)
 	else:	# requirement being defined at compile time
+		if req.is_temporal and prob != 1:
```
dfremont, line 403: This should be `req.has_temporal_operator` to catch temporal operators nested under non-temporal operators, right?

dfremont: Also, shouldn't this check be done in the runtime case as well? Maybe you can just move it to before line 385.

tests/syntax/test_dynamics.py
```diff
@@ -562,6 +562,85 @@ def test_require_eventually_2():
...
+        behavior Foo():
+            while True:
+                self.blah += 1
+                take self.blah
+        ego = Object with behavior Foo, with blah 0
+        require False until ego.blah > 3
+    """)
+    with pytest.raises(RejectSimulationException):
+        sampleEgoActions(scenario, maxSteps=5)
+
```
dfremont, line 621: Should add a test that `require A until B` is rejected if `B` never happens.

tests/syntax/test_requirements.py
```diff
@@ -86,6 +86,14 @@ def test_runtime_parse_error_in_requirement():
     with pytest.raises(ScenicSyntaxError):
         sampleScene(scenario)
 
+def test_soft_requirement_with_temporal_operators():
+    "Temporal operators are not allowed if prob != 1"
```
dfremont, line 90: Triple double-quotes, please.

dfremont (changes requested):
Looks good overall -- a couple simple fixes and points to clarify.

dfremont:
Closing this PR as the changes were ultimately merged through #113.

PR #50, opened 2022-06-03
Carla pedestrian color fix

theilem:
Added a check for color in the blueprint to fix crashes when creating pedestrians in Carla


dfremont (approved):
Thanks, Mirco!

PR #51, opened 2022-06-16
[Parser] Install Pegen and Setup Parser Generation

shumbo:
## What I did

This PR adds `pegen` to Scenic's dependencies and some files to generate the Scenic parser.

### Pegen

I added `pegen` as a dependency with `optional = true` and included it in the `dev` extra packages.

### `scenic.gram`

I added `scenic.gram` to the new `./grammar` directory. At this moment, `scenic.gram` is almost identical to Python grammar in the Pegen repository (https://github.com/we-like-parsers/pegen/blob/87f8d75d703f41f8cfa04bd862ce1ab01c660d43/data/python.gram).

The only difference is that I renamed the name of the generated parser class from `PythonParser` to `ScenicParser`. See 545d930669383bfd935c88f8d9f6cc12fca359a5 for diff.

### Makefile

Added a makefile in the root directory for generating the parser. 

### Should we include the generated parser in this git repo?

I added the generated parser to `gitignore` and excluded it from the repository. I see both pros and cons of not including the generated parser from the repository.

Pros:

- Parser can always be generated from the grammar
- Minimize diff by changing the grammar
  - pegen creates temporary grammar rules based on numbers, so changing a rule at the top affects the entire parser

Cons:

- The current version of Scenic can be installed by cloning the repository and running `poetry install`. If we do not include the parser in the repo, users will need to generate the parser before they run `poetry install`.
  - We can include the generated parser in the version we upload to PyPI
  - Poetry does not provide a way to run scripts before install, so users will need to explicitly run the command, or we can provide an installation script.

If we want the non-developer to be able to install Scenic from GitHub, we might also want to move `pegen` to normal dependencies, not the `dev` extra package set.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -131,3 +131,6 @@ venv.bak/
 dmypy.json
 
 *.cproject
+
+# generated parser
+src/scenic/parser/parser.py
--- a/grammar/scenic.gram
+++ b/grammar/scenic.gram
@@ -1,6 +1,6 @@
-# PEG grammar for Python
+# PEG grammar for Scenic
 
-@class PythonParser
+@class ScenicParser
 
 @subheader'''
 import enum
@@ -68,7 +68,7 @@ def parse_file(
             tokenize.generate_tokens(f.readline)
         )
         tokenizer = Tokenizer(tok_stream, verbose=verbose, path=path)
-        parser = PythonParser(
+        parser = ScenicParser(
             tokenizer,
             verbose=verbose,
             filename=os.path.basename(path),
@@ -93,7 +93,7 @@ def parse_string(
         tokenize.generate_tokens(io.StringIO(source).readline)
     )
     tokenizer = Tokenizer(tok_stream, verbose=verbose)
-    parser = PythonParser(tokenizer, verbose=verbose, py_version=py_version)
+    parser = ScenicParser(tokenizer, verbose=verbose, py_version=py_version)
     return parser.parse(mode if mode == "eval" else "file")
 
 
--- a/grammar/scenic.gram
+++ b/grammar/scenic.gram
@@ -1,0 +1,2070 @@
+# PEG grammar for Python
+
+@class PythonParser
+
+@subheader'''
+import enum
+import io
+import itertools
+import os
+import sys
+import token
+from typing import (
+    Any, Callable, Iterator, List, Literal, Tuple, TypeVar, Union, NoReturn
+)
+
+from pegen.tokenizer import Tokenizer
+
+# Singleton ast nodes, created once for efficiency
+Load = ast.Load()
+Store = ast.Store()
+Del = ast.Del()
+
+Node = TypeVar("Node")
+FC = TypeVar("FC", ast.FunctionDef, ast.AsyncFunctionDef, ast.ClassDef)
+
+EXPR_NAME_MAPPING = {
+    ast.Attribute: "attribute",
+    ast.Subscript: "subscript",
+    ast.Starred: "starred",
+    ast.Name: "name",
+    ast.List: "list",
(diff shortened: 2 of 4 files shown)


Eric-Vin:
I like the idea of not including the generated parser and transitioning to a Makefile. This also makes it easier to not entirely depend on Poetry in the future, which I know Daniel has talked about. For packaging on PyPi I'm sure we can manually include the generated parser or modify any GitHub actions to build and include it in the future.

grammar/scenic.gram
```diff
@@ -0,0 +1,2070 @@
+# PEG grammar for Scenic
```
dfremont, line 1: Let's acknowledge here that this grammar is adapted from the Python grammar and include a link to the latter.

pyproject.toml
```diff
@@ -22,7 +22,7 @@ classifiers = [
 ]
 
 [tool.poetry.dependencies]
-python = "^3.7, <3.11"
+python = "^3.8, <3.11"
```
dfremont, line 25: Since we're (necessarily) dropping support for 3.7, let's update the "Getting Started with Scenic" page in the documentation correspondingly.

pyproject.toml
```diff
@@ -65,6 +66,7 @@ dev = [
 	"pygments",
 	"astor",
 	"pyproj",	# for running guideways modules
+	"pegen",
 ]
 test = [	# minimum dependencies for running tests (used for tox virtualenvs)
```
dfremont, line 71: We don't necessarily need to do this now, but eventually `pegen` will probably have to be included in the `test` extra as well since the test suite should check that the grammar compiles and then use the resulting parser.

.gitignore
```diff
@@ -131,3 +131,6 @@ venv.bak/
 dmypy.json
 
 *.cproject
+
+# generated parser
+src/scenic/parser/parser.py
```
dfremont, line 136: Any particular reason to make new folders instead of putting both the grammar and the generated parser under `src/syntax`?

dfremont (changes requested):
Looks good -- just some minor comments below.

I'm fine with switching to a Makefile as the entry point for installation from the repo, as long as it's painless for people (and we update the documentation accordingly). Then we can exclude the generated parser from the repo and include its generation in the Makefile.

grammar/scenic.gram
shumbo: da576f1b009297e2e175852591cd9a820d9abeee
Added a sentence right below. Let me know if we should include more information (I think we can also discuss this later, but just adding this not to forget).

.gitignore
shumbo: I simply followed the structure of the cpython repository where `/Grammar` is just for storing the `.gram` file. If `src` is for storing Python files, maybe it makes sense to have a separate directory for non-Python sources? I don't have a strong opinion on this, so I'm fine with putting the grammar inside `src/syntax` or maybe `src/parser`.

pyproject.toml
shumbo: I added a task on Todoist on this. Let me update that page when I finalize the installation with Makefile.

pyproject.toml
shumbo: > the test suite should check that the grammar compiles and then use the resulting parser

Forgot about this. Thanks for letting me know.

Also added this on Todoist. I will work on this when I will add tests for the parser generation and use the resulting parser for the rest of the test suite.

.gitignore
shumbo: 63621cf9d1c4e14b4c8649eee1000a1499348e91

moved both the grammar and generated parser under `syntax`

dfremont (approved):

PR #52, opened 2022-06-20
[Parser] Add Scenic AST module and Scenic Compiler

shumbo:
This PR adds **Scenic AST module** and **Scenic Compiler**. It also changes `translator.py` to use the new parser, AST module, and compiler instead of the old token translator.

The goal of this PR is to complete the new parser pipeline. In order to demonstrate that the new parser actually works, I added support for ego assignments (`ego = ...`) and the new keyword without specifiers.

The Scenic CLI uses the new parser to execute Scenic programs. Most language features are missing at this moment but can execute very simple programs like

```
ego = new Object
```

### Grammar

Added two rules: ego assignment and the `new` keyword.

ego assignment is a special case of assignment where the LHS is `'ego'`. This is handled by adding a new statement that matches the syntax and giving it higher precedence than the normal `assignment`.

The `new` keyword behaves like a prefix operator and returns an expression. I added the `scenic_new_expr` rule under `expression`. 

### AST Module

Created a new module `scenic.ast` that contains AST nodes for Scenic. Currently it has three classes:

- `scenic.ast.AST` is a base class for all Scenic AST Nodes. By checking the inheritance of this class, can determine if the node is a Scenic AST node
- `scenic.ast.EgoAssign` is a class for representing `ego = ...`. It takes one argument for the value to assign.
- `scenic.ast.New` is a class for representing Scenic class instantiation. It has two fields: a) name of the class, b) specifiers. Specifiers will be a list of nodes, but it is not implemented yet.

### Compiler

`scenic.compiler` includes code for compiling Scenic AST to executable Python AST. The most important here is `ScenicToPythonTransformer`, which is a  `NodeTransformer`.

It has `visit_EgoAssign` and `visit_New` to compile `scenic.ast.EgoAssign` and `scenic.ast.New`, respectively.

### Translator

Changed `translator.py` to use the new parser and compiler to execute Scenic programs.

### Testing

Python 3.10 added [structural pattern matching](https://peps.python.org/pep-0634/). This PR (experimentally) includes test cases with this syntax. Overall, I like those test cases for better readability, but I see two problems with them.

First, running the test suite on Python 3.8 and 3.9 would raise a syntax error. While it is probably okay to only run unit tests of the parser and compiler on Python 3.10, we still need to be able to run the rest of the test cases on 3.8 and 3.9. It was possible by excluding files with pattern matching using [pytest_ignore_collect](https://docs.pytest.org/en/7.1.x/reference/reference.html#pytest.hookspec.pytest_ignore_collect). I will think a little more about the command-line interface and create a separate PR to implement it.

Second, each test case requires match-all case that asserts False, but I can imagine forgetting to put one. I wonder if there's a better way to use pattern-matching for writing test cases.


tests/syntax/test_compiler.py
```diff
@@ -0,0 +1,25 @@
...
+
+from scenic.syntax.compiler import compileScenicAST
+
+
+class TestCompiler:
+    # Special Case
+    def test_ego_assign(self):
+        node, _ = compileScenicAST(EgoAssign(Constant(1)))
+        match node:
+            case Expr(Call(func=Name(id=fn_name), args=args)):
```
dfremont, line 12: Can clean this up to `Expr(Call(Name("ego"), [Constant(1)]))` and get rid of the assertions (ditto for the similar tests below and in `test_parser.py`). For nodes with many fields it might be clearer to keep the names of the arguments (e.g. `Call(func=Name("ego"), args=[Constant(1)])`), but for nodes like these I don't think omitting the names will cause confusion.

dfremont (approved):
Looks good to me.

As we discussed "in person", one way to get rid of the `assert False` cases would be to have the matching cases `return True` and hook pytest (with a `conftest.py` in the appropriate folder) to check that all tests in these files return `True`. Looks like the easiest hook to use might be `pytest_pyfunc_call`: see an example [here](https://github.com/pytest-dev/pytest/issues/207).

src/scenic/syntax/translator.py
```diff
@@ -235,58 +237,39 @@ def compileStream(stream, namespace, params={}, model=None, filename='<stream>')
...
+			print('### End final AST')
+
+		if dumpASTPython:
+			try:
+				import astor
+			except ModuleNotFoundError as e:
+				raise RuntimeError('dumping the Python equivalent of the AST'
+									'requires the astor package')
+			print(f'### Begin Python equivalent of final AST of {filename}')
+			print(astor.to_source(tree))
```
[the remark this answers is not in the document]
shumbo: No, I unintentionally removed the option. I added the option back. fffa2c7c87b088b52016cd4341a2fbfbe439fe9e

tests/syntax/test_compiler.py
shumbo: fe04f4a732c0e8346076705d3359a3f864b500cd

I removed the argument names from `test_compiler.py`. Let me work on the same change for `test_parser.py` on another PR as it requires some work on Scenic AST classes to be able to pattern match without keywords (more specifically, I need to add `__match_args__` to every class).

dfremont: Ah, right. Not a high priority in any case, just something to keep in mind.

dfremont (approved):

PR #54, opened 2022-06-22
Adding objects that can replay completed scenarios

ameesh-shah:
*This is a working PR that is not to be immediately merged. Feedback will be gathered and used in future commits.*

The intention of this PR is to add the capability to "reconstitute" completed Scenarios, and be able to replay through the completed scenarios in a pythonic object. Additionally, new behaviors can be added to a ReplaySimulation object, and compared against the actions taken in the completed simulation. 

This PR does not yet support the transitioning from a ReplaySimulation object to a Simulation object, for the purposes of "picking up where you left off" and resuming an in-progress simulation.

Additional questions:
- How will new behaviors be added to the ReplaySimulation object? Right now, my intention is for them to be provided in a scene, but this is making the assumption that the list of behaviors and objects provided will be the same as the number of behaviors and objects that existed in the original simulation. I can add error handling to provide for this.
- Is it possible to save the 'scene' in the SimulationResult? Is this something we even want?
- It's still not clear what parts of run() should or shouldn't exist in the ReplaySimulation version.

Added fixes and changes to come.
@dfremont @ek65 let me know your feedback. Thanks!

### DIFF
--- a/src/scenic/core/simulators.py
+++ b/src/scenic/core/simulators.py
@@ -353,9 +353,10 @@ class Simulation:
         The definition of 'state' is up to the simulator; the 'state' is simply saved
         at each time step to define the 'trajectory' of the simulation.
 
-        The default implementation returns a tuple of the positions of all objects.
+        The default implementation a list of objects with their corresponding
+        information.
         """
-        return tuple(obj.position for obj in self.objects)
+        return self.objects
 
     @property
     def currentRealTime(self):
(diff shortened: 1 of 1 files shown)


ameesh-shah:
Note that the added test case is not final and will need adjustments to be pytest runnable.

src/scenic/core/simulators.py
```diff
@@ -353,9 +354,9 @@ def currentState(self):
         The definition of 'state' is up to the simulator; the 'state' is simply saved
         at each time step to define the 'trajectory' of the simulation.
 
-        The default implementation returns a tuple of the positions of all objects.
+        The default implementation a list of objects with their properties.
         """
-        return tuple(obj.position for obj in self.objects)
+        return [self.getProperties(obj, obj.properties) for obj in self.objects]
```
dfremont, line 359: We don't want to call `getProperties` again since it may be slow (e.g. actually talking to the simulator). I would just return `{prop: getattr(obj, prop) for prop in obj._dynamicProperties}`. Also the docstring should clarify that it's only *dynamic* properties that are saved (you can link to the glossary by writing ":term:\`dynamic properties\`").

PR #55, opened 2022-06-23
[Parser] Add Specifiers

shumbo:
## Overview

This PR adds specifiers to the new parser.

### Grammar

`scenic.gram` now contains grammar rules for specifiers. Currently, it does not support specifiers that span multiple lines, as it requires changes in the tokenizer. I added this to the TODO list.

### AST

Added AST nodes for representing specifiers. Each node now contains `__match_args__` so `test_parser.py` can pattern match on these nodes without specifying keywords for arguments. (See 9516b87feb957354091acc24fe6c4fd78e9f207c for details).

### Compiler

Added visitors for each specifier node. The visitor for `New` now handles specifiers.

## Notes

- `not visible` and `facing away from` are listed as specifiers in the documentation but are not implemented in the current version of the Scenic. I did not include these two specifiers in this PR.
- Most keywords are defined as soft keywords, so words like "left" "facing" and "following" can be used as variable names, etc. outside the specifier context.
  - See https://we-like-parsers.github.io/pegen/peg_parsers.html?highlight=soft#hard-and-soft-keywords for more information on hard and soft keywords
- I have not updated the existing tests yet because most of the cases in `test_specifiers` depend on some Scenic operators. I will do this once I have implemented enough features to be able to see results.

### DIFF
--- a/src/scenic/syntax/ast.py
+++ b/src/scenic/syntax/ast.py
@@ -153,3 +153,11 @@ class FacingSpecifier(AST):
     def __init__(self, heading: ast.AST, *args: any, **kwargs: any) -> None:
         super().__init__(*args, **kwargs)
         self.heading = heading
+
+
+class FacingTowardSpecifier(AST):
+    __match_args__ = ("position",)
+
+    def __init__(self, position: ast.AST, *args: any, **kwargs: any) -> None:
+        super().__init__(*args, **kwargs)
+        self.position = position
--- a/src/scenic/syntax/ast.py
+++ b/src/scenic/syntax/ast.py
@@ -115,6 +115,23 @@ class Behind(AST):
         super().__init__(*args, **kwargs)
 
 
+class BeyondSpecifier(AST):
+    __match_args__ = ("position", "offset", "base")
+
+    def __init__(
+        self,
+        position: ast.AST,
+        offset: ast.AST,
+        base: Optional[ast.AST] = None,
+        *args: any,
+        **kwargs: any
+    ) -> None:
+        super().__init__(*args, **kwargs)
+        self.position = position
+        self.offset = offset
+        self.base = base
+
+
 class VisibleSpecifier(AST):
     __match_args__ = ("base",)
 
--- a/src/scenic/syntax/ast.py
+++ b/src/scenic/syntax/ast.py
@@ -77,7 +77,7 @@ class OffsetAlongSpecifier(AST):
         self._fields = ["direction", "offset"]
 
 
-class PositionSpecifier(AST):
+class DirectionOfSpecifier(AST):
     __match_args__ = ("direction", "position", "distance")
 
     def __init__(
--- a/src/scenic/syntax/ast.py
+++ b/src/scenic/syntax/ast.py
@@ -12,9 +12,20 @@ class EgoAssign(AST):
         self.value = value
         self._fields = ["value"]
 
+# Instance Creation
+
 class New(AST):
     def __init__(self, className: str, specifiers: list, *args: any, **kwargs: any) -> None:
         super().__init__(*args, **kwargs)
         self.className = className
         self.specifiers = specifiers
         self._fields = ["className", "specifiers"]
+
+# Specifiers
+
+class WithSpecifier(AST):
+    def __init__(self, prop: str, value: ast.AST, *args: any, **kwargs: any) -> None:
+        super().__init__(*args, **kwargs)
+        self.prop = prop
+        self.value = value
+        self._fields = ["prop", "value"]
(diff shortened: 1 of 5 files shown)


src/scenic/syntax/ast.py
```diff
@@ -1,20 +1,187 @@
...
+    def __init__(
+        self, direction: ast.AST, offset: ast.AST, *args: any, **kwargs: any
+    ) -> None:
+        super().__init__(*args, **kwargs)
+        self.direction = direction
+        self.offset = offset
+        self._fields = ["direction", "offset"]
+
+
+class PositionSpecifier(AST):
```
dfremont, line 80: Let's give this a more precise name, since there are other "position" specifiers. Maybe `DirectionOfSpecifier`?

src/scenic/syntax/compiler.py
```diff
@@ -4,42 +4,162 @@
...
+        if isinstance(node.direction, s.LeftOf):
+            fn = "LeftSpec"
+        elif isinstance(node.direction, s.RightOf):
+            fn = "RightSpec"
+        elif isinstance(node.direction, s.AheadOf):
+            fn = "Ahead"
+        elif isinstance(node.direction, s.Behind):
+            fn = "Behind"
+        else:
+            raise TypeError(
```
dfremont, line 100: This should be unreachable unless there's a bug in the parser, right? So it should be an `assert` or something else indicating an internal error.

shumbo: Changed to assert
8713ad19a3c3f7d2ab58aa4bb60419c95502d797

src/scenic/syntax/ast.py
shumbo: Renamed it to `DirectionOfSpecifier` e69e2e7427200ae56b203cb6d982786b0ecf67a5

tests/syntax/test_compiler.py
```diff
@@ -1,4 +1,6 @@
 from ast import *
+
+import pytest
 from scenic.syntax.ast import *
```
dfremont, line 4: Let's move this down one line. I like to keep the imports separated into 3 blocks in general: standard Python libraries, other dependencies, then Scenic modules.

src/scenic/syntax/scenic.gram
```diff
@@ -1211,7 +1211,33 @@ inversion (memo):
...
+scenic_specifier:
+    | 'with' p=NAME v=expression { s.WithSpecifier(prop=p.string, value=v, LOCATIONS) }
+    | 'at' position=expression { s.AtSpecifier(position=position, LOCATIONS) }
+    | 'offset' 'by' o=expression { s.OffsetBySpecifier(offset=o, LOCATIONS) }
+    | 'offset' 'along' d=expression 'by' o=expression { s.OffsetAlongSpecifier(direction=d, offset=o, LOCATIONS) }
+    | direction=scenic_specifier_position_direction position=expression distance=['by' e=expression { e }] {
+        s.PositionSpecifier(direction=direction, position=position, distance=distance, LOCATIONS)
+     }
+    | "visible" b=['from' r=expression { r }] { s.VisibleSpecifier(base=b, LOCATIONS) }
+    | 'not' "visible" b=['from' r=expression { r }] { s.NotVisibleSpecifier(base=b, LOCATIONS) }
```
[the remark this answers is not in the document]
shumbo: Yes, I must have forgotten to remove it. Removed it for now. c59601ac7300bd0e5739ee72406df6f4f61f65de

tests/syntax/test_compiler.py
shumbo: 199cc9e52bf76f5308b0b7ebeec1b26cdcf2df76

Fixed the same issues for newly added files

src/scenic/syntax/ast.py
```diff
@@ -1,20 +1,187 @@
...
+        **kwargs: any
+    ) -> None:
+        super().__init__(*args, **kwargs)
+        self.direction = direction
+        self.position = position
+        self.distance = distance
+        self._fields = ["direction", "position", "distance"]
+
+
+class LeftOf(AST):
```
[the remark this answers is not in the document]
shumbo: On second thought, the reason CPython uses AST nodes might be that their AST nodes are auto-generated by ASDL definition and it does not support string literals.

https://github.com/python/cpython/blob/e69306f08b9be84ccdd0a1c6601ec229c4e5b377/Parser/Python.asdl
https://www.cs.princeton.edu/research/techreps/TR-554-97

Another difference I can think of is how NodeVisitor works on these classes.

```python
import ast

class V(ast.NodeVisitor):
    def visit_Not(self, node):
        print("found", ast.dump(node))

V().visit(ast.parse("not x"))
```

For example, the code above would print "found Not()". We can define `visit_Not` because Not is an AST node, and if it was some kind of ENUM representation (string literal, number, etc.), we would have to check `op` inside `visit_UnaryOp`.

I'd appreciate hearing your opinion.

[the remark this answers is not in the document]
dfremont: Ah, OK, if it's more uniform we can just keep them as AST nodes.

dfremont (approved):

dfremont (approved):

PR #56, opened 2022-06-25
[Parser] Implement Operators

shumbo:
## Overview

This PR adds Scenic operators to the new parser.

### Prefix Operators

Prefix operators are operators that begin with some keywords, not expressions. This PR implements the following prefix operators.

- `relative position of <vector> [from <vector>]`
  - not documented but exists in implementation
- `relative heading of <heading> [from <heading>]`
- `apparent heading of <oriented point> [from <vector>]`
- `distance from {X} to {Y}`
  - One of `X` and `Y` is required. If only one of the two is given, the other will have the value `ego`
  - `from` and `to` are interchangeable. `distance from X to Y` and `distance to X from Y` are both valid
  - The implementation assumes this operation is commutative.
- `distance past {vector} of {OP}`
- `angle to {X} from {Y}`
  - Similarly to the `distance from` operator, one of the two operands can be omitted and `ego` will be used
  - `from` and `to` are interchangeable.
  - This operation is NOT commutative, so we must remember which operand was specified with `from` and `to`
- `follow <field> from <vector> for <number>`
- `visible <region>`
- `not visible <region>`
  - not documented but implemented in code
  - requires changes to the Python grammar to disambiguate with the inversion of the variable `visible`
- `(front | back | left | right) of Object`
- `(front | back) (left | right) of Object`

In Scenic, prefix operators have the lowest precedence. In other words, everything that comes after the keywords will be an operand of the operator. For this reason, `scenic_prefix_operators` that contains grammar rules for all prefix operators are placed in between `factor` and `atom`. 

### Infix/Postfix Operators

Other operators begin with some expressions followed by keywords. With some exceptions, most infix operators in Scenic are implemented by substituting unused Python operators. When implementing a parser, we want to preserve the same precedence as the operator used for substitution.

All operators are left associative, which also matches the current implementation.

- `<number> deg`
  - Precedence: the same as multiplication (`*`)
  - The old compiler simply replaces the occurrence of `deg` with `* 0.01745329`
- `{X} @ {Y}`
  - override the Python's matrix multiplication operator with a rule with higher precedence
- `X relative to Y`
  - Precedence: the same as bitwise and (`&`)
- `X offset along H by Y`
  - Precedence: the same as bitwise xor (`^`)
- `X at Y`
  - Precedence: the same as left shift (`<<`)
- `X can see Y`
  - Precedence: the same as bitwise or (`|`)

### DIFF
--- a/src/scenic/syntax/ast.py
+++ b/src/scenic/syntax/ast.py
@@ -400,6 +400,86 @@ class NotVisibleOp(AST):
         self._fields = ["region"]
 
 
+class PositionOfOp(AST):
+    __match_args__ = ("position", "target")
+
+    def __init__(
+        self,
+        position: Union[
+            "Front",
+            "Back",
+            "Left",
+            "Right",
+            "FrontLeft",
+            "FrontRight",
+            "BackLeft",
+            "BackRight",
+        ],
+        target: ast.AST,
+        *args: any,
+        **kwargs: any
+    ) -> None:
+        super().__init__(*args, **kwargs)
+        self.position = position
+        self.target = target
+
+
+class Front(AST):
+    "Represents position of `front of` operator"
+
+    def __init__(self, *args: any, **kwargs: any) -> None:
+        super().__init__(*args, **kwargs)
+
+
+class Back(AST):
+    "Represents position of `back of` operator"
+
+    def __init__(self, *args: any, **kwargs: any) -> None:
+        super().__init__(*args, **kwargs)
+
+
+class Left(AST):
+    "Represents position of `left of` operator"
+
+    def __init__(self, *args: any, **kwargs: any) -> None:
+        super().__init__(*args, **kwargs)
+
+
+class Right(AST):
+    "Represents position of `right of` operator"
+
+    def __init__(self, *args: any, **kwargs: any) -> None:
+        super().__init__(*args, **kwargs)
+
+
+class FrontLeft(AST):
+    "Represents position of `front left of` operator"
+
+    def __init__(self, *args: any, **kwargs: any) -> None:
+        super().__init__(*args, **kwargs)
+
+
+class FrontRight(AST):
+    "Represents position of `front right of` operator"
+
+    def __init__(self, *args: any, **kwargs: any) -> None:
+        super().__init__(*args, **kwargs)
+
+
+class BackLeft(AST):
+    "Represents position of `back left of` operator"
+
+    def __init__(self, *args: any, **kwargs: any) -> None:
+        super().__init__(*args, **kwargs)
+
+
+class BackRight(AST):
+    "Represents position of `back right of` operator"
+
+    def __init__(self, *args: any, **kwargs: any) -> None:
+        super().__init__(*args, **kwargs)
+
+
 class VectorOp(AST):
     __match_args__ = ("left", "right")
 
--- a/src/scenic/syntax/ast.py
+++ b/src/scenic/syntax/ast.py
@@ -378,3 +378,11 @@ class VisibleOp(AST):
         super().__init__(*args, **kwargs)
         self.region = region
         self._fields = ["region"]
+
+class NotVisibleOp(AST):
+    __match_args__ = ("region",)
+
+    def __init__(self, region: ast.AST, *args: any, **kwargs: any) -> None:
+        super().__init__(*args, **kwargs)
+        self.region = region
+        self._fields = ["region"]
--- a/src/scenic/syntax/ast.py
+++ b/src/scenic/syntax/ast.py
@@ -330,8 +330,9 @@ class DistanceFromOp(AST):
 
     def __init__(
         self,
-        target: ast.AST,  # to
-        base: Optional[ast.AST] = None,  # from
+        # because `to` and `from` are symmetric, the first operand will be `target` and the second will be `base`
+        target: ast.AST,
+        base: Optional[ast.AST] = None,
         *args: any,
         **kwargs: any
     ) -> None:
--- a/src/scenic/syntax/ast.py
+++ b/src/scenic/syntax/ast.py
@@ -351,3 +351,21 @@ class AngleFromOp(AST):
         self.target = target
         self.base = base
         self._fields = ["target", "base"]
+
+
+class FollowOp(AST):
+    __match_args__ = ("target", "base", "distance")
+
+    def __init__(
+        self,
+        target: ast.AST,
+        base: ast.AST,
+        distance: ast.AST,
+        *args: any,
+        **kwargs: any
+    ) -> None:
+        super().__init__(*args, **kwargs)
+        self.target = target
+        self.base = base
+        self.distance = distance
+        self._fields = ["target", "base", "distance"]
--- a/src/scenic/syntax/ast.py
+++ b/src/scenic/syntax/ast.py
@@ -402,10 +402,22 @@ class NotVisibleOp(AST):
 
 class VectorOp(AST):
     __match_args__ = ("left", "right")
+
     def __init__(
         self, left: ast.AST, right: ast.AST, *args: any, **kwargs: any
     ) -> None:
         super().__init__(*args, **kwargs)
         self.left = left
         self.right = right
-        self._fields=["left", "right"]
+        self._fields = ["left", "right"]
+
+
+class FieldAtOp(AST):
+    __match_args__ = ("left", "right")
+
+    def __init__(
+        self, left: ast.AST, right: ast.AST, *args: any, **kwargs: any
+    ) -> None:
+        super().__init__(*args, **kwargs)
+        self.left = left
+        self.right = right
--- a/src/scenic/syntax/ast.py
+++ b/src/scenic/syntax/ast.py
@@ -299,3 +299,15 @@ class RelativeHeadingOp(AST):
         self.target = target
         self.base = base
         self._fields = ["target", "base"]
+
+class ApparentHeadingOp(AST):
+    __match_args__ = ("target", "base")
+
+    def __init__(
+        self, target: ast.AST, base: ast.AST = None, *args: any, **kwargs: any
+    ) -> None:
+        super().__init__(*args, **kwargs)
+        self.target = target
+        self.base = base
+        self._fields = ["target", "base"]
+
--- a/src/scenic/syntax/ast.py
+++ b/src/scenic/syntax/ast.py
@@ -398,3 +398,14 @@ class NotVisibleOp(AST):
         super().__init__(*args, **kwargs)
         self.region = region
         self._fields = ["region"]
+
+
+class VectorOp(AST):
+    __match_args__ = ("left", "right")
+    def __init__(
+        self, left: ast.AST, right: ast.AST, *args: any, **kwargs: any
+    ) -> None:
+        super().__init__(*args, **kwargs)
+        self.left = left
+        self.right = right
+        self._fields=["left", "right"]
--- a/src/scenic/syntax/ast.py
+++ b/src/scenic/syntax/ast.py
@@ -298,6 +298,7 @@ class RelativePositionOp(AST):
         super().__init__(*args, **kwargs)
         self.target = target
         self.base = base
+        self._fields = ["target", "base"]
 
 
 class RelativeHeadingOp(AST):
--- a/src/scenic/syntax/ast.py
+++ b/src/scenic/syntax/ast.py
@@ -286,3 +286,16 @@ class ApparentlyFacingSpecifier(AST):
         self.heading = heading
         self.base = base
         self._fields = ["heading", "base"]
+
+
+# Operators
+class RelativeHeadingOp(AST):
+    __match_args__ = ("target", "base")
+
+    def __init__(
+        self, target: ast.AST, base: ast.AST = None, *args: any, **kwargs: any
+    ) -> None:
+        super().__init__(*args, **kwargs)
+        self.target = target
+        self.base = base
+        self._fields = ["target", "base"]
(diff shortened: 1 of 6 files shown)


src/scenic/syntax/compiler.py
```diff
@@ -43,3 +52,156 @@ def visit_New(self, node: s.New):
...
+        assert f is not None, f"impossible position {node.position.__class__.__name__}"
+        return ast.Call(
+            func=ast.Name(id=f, ctx=loadCtx),
+            args=[self.visit(node.target)],
+            keywords=[],
+        )
+
+    def visit_DegOp(self, node: s.DegOp):
+        return ast.BinOp(
+            left=self.visit(node.operand), op=ast.Mult(), right=ast.Constant(0.01745329)
```
shumbo, line 164: I wonder if having a magic number here is a good idea. For example, we can replace the magic number with `math.radian(1)`

Or better, we can make it that this visitor replaces the node to a call to `math.radian` with the operand as an argument. Because of how `math.radian` is implemented, this will not change the result at all. https://github.com/python/cpython/blob/605e9c66ad367b54a847f9fc65447a071742f554/Modules/mathmodule.c#L2938-L2943

dfremont: Agreed, it's clearer to call `math.radians`. This could break if someone bound the name `math`, but maybe we should put off fixing that for now as the same problem occurs with other names imported from the veneer: e.g. if someone defined something called `CanSee` it would break the `can see` operator. Probably we should move all such names into attributes of some name which is unlikely to collide, e.g. `__scenic.CanSee`.

tests/syntax/test_parser.py
```diff
@@ -184,13 +184,13 @@ def test_specifier_beyond(self):
                 assert False
 
     def test_specifier_beyond_from(self):
-        mod = parse_string_helper("new Object beyond position by distance from base")
```
shumbo, line 187: The old test now means `new Object beyond position by (distance from base)` where `distance from` is a prefix operator

src/scenic/syntax/ast.py
```diff
@@ -202,3 +202,260 @@ def __init__(
...
+        super().__init__(*args, **kwargs)
+        self.left = left
+        self.right = right
+
+
+class OffsetAlongOp(AST):
+    __match_args__ = ("left", "middle", "right")
+
+    def __init__(
+        self, left: ast.AST, middle: ast.AST, right: ast.AST, *args: any, **kwargs: any
```
shumbo, line 445: This node represents `X offset along H by Y`. I wanted to give meaningful names to the parameters but I couldn't come up with good ones. Any ideas?

src/scenic/syntax/compiler.py
shumbo: It turns out I cannot use `math` inside the compiled Python AST because it is not included. If I want to use `math` I need to either

1. Import `math` from `veneer`, write a `Deg` function that wraps `math.radian` and call that function, or
2. Add `import math` (or similar) to [`preamble`](https://github.com/BerkeleyLearnVerify/Scenic/blob/952609262e4f9178b1a5c6956c6ba0de7dbfe776/src/scenic/syntax/translator.py#L310-L315)

both of which I think go beyond the scope of this PR. So I will leave it as is for now and added to the TODO list.

It feels like `from scenic.syntax.veneer import *` is the root cause of the name collision issue you mentioned. We can change it to something like `from scenic.syntax.veneer import __scenic`, and then we can add a `__scenic.` to every function call that the compiler generates. I will have to see if this change has side effects, but looks promising to me. Also added this to the TODOs.

dfremont: OK. Yes, just importing a single name from the veneer and routing everything through that is what I had in mind. It should be OK since we only ever read from the imported names, never rebind them.

src/scenic/syntax/ast.py
dfremont: `base`, `direction`, and `offset`?

src/scenic/syntax/compiler.py
```diff
@@ -171,3 +172,158 @@ def visit_ApparentlyFacingSpecifier(self, node: s.ApparentlyFacingSpecifier):
...
+            args=[self.visit(node.target)],
+            keywords=[]
+            if node.base is None
+            else [ast.keyword(arg="Y", value=self.visit(node.base))],
+        )
+
+    def visit_DistanceFromOp(self, node: s.DistanceFromOp):
+        assert node.target is not None or node.base is not None
+        target = ego if node.target is None else self.visit(node.target)
+        base = ego if node.base is None else self.visit(node.base)
```
dfremont, line 208: If `ego` is still a function in the veneer, don't these need to be calls rather than just names?

In any case, for most of the operators which use ego implicitly, you used the approach of passing them an optional keyword argument when specified. Why not do that here as well? The `DistanceFrom(X, Y)` function in the veneer has a default value of `None` for `Y`, which it then replaces with the ego.

src/scenic/syntax/compiler.py
```diff
@@ -171,3 +172,158 @@ def visit_ApparentlyFacingSpecifier(self, node: s.ApparentlyFacingSpecifier):
...
+            if node.base is None
+            else [ast.keyword(arg="Y", value=self.visit(node.base))],
+        )
+
+    def visit_AngleFromOp(self, node: s.AngleFromOp):
+        assert (
+            node.base is not None or node.target is not None
+        ), "neither target nor base were specified in AngleFromOp"
+        base = ego if node.base is None else self.visit(node.base)
+        target = ego if node.target is None else self.visit(node.target)
```
dfremont, line 229: As for `distance from`, for consistency maybe we should push the implicit ego behavior into the implementation in the veneer (although in this case you would have to change `AngleFrom` since it doesn't currently have an optional argument).

src/scenic/syntax/compiler.py
```diff
@@ -171,3 +172,158 @@ def visit_ApparentlyFacingSpecifier(self, node: s.ApparentlyFacingSpecifier):
...
+
+    def visit_NotVisibleOp(self, node: s.VisibleOp):
+        return ast.Call(
+            func=ast.Name(id="NotVisible", ctx=loadCtx),
+            args=[self.visit(node.region)],
+            keywords=[],
+        )
+
+    def visit_PositionOfOp(self, node: s.PositionOfOp):
+        classToFunctionName = {
```
dfremont, line 262: Rather than have this table, which would need to be kept in sync with the node definitions, what about adding a `functionName` attribute to the node classes? That would then only need to be kept in sync with the veneer. (And we could put in a sanity check to make sure the functions exist there, like we currently do in `translator.py`.)

src/scenic/syntax/scenic.gram
```diff
@@ -1206,7 +1206,8 @@ conjunction (memo):
     | inversion
 
 inversion (memo):
-    | 'not' a=inversion { ast.UnaryOp(op=ast.Not(), operand=a, LOCATIONS) }
+    # [SCENIC NOTE]: Fail `not visible` to be handled later
+    | 'not' !"visible" a=inversion { ast.UnaryOp(op=ast.Not(), operand=a, LOCATIONS) }
```
dfremont, line 1210: This makes `visible` a hard keyword in practice, since you couldn't use it as a boolean variable and then write `x = not visible`. To get better error messages we should probably just make it an actual hard keyword if we go this route; alternatively, we could consider changing the syntax `not visible` to `invisible` (but I think it could be confusing this way).

src/scenic/syntax/scenic.gram
```diff
@@ -1278,22 +1279,85 @@ is_bitwise_or: 'is' a=bitwise_or { (ast.Is(), a) }
 # -----------------
 
 bitwise_or:
+    | scenic_bitwise_or
     | a=bitwise_or '|' b=bitwise_xor { ast.BinOp(left=a, op=ast.BitOr(), right=b, LOCATIONS) }
     | bitwise_xor
 
+scenic_bitwise_or:
```
dfremont, line 1286: These rule names are confusing to me, since there are no bitwise operators in Scenic. Why not call this `scenic_can_see`, and similarly for the other operators?

tests/syntax/test_parser.py
dfremont: Hmm, I would like `distance` to be a valid variable name, if possible. Can we set up the precedence so that this parses as `new Object beyond (position) by (distance) from (base)`, or would that break other natural cases?

tests/syntax/test_parser.py
```diff
@@ -336,3 +336,849 @@ def test_specifier_apparently_facing_from(self):
...
+        stmt = mod.body[0]
+        match stmt:
+            case Expr(RelativeHeadingOp(Name("x"), Name("y"))):
+                assert True
+            case _:
+                assert False
+
+    def test_relative_heading_precedence(self):
+        # This is incorrect code, but asserts the operator has the right precedence
+        mod = parse_string_helper("relative heading of x or y")
```
dfremont, line 451: We ultimately want this to parse as `(relative heading of x) or y`, though; are you planning to do that in a later PR, preserving the current (bad) behavior of the `TokenTranslator` for now?

tests/syntax/test_parser.py
```diff
@@ -336,3 +336,849 @@ def test_specifier_apparently_facing_from(self):
...
+                    LShift(),
+                    Name("F", Load()),
+                ),
+            ),
+        ],
+    )
+    def test_follow_precedence(self, code, expected):
+        mod = parse_string_helper(code)
+        stmt = mod.body[0].value
+        assert dump(stmt, annotate_fields=False) == dump(
```
dfremont, line 952: This approach of using `dump` to compare the ASTs is used frequently in this file, and I'm a little worried it might not be stable enough for this purpose, since `dump` is intended for debugging. Can we define a helper function `compareASTs` or something so that we can adjust all of these tests uniformly in one place if needed later? It would also simplify the code.

dfremont (changes requested):
Looks good overall -- just a few minor comments.

src/scenic/syntax/scenic.gram
shumbo: Here, I meant "scenic operators with the same precedence as bitwise or". For example, we _can_ think of adding more operators with the same precedence (like `+` and `-`).

However, now I think about it, there won't be such cases because we usually want one operator to bound tighter than another. I will change those names to more specific ones.

tests/syntax/test_parser.py
shumbo: Are you sure `(relative heading of x) or y` is the desired result of this program? The current setup is that `relative heading of` has precedence higher than bitwise shifts and lower than addition/subtraction. 

https://docs.python.org/3/reference/expressions.html#operator-precedence

Because `or` has further lower precedence than shifts, currently it parses as `relative heading of (x or y)`. 

Maybe we should create a table and agree on the desired precedence of each of the operators. Changing the precedence itself should be relatively easy (just move rules in the grammar and adjust test cases)

src/scenic/syntax/compiler.py
shumbo: You're right that the default operands should be defined inside the veneer. I changed it so that 

- if there is only one operand (either `from` or `to`), use it as `X`
- if there are two operands (both `from` and `to`), use `to` as `X` and `from` as `Y`

b78f733d00cdfbac0f9f251d90a329c719d3153a

src/scenic/syntax/compiler.py
shumbo: I changed `AngleFrom` and made both arguments optional. If missing, `X` and `Y` will have `None` which will be replaced with `ego()`.

61e2e26f0471916125b9d34693bcd55470c02e78

src/scenic/syntax/compiler.py
shumbo: fixed in c2e86eb0a062f0663f70b2675800ec0d94450ea9

src/scenic/syntax/scenic.gram
shumbo: Fixed in 98f59d79489e08bab00cd62615b3fa2a1ab3ef8f

tests/syntax/test_parser.py
dfremont: Per our discussion, we'll leave the behavior as it is for now; users will need to introduce parentheses if they want to split `distance` from `from base`.

src/scenic/syntax/scenic.gram
```diff
@@ -1279,22 +1280,81 @@ is_bitwise_or: 'is' a=bitwise_or { (ast.Is(), a) }
...
+    | "visible" e=scenic_prefix_operators { s.VisibleOp(region=e, LOCATIONS) }
+    | 'not' "visible" e=scenic_prefix_operators { s.NotVisibleOp(region=e, LOCATIONS) }
+    | p=scenic_position_of_op_position 'of' e=scenic_prefix_operators { s.PositionOfOp(position=p, target=e, LOCATIONS) }
     | sum
 
+scenic_distance_from_op:
+    | "distance" 'from' e1=expression 'to' e2=scenic_prefix_operators { s.DistanceFromOp(base=e1, target=e2, LOCATIONS) }
+    | "distance" 'to' e1=expression 'from' e2=scenic_prefix_operators { s.DistanceFromOp(target=e1, base=e2, LOCATIONS) }
+    | "distance" 'to' e1=scenic_prefix_operators { s.DistanceFromOp(target=e1, LOCATIONS) }
+    | "distance" 'from' e2=scenic_prefix_operators { s.DistanceFromOp(base=e2, LOCATIONS) }
```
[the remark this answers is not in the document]
shumbo: Fixed in c13cb337c2df2b2dcc4a39a084e962e3dae845b0 and 62df366cb8e436b5fbd517f1e72c028a957c7c7a 

https://github.com/BerkeleyLearnVerify/Scenic/compare/dac4b91c90026ad138f96936a4f997d3358447ca..62df366cb8e436b5fbd517f1e72c028a957c7c7a

dfremont (approved):

PR #57, opened 2022-06-28
Sensors

theilem:
This code adds the "sensors" capability to scenic which are read every time step, with the result written into an "observations" dictionary of the object the sensor is attached to. Additionally, the sensors can be logged automatically.


Eric-Vin:
Closing this as it is superseded by https://github.com/BerkeleyLearnVerify/Scenic/pull/195

PR #58, opened 2022-06-30
[Parser] implement `param` statements

shumbo:
This PR adds `param identifier=value, ...` statements.

- The grammar matches the syntax
  - identifiers can be a NAME (without quotes) or STRING (with quotes)
- `Param` AST Node represents a single `param` statement
- `parameter` AST Node represents a identifier-value pair
- Compiler converts a list of parameters into a dictionary
  - If the list contains two or more definitions, the last definition overwrites the previous ones
- Changed the signature of `param` to take a single dictionary whose keys are identifiers and values are values.


src/scenic/syntax/ast.py
```diff
@@ -30,7 +55,7 @@ def __init__(
     ) -> None:
         super().__init__(*args, **kwargs)
         self.className = className
-        self.specifiers = specifiers
+        self.specifiers = specifiers if specifiers is not None else []
```
shumbo, line 58: Because specifiers can be empty, and the grammar rule generates `None` in such cases, we need to manually set it to the empty list if `specifiers` is None

src/scenic/syntax/compiler.py
```diff
@@ -43,6 +43,23 @@ def visit_EgoAssign(self, node: s.EgoAssign):
...
+        d = dict()
+        for parameter in node.elts:
+            d[parameter.identifier] = self.visit(parameter.value)
+        return ast.Expr(
+            value=ast.Call(
+                func=ast.Name(id="param", ctx=loadCtx),
+                args=[
+                    ast.Dict(
+                        [ast.Constant(k) for k in d.keys()],
+                        [v for v in d.values()],
```
shumbo, line 56: `d.values()` returns [a dictionary view object](https://docs.python.org/3.7/library/stdtypes.html?highlight=dict%20values#dictionary-view-objects), but the AST takes a list. Here, I'm converting it to a list.

dfremont: It'd be clearer to just write `list(d.values())` I think -- unless you're trying to mirror the syntax on the previous line?

dfremont (changes requested):
Looks good, except I think we should raise a syntax error if a parameter name is repeated: this can't ever be useful, so it must be a mistake and we should warn the user.

src/scenic/syntax/compiler.py
shumbo: Fixed in 41573b2572e0cf7246e1fc45ad57819c57ea5b3d

shumbo:
44a17908636123df8306a94ffc171c87bd6fad26

If the same parameter name is given twice, that will be an error.

dfremont (approved):

PR #60, opened 2022-07-10
[Parser] implement `model` statements

shumbo:
This PR implements `model ...` statements.

- The grammar is defined using `dotted_name` which is used for `import` statements
- model is a soft keyword because otherwise we wouldn't be able to parse `model scenic.simulators.carla.model`
  - the last `model` is a token that's not part of the statement

### DIFF
--- a/tests/syntax/test_parser.py
+++ b/tests/syntax/test_parser.py
@@ -37,6 +37,26 @@ class TestEgoAssign:
             parse_string_helper("ego, x = 10, 20")
 
 
+class TestModel:
+    def test_basic(self):
+        mod = parse_string_helper("model some_model")
+        stmt = mod.body[0]
+        match stmt:
+            case Model("some_model"):
+                assert True
+            case _:
+                assert False
+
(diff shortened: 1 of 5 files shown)


tests/syntax/test_parser.py
```diff
@@ -37,6 +37,26 @@ def test_ego_keyword(self):
...
+        mod = parse_string_helper("model some_model")
+        stmt = mod.body[0]
+        match stmt:
+            case Model("some_model"):
+                assert True
+            case _:
+                assert False
+
+    def test_dotted(self):
+        mod = parse_string_helper("model scenic.simulators.carla")
```
dfremont, line 51: Why not have this use `scenic.simulators.carla.model` to check the soft-keywordness of `model`?

shumbo: Not sure why I omitted the most important part. Fixed in 9d130fd7daf0c89e063699c130d40b024f6d5e0e

dfremont (approved):

PR #64, opened 2022-07-14
[Parser] Support class definitions and properties definitions

shumbo:
## Overview

This PR implements the new strategy to compile Scenic class definitions and property definitions.

An overview of the strategy can be found in [my note](https://shun-notes.notion.site/How-to-compile-class-definitions-482061588cd14ec091c51e74d7eaba8b).

## Change in Grammar

Several changes were made to the grammar to generate AST nodes that represent property definitions.

In Scenic, properties can be added to the class using the following syntax.

```python
class MyScenicClass(Object):
    # without attributes
    property: value
    # with attributes
    property[additive]: value
```

Although the code above is syntactically valid in Python, it has a different semantic, and we want to generate different types of AST nodes.

This PR adds `PropertyDef` for this purpose. `PropertyDef` has three fields: `property` for the name of the property, `attributes` for meta attributes and `value` for the default value of the property.

Note that `PropertyDef` can only be at the top level of the class definition body. For example, the following code would not produce `PropertyDef` and instead generate Python's `AnnAssign`.

```python
class MyScenicClass(Object):
    def method():
        property: value # -> AnnAssign to the variable `property`
        property[additive]: value # -> AnnAssign to `property` subscript `additive`
```

To achieve this, I added `scenic_class_def_block`, `scenic_class_statements` and `scenic_class_statement` to the grammar. They are almost identical to the normal `block` `statements` `statement`, but in addition to all statements, it allows `scenic_class_property_stmt` with the highest precedence. By having `scenic_class_def_block` for class body, property definitions are parsed as `PropertyDef`.

Only the two literal `additive` and `dynamic` can be inside the square brackets (with a comma to separate the two if need to specify both). If a different literal is specified, that raises a syntax error.

## Compiling `ClassDef` `PropertyDef`

With this PR, we are moving away from passing default values using annotations. Instead of annotations, the compiler creates a special dictionary `_scenic_properties` that stores all properties and corresponding values for a class.

As stated, `PropertyDef` can only be in a body of `ClassDef`. On `visit_ClassDef`, we extract all property definitions and turn them into a single dictionary declaration.

For example, 

```python
class MyScenicClass(Object):
    property1: value1
    property2: value2
```

would be compiled to

```python
class MyScenicClass(Object):
    _scenic_properties = {
        "property1": PropertyDefault(set(), set(), lambda self: value1),
        "property2": PropertyDefault(set(), set(), lambda self: value2),
    }
```

Because we do this transformation in `ClassDef`, `visit_PropertyDef` should not be called. If the function is called, it will raise an assertion error.

`visit_ClassDef` also adds `Object` for the base class if no base classes were specified.

## Reading property values

`object_types.py` was changed to read property values from `_scenic_properties` instead of `__annotations__`.

The file has definitions of the built-in types (`Object`, `Point`, `OrientedPoint`). They were rewritten to define properties using `_scenic_properties`.

Finally, `__init__subclass__` on `Constructible` now asserts that no more than one Scenic class was given as the base class to enforce single inheritance of Scenic classes.

## `new` veneer function

The `new` keyword now asserts that the given class is a Scenic class (i.e., the subclass of `Constructible`).

The `new` keyword is compiled to a call to the `new` veneer function. It performs the check, raises an exception if the check fails, and returns an instance.

### DIFF
--- a/src/scenic/syntax/scenic.gram
+++ b/src/scenic/syntax/scenic.gram
@@ -726,11 +726,12 @@ scenic_class_property_stmt:
         )
      }
 
-scenic_class_property_attribute:
-    | "additive" { s.Additive(LOCATIONS) }
+# fail if `NAME [ <expr> ]` pattern is found at top level of class definition and
+# <expr> is neither `additive` nor `dynamic`
+scenic_class_property_attribute: &&(
+      "additive" { s.Additive(LOCATIONS) }
     | "dynamic" { s.Dynamic(LOCATIONS) }
-    # if '['' follows NAME at class definition top level and what follows is neither "additive" nor "dynamic", raise an error
-    | a=NAME { self.raise_syntax_error_known_location('unknown attribute "' + a.string + '"', a) }
+)
 
 # Function definitions
 # --------------------
--- a/src/scenic/syntax/scenic.gram
+++ b/src/scenic/syntax/scenic.gram
@@ -717,7 +717,7 @@ scenic_class_statement[list]:
 
 scenic_class_property_stmt:
     # not a simple statement; reads NEWLINE
-    | a=NAME b=['[' attrs=','.NAME+ ']' { [a.string for a in attrs] } ] ':' c=expression NEWLINE { 
+    | a=NAME b=['[' attrs=','.scenic_class_property_attribute+ ']' { attrs } ] ':' c=expression NEWLINE { 
         s.PropertyDef(
             property=a.string,
             attributes=b if b is not None else [],
@@ -725,6 +725,12 @@ scenic_class_property_stmt:
             LOCATIONS,
         )
      }
+
+scenic_class_property_attribute:
+    | "additive" { s.Additive(LOCATIONS) }
+    | "dynamic" { s.Dynamic(LOCATIONS) }
+    # if '['' follows NAME at class definition top level and what follows is neither "additive" nor "dynamic", raise an error
+    | a=NAME { self.raise_syntax_error_known_location('unknown attribute "' + a.string + '"', a) }
 
 # Function definitions
 # --------------------
--- a/src/scenic/syntax/scenic.gram
+++ b/src/scenic/syntax/scenic.gram
@@ -691,7 +691,7 @@ class_def[ast.ClassDef]:
 
 class_def_raw[ast.ClassDef]:
     | invalid_class_def_raw
-    | 'class' a=NAME b=['(' z=[arguments] ')' { z }] &&':' c=block {
+    | 'class' a=NAME b=['(' z=[arguments] ')' { z }] &&':' c=scenic_class_def_block {
         ast.ClassDef(
             a.string,
             bases=b[0] if b else [],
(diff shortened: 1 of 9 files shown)


tests/syntax/test_classes.py
```diff
@@ -72,13 +61,13 @@ class Bar(Foo):
 def test_isinstance_issubclass():
     scenario = compileScenic("""
         class Foo: pass
-        ego = Foo
+        ego = new Foo
         if isinstance(ego, Foo):
```
shumbo, line 65: needs #62 to correctly compile `ego` and this case doesn't pass at the moment.

tests/syntax/test_classes.py
```diff
@@ -46,9 +35,9 @@ class Foo:\n
 def test_property_simple():
     scenario = compileScenic("""
         class Foo:
-            position: 3 @ 9
+            position: (3, 9)
```
shumbo, line 38: `@` needs #56, so changing it to tuples. Can revert the change once #56 is merged.

src/scenic/syntax/scenic.gram
```diff
@@ -701,6 +701,37 @@ class_def_raw[ast.ClassDef]:
...
+            property=a.string,
+            attributes=b if b is not None else [],
+            value=c,
+            LOCATIONS,
+        )
+     }
+
+# fail if `NAME [ <expr> ]` pattern is found at top level of class definition and
+# <expr> is neither `additive` nor `dynamic`
+scenic_class_property_attribute: &&(
```
shumbo, line 730: [`&&`](https://we-like-parsers.github.io/pegen/grammar.html#e-6) is an operator that fails immediately if the expression after that fails to parse.

If what follows after `[` is not `additive` or `dynamic`, it raises a syntax error.
Screen Shot 2022-07-14 at 9 08 17

shumbo: I first wrote a rule like the following:

```
scenic_class_property_attribute:
    | "additive" { s.Additive(LOCATIONS) }
    | "dynamic" { s.Dynamic(LOCATIONS) }
    | !NAME
```

However, if you put a non-Name expression inside the square brackets (like quoted strings or numbers), this rule would fail but will be parsed as a normal annotated assignment.

We can list possible tokens and fail all of them, but I think it would be better to use `&&` here.

tests/syntax/test_classes.py
```diff
@@ -5,39 +5,28 @@
...
 
 def test_old_constructor_statement():
-    compileScenic("""
-        constructor Foo:
-            blah: 19 @ -3
-        ego = Foo with blah 12
-    """)
+    with pytest.raises(SyntaxError):
+        compileScenic("""
+            constructor Foo:
```
shumbo, line 12: the new parser does not support the old `constructor` keyword. Use of the keyword would result in a syntax error.

src/scenic/syntax/scenic.gram
dfremont: Agreed.

dfremont (approved):
Looks good to me!

PR #66, opened 2022-07-23
[Parser] Fix Static Test Syntax + Small Bug Fixes

shumbo:
This PR updates the syntax of test files under `tests/syntax` to use the `new` keyword.

In addition, it has minor bug fixes (~5 lines) to the parser and compiler. I left comments about those changes in this PR.

The following files include tests with unimplemented syntax, and I did not modify those test cases in this PR.

- `test_dynamics`
- `test_modular`
- `test_errors.py`
  - partially, most cases are passing

To ignore those files, we run `pytest ./tests/syntax --ignore "./tests/syntax/test_dynamics.py" --ignore "./tests/syntax/test_errors.py" --ignore "./tests/syntax/test_modular.py"`. It revealed several major missing parts in the new parser/compiler, which are:

- specifiers that span across multiple lines
- `not visible` specifier
  - this was missing and added to the test suite after I merged #55 
  - PR: #67 
- unpacking distributions

I will work on these on separate branches.

### DIFF
--- a/src/scenic/syntax/compiler.py
+++ b/src/scenic/syntax/compiler.py
@@ -218,7 +218,7 @@ class ScenicToPythonTransformer(ast.NodeTransformer):
                     ast.Constant(value=node.lineno),
                     ast.Constant(value=node.name),
                 ],
-                keywords=[ast.keyword(arg="prob", value=node.prob)]
+                keywords=[ast.keyword(arg="prob", value=ast.Constant(node.prob))]
                 if node.prob is not None
                 else [],
             )
--- a/src/scenic/syntax/scenic.gram
+++ b/src/scenic/syntax/scenic.gram
@@ -1256,8 +1256,8 @@ scenic_specifiers: ss=','.scenic_specifier+ { ss }
 scenic_specifier:
     | 'with' p=NAME v=expression { s.WithSpecifier(prop=p.string, value=v, LOCATIONS) }
     | 'at' position=expression { s.AtSpecifier(position=position, LOCATIONS) }
-    | 'offset' 'by' o=expression { s.OffsetBySpecifier(offset=o, LOCATIONS) }
-    | 'offset' 'along' d=expression 'by' o=expression { s.OffsetAlongSpecifier(direction=d, offset=o, LOCATIONS) }
+    | "offset" 'by' o=expression { s.OffsetBySpecifier(offset=o, LOCATIONS) }
+    | "offset" 'along' d=expression 'by' o=expression { s.OffsetAlongSpecifier(direction=d, offset=o, LOCATIONS) }
     | direction=scenic_specifier_position_direction position=expression distance=['by' e=expression { e }] {
         s.DirectionOfSpecifier(direction=direction, position=position, distance=distance, LOCATIONS)
      }
--- a/src/scenic/syntax/scenic.gram
+++ b/src/scenic/syntax/scenic.gram
@@ -468,6 +468,10 @@ class Parser(Parser):
 '''
 
 
+# rule for adding hard keywords
+scenic_hard_keyword: 'PropertyDefault'
+
+
 # STARTING RULES
 # ==============
 
--- a/src/scenic/syntax/veneer.py
+++ b/src/scenic/syntax/veneer.py
@@ -1083,8 +1083,6 @@ def Following(field, dist, fromPt=None):
 	"""
 	if fromPt is None:
 		fromPt = ego()
-	else:
-		dist, fromPt = fromPt, dist
 	if not isinstance(field, VectorField):
 		raise RuntimeParseError('"following F" specifier with F not a vector field')
 	fromPt = toVector(fromPt, '"following F from X for D" with X not a vector')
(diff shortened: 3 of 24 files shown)


src/scenic/syntax/compiler.py
```diff
@@ -218,7 +218,7 @@ def visit_Require(self, node: s.Require):
                     ast.Constant(value=node.lineno),
                     ast.Constant(value=node.name),
                 ],
-                keywords=[ast.keyword(arg="prob", value=node.prob)]
+                keywords=[ast.keyword(arg="prob", value=ast.Constant(node.prob))]
```
shumbo, line 221: Somehow forgot to swap the probability

src/scenic/syntax/scenic.gram
```diff
@@ -468,6 +468,10 @@ class Parser(Parser):
 '''
 
 
+# rule for adding hard keywords
+scenic_hard_keyword: 'PropertyDefault'
```
shumbo, line 472: We want to disallow users from using `PropertyDefault`. `scenic_hard_keyword` is a dangling rule just for defining hard keywords.

src/scenic/syntax/veneer.py
```diff
@@ -1079,8 +1079,6 @@ def Following(field, dist, fromPt=None):
 	"""
 	if fromPt is None:
 		fromPt = ego()
-	else:
```
shumbo, line 1082: This swap is no longer necessary as the new compiler changes the order

tests/syntax/test_compiler.py
```diff
@@ -317,7 +317,7 @@ def test_require(self, options, expected_name, expected_prob):
             ):
                 assert name is None if expected_name is None else name == expected_name
                 compiled_prob = (
-                    kwargs[0].value if kwargs else 1.0
+                    kwargs[0].value.value if kwargs else 1.0
```
shumbo, line 320: `kwargs[0]` is `ast.keyword` and `kwargs[0].value` is `ast.Constant`

tests/syntax/test_errors.py
```diff
@@ -24,49 +25,25 @@ def test_bad_extension(tmpdir):
 
 ### Parse errors during token translation
 
-## Operators used in Python but not Scenic
```
shumbo, line 27: those tests were for the token translator and not relevant anymore

src/scenic/syntax/scenic.gram
```diff
@@ -1252,8 +1256,8 @@ scenic_specifiers: ss=','.scenic_specifier+ { ss }
 scenic_specifier:
     | 'with' p=NAME v=expression { s.WithSpecifier(prop=p.string, value=v, LOCATIONS) }
     | 'at' position=expression { s.AtSpecifier(position=position, LOCATIONS) }
-    | 'offset' 'by' o=expression { s.OffsetBySpecifier(offset=o, LOCATIONS) }
-    | 'offset' 'along' d=expression 'by' o=expression { s.OffsetAlongSpecifier(direction=d, offset=o, LOCATIONS) }
+    | "offset" 'by' o=expression { s.OffsetBySpecifier(offset=o, LOCATIONS) }
```
shumbo, line 1259: `offset` should be a soft keyword

tests/syntax/test_errors.py
```diff
@@ -79,52 +56,52 @@ def test_new_python_class():
...
 def test_reserved_functions():
-    with pytest.raises(TokenParseError):
+    with pytest.raises(SyntaxError):
         compileScenic('PropertyDefault()')
 
 ## Unmatched parentheses and multiline strings
 
 def test_unmatched_parentheses():
-    with pytest.raises(TokenParseError):
+    with pytest.raises(TokenError):
```
shumbo, line 83: The tokenizer raises `TokenError` for errors like unmatched parenthesis. CPython uses the internal state of the tokenizer to produce more readable error messages, but it seems that the state is not accessible from Python. Need more investigation on this matter.

tests/syntax/test_errors.py
dfremont: Agreed, but we should also change the preceding comment (on line 25) since there is no token translation phase any more.

tests/syntax/test_errors.py
dfremont: Interesting. Given that we didn't provide more helpful errors with the old parser I think we can leave this for later (I see you made a TODO item).

tests/syntax/test_errors.py
```diff
@@ -133,14 +110,14 @@ def test_incomplete_infix_operator():
 
 def test_incomplete_infix_package():
     """Packaged (3+-ary) infix operator with too few arguments."""
```
dfremont, line 112: We probably should rename these tests & change the docstrings, since the parser no longer has a concept of argument packages. For example this one could just be `test_incomplete_ternary_operator`.

src/scenic/syntax/scenic.gram
dfremont: OK. We can revisit this once we hide the imports from `veneer` (either putting some prefix in front of them or making them attributes of a single import with an unlikely name).

dfremont (changes requested):
Looks good -- just a couple minor comments.

tests/syntax/test_errors.py
shumbo: Fixed in c2f474ef3c37f0771756c051029443faec6bc5bb

dfremont (approved):

PR #67, opened 2022-07-23
[Parser] port `not visible` specifier to new parser

shumbo:
This PR adds `not visible` specifier to the new parser. The specifier was found to be missing when implementing the new parser, and added to the old parser at dfc58def1f23042f1a8194be0dc513168a9bd5b8.


dfremont (changes requested):
Parser tests are missing but otherwise looks good.

dfremont (approved):

PR #68, opened 2022-07-23
[Parser] support `by` keyword for `mutate` statements

shumbo:
This PR adds support for the `by` keyword for `mutate` statements that was implemented in 9a5a88d7dad90a4f2e1efe5f4fec2793af88613b .

- The `mutate` AST node now has `scale` to store value given with `by`
- `mutate` veneer function now takes `scale` as a keyword argument
  - with the new parser, we can distinguish values passed before and after the `by`
- if `scale` is passed but is not a number, raise an error

One difference from the implementation on `main` is that this grammar supports the statements with `by` without objects (i.e. `mutate by <number>`). In the old parser, the token translator would produce an illegal code `mutate(<< <number>)` and fails to parse, but I believe this should "mutate every object by the given scale factor".

### DIFF
--- a/src/scenic/syntax/ast.py
+++ b/src/scenic/syntax/ast.py
@@ -126,12 +126,19 @@ class Require(AST):
 
 
 class Mutate(AST):
-    __match_args__ = ("elts",)
+    __match_args__ = ("elts", "scale")
 
-    def __init__(self, elts: list[ast.Name], *args: any, **kwargs: any) -> None:
+    def __init__(
+        self,
+        elts: list[ast.Name],
+        scale: Optional[ast.AST] = None,
+        *args: any,
+        **kwargs: any
+    ) -> None:
         super().__init__(*args, **kwargs)
         self.elts = elts
-        self._fields = ["elts"]
+        self.scale = scale
+        self._fields = ["elts", "scale"]
 
 
 # Instance Creation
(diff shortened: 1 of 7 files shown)


dfremont (changes requested):
Looks good (and I agree `mutate by X` should be legal), although we should add a test for `mutate X, Y by Z`.

shumbo:
@dfremont Added two commits to address your comment.

1f3a1b5193c289af7fc71880ac419a50a67070fa adds parser tests for `mutate X, Y by Z` syntax
5e26dce040682a439f58c4f8b749938b2ea09296 adds integration tests for missing cases with `mutate` statements

The second commit uses the `new` keywords while the syntax of other cases on the same file is changed in #66, which is not merged yet. When both PRs are merged, all test cases will be in the new syntax.

Since it is the first time for me to add new cases to `test_basic.py`, if there is a better way of writing tests that asserts values in the Scenic, please let me know.

dfremont (changes requested):
The `sampleParamP` helper is for tests where we just need to check a single number or similar value. For these tests it would be clearer to remove the `param p = other` line, use `sampleScene` to generate a single scene, and access the objects via `scene.objects` (e.g. `ego, other = scene.objects`).

shumbo:
> The `sampleParamP` helper is for tests where we just need to check a single number or similar value. For these tests it would be clearer to remove the `param p = other` line, use `sampleScene` to generate a single scene, and access the objects via `scene.objects` (e.g. `ego, other = scene.objects`).

@dfremont Thanks for pointing this out. Fixed in 00b9fa32fda784667d3dfecba4eedb042ac9e23b

dfremont (approved):

PR #69, opened 2022-07-25
[Parser] Unpack distribution

shumbo:
This PR adds support for unpacking operator (`*`) on distributions. This procedure was done during the AST surgery phase and this PR implements the same to the new compiler.


dfremont (approved):
It's OK if you'd like to leave the inside-a-behavior check out for now, but eventually I'd like to put it back since the superfluous calls to `wrapStarredValue` and `callWithStarArgs` would make debugging behaviors more difficult.

PR #71, opened 2022-08-14
[Parser] Support Multiline Specifiers

shumbo:
This PR adds support for multiline specifiers. In addition, this PR contains migration to the new syntax in several test cases.

## Multiline Specifiers

We concluded that supporting multiline specifiers in general is hard for several reasons. First, this requires substantial changes to the grammar because no Python expression involves indents/dedents. Moreover, it is unclear how multiline specifiers work with other expressions with indents (e.g. lists, tuples). 

Therefore, multiline specifiers implemented by this PR only support three essential cases, which are

- Top-level expressions
  - `new Object...`
- Tracked name assignment
  - `ego = new Object` 
- Assignment
  - `obj = new Object`

This PR adds three compound statements that take a `new` expression with multiline specifiers.

### Trailing Comma

In this PR, a trailing comma is optional at the last line of the specifier suite. For example,

```
new Object with foo 1,
  with bar 2
```

and

```
new Object with foo 1,
  with bar 2,
```

are both valid.

## Fixing tests and overall progress

This PR also adds the `new` keyword to the tests that I forgot to change in #66. As a result, every test under `syntax` except `test_dynamics` and `test_modular` are passing.

Screen Shot 2022-08-14 at 16 31 17

### DIFF
--- a/src/scenic/syntax/scenic.gram
+++ b/src/scenic/syntax/scenic.gram
@@ -465,6 +465,10 @@ class Parser(Parser):
 
         self.raise_syntax_error_known_location(msg, invalid_target)
 
+    # scenic helpers
+    def extend_new_specifiers(self, node: s.New, specifiers: list[ast.AST]) -> s.New:
+        node.specifiers.extend(specifiers)
+        return node
 '''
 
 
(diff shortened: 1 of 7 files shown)


dfremont (approved):
Looks great!

PR #73, opened 2022-08-19
[Parser] implement `require always` and `require eventually`

shumbo:
This PR adds `require always` and `require eventually` statements.

### DIFF
--- a/src/scenic/syntax/compiler.py
+++ b/src/scenic/syntax/compiler.py
@@ -253,6 +253,44 @@ class ScenicToPythonTransformer(ast.NodeTransformer):
             )
         )
 
+    def visit_RequireAlways(self, node: s.RequireAlways):
+        condition = self.visit(node.cond)
+        syntax_id = self._register_requirement_syntax(condition)
+        return ast.Expr(
+            value=ast.Call(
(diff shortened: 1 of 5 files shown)


src/scenic/syntax/compiler.py
```diff
@@ -253,6 +253,44 @@ def visit_Require(self, node: s.Require):
             )
         )
 
+    def visit_RequireAlways(self, node: s.RequireAlways):
+        condition = self.visit(node.cond)
+        syntax_id = self._register_requirement_syntax(condition)
+        return ast.Expr(
```
dfremont, line 259: Since this same code is used for `require`, `require always`, `require eventually`, `record`, `record initial`, etc. with only the veneer function to call being different (and the optional probability in the case of `require`), I'd suggest putting it into a helper function.

dfremont (changes requested):
Looks good -- one small refactoring suggestion above.

shumbo:
Because the helper function suggested here[https://github.com/BerkeleyLearnVerify/Scenic/pull/73#discussion_r950779494] is needed by all requirement-like features (`terminate_when`, `record`, etc.), I'm going to close the related opening PRs and create a single PR.

PR #74, opened 2022-08-19
[Parser] implement `record`, `record initial` and `record final`

shumbo:
This PR implements `record`, `record initial`, and `record final` statements.

Although they are similar, these three statements are implemented in three different functions in `veneer.py` so I chose to implement them as three separate statements each with an AST node, rather than adding optional `initial` or `final` modifiers.


dfremont:
I'll hold off on reviewing this for now, since if you implement the change I suggested for #73 it will change this PR as well.

shumbo:
Closing this in favor of #80

PR #75, opened 2022-08-19
[Parser] Implement `abort` statements

shumbo:
This PR adds `abort` statements which exit functions by returning a special constant value.

### DIFF
--- a/src/scenic/syntax/compiler.py
+++ b/src/scenic/syntax/compiler.py
@@ -517,11 +517,7 @@ class ScenicToPythonTransformer(ast.NodeTransformer):
 
     def visit_Abort(self, node: s.Abort):
         return ast.copy_location(
-            ast.Return(
-                ast.Attribute(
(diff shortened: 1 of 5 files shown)


src/scenic/syntax/compiler.py
```diff
@@ -253,6 +256,16 @@ def visit_Require(self, node: s.Require):
             )
         )
 
+    def visit_Abort(self, node: s.Abort):
+        return ast.copy_location(
+            ast.Return(
+                ast.Attribute(
```
dfremont, line 262: Did you mean to use `abortFlag` here?

dfremont (changes requested):
Looks good -- just one small comment.

src/scenic/syntax/compiler.py
shumbo: You're right. Thanks for the catch. Fixed in d47c02c98b8d1f298906154cf7640089385d9164

dfremont (approved):

PR #77, opened 2022-08-19
[Parser] implement `behavior` and `monitor` definitions

shumbo:
This PR adds `behavior` and `monitor` compound statements for defining behaviors and monitors.

## Changes

While our primary goal is to preserve compatibility, I've made a couple of changes from the current compiler. 

### Position of preconditions/invariants declaration

Scenic 2.0 allows arbitrary ordering of preconditions and invariant declarations. For example,

```python
behavior B:
    precondition: C1
    hello()
    invariant: C2
```

This is valid in the current version of the compiler. In the new parser, preconditions and invariants **must** be placed before any other statements. The example above is invalid because the invariant declaration comes after a normal function call `hello()`.

**TODO:** If this gets merged, add a TODO item to document ordering of `precondition` and `invariant` declarations.

### Invariants and preconditions on `monitor`

The current compiler allows the use of `precondition: ...` and `invariant: ...` inside monitors as well. However, the documentation does not mention the use of these declarations for monitors. In the new parser, preconditions and invariants can only be defined inside behaviors. The use of those syntax inside monitors will be a syntax error.

### DIFF
--- a/src/scenic/syntax/compiler.py
+++ b/src/scenic/syntax/compiler.py
@@ -437,142 +437,26 @@ class ScenicToPythonTransformer(ast.NodeTransformer):
     def visit_BehaviorDef(self, node: s.BehaviorDef):
         # TODO(shun): assert not in behavior or compose
 
-        # --- Extract preconditions and invariants ---
-        preconditions: list[s.Precondition] = []
-        invariants: list[s.Invariant] = []
-        for n in node.header:
-            if isinstance(n, s.Precondition):
-                preconditions.append(n)
-            elif isinstance(n, s.Invariant):
-                invariants.append(n)
-            else:
-                assert (
-                    False
-                ), f"Unexpected node type {n.__class__.__name__} in BehaviorDef header"
-
-        # --- Copy arguments to the behavior object's namespace ---
-
-        # list of all arguments
-        allArgs = itertools.chain(
-            node.args.posonlyargs, node.args.args, node.args.kwonlyargs
+        return self.makeBehaviorLikeDef(
+            baseClassName="Behavior",
+            name=node.name,
+            args=node.args,
+            docstring=node.docstring,
+            header=node.header,
+            body=node.body,
         )
-        # statements that create argument variables
-        copyArgs: list[ast.AST] = []
-        for arg in allArgs:
-            dest = ast.Attribute(
-                ast.Name(behaviorArgName, loadCtx), arg.arg, ast.Store()
-            )
-            copyArgs.append(
-                ast.copy_location(
-                    ast.Assign([dest], ast.Name(arg.arg, loadCtx)), arg
-                )
-            )
-
-        # --- Create a new `arguments` ---
-        newArgs: ast.arguments = self.visit(node.args)
-        # add private current behavior argument and implicit `self` argument
-        newArgs.posonlyargs = initialBehaviorArgs + newArgs.posonlyargs
-
-        # --- Process body ---
-        self.inBehavior = True
-        oldBehaviorLocals = self.behaviorLocals
-
-        self.behaviorLocals = allLocals = LocalFinder.findIn(node.body)
-
-        # handle docstring
-        body = node.body
-        docstring = []
-        if node.docstring is not None:
-            docstring = [ast.Expr(ast.Constant(node.docstring))]
-
-        # process body
-        statements = self.visit(body)
-
-        generatorBody = copyArgs + statements
-        generatorDefinition = ast.FunctionDef(
-            name="makeGenerator",
-            args=newArgs,
-            body=generatorBody,
-            decorator_list=[],
-            returns=None,
-        )
-
-        # --- Save local variables ---
-        saveLocals = ast.Assign(
-            [ast.Name("_locals", ast.Store())], ast.Constant(frozenset(allLocals))
-        )
-
-        # --- Guards ---
-        guardCheckers = self.makeGuardCheckers(newArgs, preconditions, invariants)
-
-        # --- Create class definition ---
-        classBody = docstring + guardCheckers + [saveLocals, generatorDefinition]
-        classDefinition = ast.ClassDef(
-            node.name, [ast.Name("Behavior", loadCtx)], [], classBody, []
-        )
-
-        self.inBehavior = False
-        self.behaviorLocals = oldBehaviorLocals
-
-        return ast.copy_location(classDefinition, node)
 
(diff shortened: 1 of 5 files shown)


src/scenic/syntax/scenic.gram
```diff
@@ -737,6 +741,71 @@ scenic_class_property_attribute: &&(
...
+    # if only preconditions and/or invariants are specified
+    | NEWLINE INDENT a=[x=STRING NEWLINE { x.string }] b=scenic_behavior_header DEDENT { (a, b, [ast.Pass()]) }
+    | NEWLINE INDENT a=[x=STRING NEWLINE { x.string }] b=[scenic_behavior_header] c=scenic_behavior_statements DEDENT { (a, b or [], c) }
+    | invalid_block
+
+scenic_behavior_statements[list]: a=scenic_behavior_statement+ { list(itertools.chain.from_iterable(a)) }
+
+# statements available inside behavior (normal statements + dynamic statements - precondition/invariant)
+scenic_behavior_statement[list]:
+    | scenic_invalid_behavior_statement
```
shumbo, line 770: Because `precondition: ...` and `invariant: ...` can be parsed as typed variable initialization, use this rule to catch any precondition and invariant definition after any other statements and raise syntax error.

src/scenic/syntax/compiler.py
```diff
@@ -162,6 +259,202 @@ def transformPropertyDef(self, node: s.PropertyDef):
...
+            copyArgs.append(
+                ast.copy_location(
+                    ast.Assign([dest], ast.Name(arg.arg, loadCtx)), arg
+                )
+            )
+
+        # --- Create a new `arguments` ---
+        newArgs: ast.arguments = self.visit(node.args)
+        # add private current behavior argument and implicit `self` argument
+        newArgs.posonlyargs = initialBehaviorArgs + newArgs.posonlyargs
```
shumbo, line 299: https://github.com/BerkeleyLearnVerify/Scenic/blob/eb7e7f7ad1879c70f0f6d9f5cff7b83afd23ef7a/src/scenic/syntax/translator.py#L1887

The current compiler adds `initialBehaviorArgs` to the beginning of `args`. Because `posonlyargs` precedes `args`, if you write `behavior(posonly, /, arg, *, kwdonly)`, `initialBehaviorArgs` will be added in between `posonly` and `kwdonly`. When this function is called, the contents for `initialBehaviorArgs` are passed as positioned arguments, so this will not work.

Although I'm not sure if someone would want to use position only arguments for behaviors, the easy fix is to add `initialBehaviorArgs` to `posonlyargs` of the new argument node.

src/scenic/syntax/scenic.gram
```diff
@@ -737,6 +741,71 @@ scenic_class_property_attribute: &&(
...
+            docstring=c[0],
+            header=c[1],
+            body=c[2],
+            LOCATIONS,
+        )
+     }
+
+scenic_behavior_def_block:
+    # if only preconditions and/or invariants are specified
+    | NEWLINE INDENT a=[x=STRING NEWLINE { x.string }] b=scenic_behavior_header DEDENT { (a, b, [ast.Pass()]) }
```
shumbo, line 762: This rule is needed to parse behaviors only with preconditions or invariants and without any plain statements.

e.g.

```
behavior B:
    precondition: condition
```

dfremont: Let's simplify the grammar by disallowing these: such behaviors already trigger an error at simulation time since they take no actions.

shumbo: Fixed in 9e4d682fdffffdc802a9aa7cc3d981d17a23ae6d

src/scenic/syntax/compiler.py
```diff
@@ -162,6 +259,202 @@ def transformPropertyDef(self, node: s.PropertyDef):
...
+            node.name, [ast.Name("Behavior", loadCtx)], [], classBody, []
+        )
+
+        self.inBehavior = False
+        self.behaviorLocals = oldBehaviorLocals
+
+        return ast.copy_location(classDefinition, node)
+
+    def visit_MonitorDef(self, node: s.MonitorDef):
+        # TODO(shun): assert not in behavior or compose
```
dfremont, line 345: I think it would be clearer to unify this with `visit_BehaviorDef`, i.e. break out the code into a common helper function. Once we allow monitors to take arguments, the only difference will be in the superclass used (`Behavior` vs. `Monitor`).

tests/syntax/test_parser.py
```diff
@@ -177,6 +177,222 @@ class C:
...
+                """
+            )
+
+    def test_invalid_invariant(self):
+        with pytest.raises(SyntaxError) as e:
+            parse_string_helper(
+                """
+                behavior test:
+                    hello()
+                    precondition: True
```
dfremont, line 326: Should be `invariant`

dfremont (changes requested):
Looks good overall -- just a few minor comments.

tests/syntax/test_parser.py
shumbo: Fixed in 3c42270a0210e559af93795194268623f77c8b48

src/scenic/syntax/compiler.py
shumbo: Fixed this in b86446e629391f0a45b1c026d4dbc17018825ff9. Created the `makeBehaviorLikeDef` helper function and moved the shared code there.

At this moment, `Behavior` takes two additional arguments: `args` and `header` (preconditions&invariants). When calling the helper from `visit_Monitor`, it uses the empty arguments for `args` and the empty list for the header.

src/scenic/syntax/compiler.py
```diff
@@ -61,6 +71,75 @@ def visit_Attribute(self, node):
...
+        if isinstance(node.ctx, (ast.Store, ast.Del)):
+            self.names.add(node.id)
+
+    def visit_ExceptHandler(self, node):
+        if node.name is not None:
+            self.names.add(node.name)
+        self.generic_visit(node)
+
+
+def unquote(s: str) -> str:
```
shumbo, line 135: The parser uses `token.string` to create the `docstring` field of `BehaviorDef` and `MonitorDef`. This includes quotation marks and we need to get rid of them at the compiler level (this can probably be done at the parser level, but I don't want to make the grammar too complex by writing a Python helper function).

This `unquote` function looks for four possible patterns for defining strings (`"` `'` `'''` and `"""`) and extract the body.

dfremont (approved):

PR #78, opened 2022-08-19
[Parser] invoke dynamic actions

shumbo:
This PR implements six statements that invoke dynamic actions, namely

- take
- wait
- terminate
- do for
- do until
- do

### How many operands do `do` statements take?

~The documentation says the `do` statement can take more than one behaviors/actions. However, the implementation only allows one behavior/action at a time. There also is a test case that asserts an error is raised when more than one operands are specified.~

https://github.com/BerkeleyLearnVerify/Scenic/blob/f1c42f1559e555ec50073e30ee4a1f4a50fdadf2/tests/syntax/test_dynamics.py#L424

~In this PR, I chose to keep the current behavior and the statement can only take one operand. However, I believe it makes sense to add support for invoking more than one behaviors/scenarios.~

`do` can take multiple actions inside modular scenarios but not in behaviors and monitors. At the parser level, we allow multiple operands to be passed to do statements. At the compiler level, we throw an error if more than one operands are given inside behaviors/monitors.

### DIFF
--- a/src/scenic/syntax/compiler.py
+++ b/src/scenic/syntax/compiler.py
@@ -702,6 +702,91 @@ class ScenicToPythonTransformer(ast.NodeTransformer):
             node,
         )
 
+    def visit_Take(self, node: s.Take):
+        action = ast.Tuple(self.visit(node.elts), loadCtx)
+        return self.generateInvocation(node, ast.copy_location(action, node))
+
+    def visit_Wait(self, node: s.Wait):
+        return self.generateInvocation(node, ast.Constant(()))
+
+    def visit_Terminate(self, node: s.Terminate):
+        termination = ast.Call(
+            ast.Name("makeTerminationAction", loadCtx), [ast.Constant(node.lineno)], []
+        )
+        return self.generateInvocation(node, termination)
+
+    def visit_Do(self, node: s.Do):
+        subHandler = ast.Attribute(
+            ast.Name(behaviorArgName, loadCtx), "_invokeSubBehavior", loadCtx
+        )
+        subArgs = [
+            ast.Name("self", loadCtx),
+            ast.Tuple([self.visit(node.value)], loadCtx),
+        ]
+        subRunner = ast.Call(subHandler, subArgs, [])
+        return self.generateInvocation(node, subRunner, ast.YieldFrom)
+
+    def visit_DoFor(self, node: s.DoFor):
+        subHandler = ast.Attribute(
+            ast.Name(behaviorArgName, loadCtx), "_invokeSubBehavior", loadCtx
+        )
+        subArgs = [
+            ast.Name("self", loadCtx),
+            ast.Tuple([self.visit(node.value)], loadCtx),
+            ast.Call(
+                func=ast.Name("Modifier", loadCtx),
+                args=[
+                    ast.Constant("for"),
+                    self.visit(node.duration.value),
+                    ast.Constant(node.duration.unitStr),
+                ],
+                keywords=[],
+            ),
+        ]
+        subRunner = ast.Call(subHandler, subArgs, [])
+        return self.generateInvocation(node, subRunner, ast.YieldFrom)
+
+    def visit_DoUntil(self, node: s.DoUntil):
+        subHandler = ast.Attribute(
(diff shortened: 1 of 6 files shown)


src/scenic/syntax/compiler.py
```diff
@@ -253,6 +274,91 @@ def visit_Require(self, node: s.Require):
...
+                    ast.Constant("until"),
+                    ast.Lambda(noArgs, self.visit(node.cond)),
+                ],
+                keywords=[],
+            ),
+        ]
+        subRunner = ast.Call(subHandler, subArgs, [])
+        return self.generateInvocation(node, subRunner, ast.YieldFrom)
+
+    def generateInvocation(self, node: ast.AST, actionlike, invoker=ast.Yield):
```
shumbo, line 340: This is the method for generating code that invokes behaviors and used by all other visitors.

tests/syntax/test_compiler.py
```diff
@@ -363,6 +363,233 @@ def test_require(self, options, expected_name, expected_prob):
...
+                        )
+                    )
+                ),
+                checkInvariants,
+            ]:
+                self.assert_invocation_check_invariants(checkInvariants)
+            case _:
+                assert False
+
+    def assert_invocation_check_invariants(self, node):
```
shumbo, line 549: This method asserts that `node` is the code for running invariant checks. Used by all other cases added in this PR.

src/scenic/syntax/compiler.py
```diff
@@ -76,6 +82,21 @@ def generic_visit(self, node):
             )
         return super().generic_visit(node)
 
+    def visit(self, node: ast.AST | list[ast.AST]):
```
shumbo, line 85: Because `generateInvocation` returns a list of AST nodes, we need to change the `visit` to support list

src/scenic/syntax/compiler.py
```diff
@@ -9,10 +9,16 @@
 def compileScenicAST(scenicAST: ast.AST) -> Tuple[ast.AST, List[ast.AST]]:
     """Compiles Scenic AST to Python AST"""
     compiler = ScenicToPythonTransformer()
-    node = ast.fix_missing_locations(compiler.visit(scenicAST))
+    node = compiler.visit(scenicAST)
+    # call `fix_missing_locations` if node is an AST node
+    if isinstance(node, ast.AST):
```
shumbo, line 14: `node` can be a list in tests and `fix_missing_locations` cannot be called. We skip this if `node` is not an AST node.

src/scenic/syntax/compiler.py
```diff
@@ -253,6 +274,94 @@ def visit_Require(self, node: s.Require):
...
+        return self.generateInvocation(node, ast.Constant(()))
+
+    def visit_Terminate(self, node: s.Terminate):
+        termination = ast.Call(
+            ast.Name("makeTerminationAction", loadCtx), [ast.Constant(node.lineno)], []
+        )
+        return self.generateInvocation(node, termination)
+
+    def visit_Do(self, node: s.Do):
+        subHandler = ast.Attribute(
```
shumbo, line 291: `visit_Do` `visit_DoFor` and `visit_DoUntil` are similar and might be able to extract the common code to a helper function. I want to wait until I implement `do choose` and `do shuffle` so I can choose the right level of abstraction for that function.

src/scenic/syntax/compiler.py
```diff
@@ -253,6 +274,94 @@ def visit_Require(self, node: s.Require):
...
+            ast.Name(behaviorArgName, loadCtx), "_invokeSubBehavior", loadCtx
+        )
+        # TODO(shun): Check node has no more than one element inside behavior/monitors
```
shumbo, line 294: `do` can only take one argument inside behaviors and monitors. Need to check that `len(node.elts) == 1` if `self.inCompose` being added in #77 is `True`.

dfremont (approved):
Looks good. Let's make a note to consider cleaning up the uses of `Modifier`, which were an artifact of the old `TokenTranslator` and can probably be clarified now. (Let's hold off for the moment since it will require changes to the veneer.)

PR #79, opened 2022-08-19
[Parser] implement `try-interrupt` statements

shumbo:
This PR implements `try-interrupt` statements.

### DIFF
--- a/src/scenic/syntax/ast.py
+++ b/src/scenic/syntax/ast.py
@@ -13,7 +13,7 @@ class AST(ast.AST):
 
 
 class TryInterrupt(AST):
-    """Scenic AST node that represents try-catch statements with `interrupt when` clauses"""
+    """Scenic AST node that represents try-interrupt statements"""
 
     __match_args__ = (
         "body",
--- a/src/scenic/syntax/ast.py
+++ b/src/scenic/syntax/ast.py
@@ -10,6 +10,57 @@ class AST(ast.AST):
 
 
 # special statements
+
+
+class ScenicTry(AST):
+    """Scenic AST node that represents try-catch statements with `interrupt when` clauses"""
+
+    __match_args__ = (
+        "body",
+        "interrupt_when_handlers",
+        "except_handlers",
+        "orelse",
+        "finalbody",
+    )
+
+    def __init__(
+        self,
+        body: list[ast.stmt],
+        interrupt_when_handlers: list["InterruptWhenHandler"],
+        except_handlers: list[ast.ExceptHandler],
+        orelse: list[ast.stmt],
+        finalbody: list[ast.AST],
+        *args: any,
+        **kwargs: any
+    ) -> None:
+        super().__init__(*args, **kwargs)
+        self.body = body
+        self.interrupt_when_handlers = interrupt_when_handlers
+        self.except_handlers = except_handlers
+        self.orelse = orelse
+        self.finalbody = finalbody
+        self._fields = [
+            "body",
+            "interrupt_when_handlers",
+            "except_handlers",
+            "orelse",
+            "finalbody",
+        ]
+        self._attributes = []
+
+
+class InterruptWhenHandler(AST):
+    __match_args__ = ("cond", "body")
+
+    def __init__(
+        self, cond: ast.AST, body: list[ast.AST], *args: any, **kwargs: any
+    ) -> None:
+        super().__init__(*args, **kwargs)
+        self.cond = cond
+        self.body = body
+        self._fields = ["cond", "body"]
+
+
 class TrackedAssign(AST):
     __match_args__ = (
         "target",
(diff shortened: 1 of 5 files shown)


src/scenic/syntax/ast.py
```diff
@@ -10,6 +10,57 @@ def __init__(self, *args: any, **kwargs: any) -> None:
 
 
 # special statements
+
+
+class ScenicTry(AST):
+    """Scenic AST node that represents try-catch statements with `interrupt when` clauses"""
```
dfremont, line 16: I'd call the class `TryInterrupt` and change the docstring accordingly, since these are called "try-interrupt" statements in the paper and documentation. It's true their syntax generalizes the Python `try` statement, but they're only valid in limited contexts (inside behaviors and `compose` blocks of modular scenarios), so I wouldn't consider them a form of the ordinary `try`.

src/scenic/syntax/compiler.py
```diff
@@ -9,10 +10,34 @@
 def compileScenicAST(scenicAST: ast.AST) -> Tuple[ast.AST, List[ast.AST]]:
     """Compiles Scenic AST to Python AST"""
     compiler = ScenicToPythonTransformer()
-    node = ast.fix_missing_locations(compiler.visit(scenicAST))
+    tree = compiler.visit(scenicAST)
+    if isinstance(tree, list):
+        node = [ast.fix_missing_locations(n) for n in tree]
```
dfremont, line 15: If this can return a list, the type annotation should change.

src/scenic/syntax/compiler.py
```diff
@@ -94,8 +215,150 @@ def visit_Name(self, node: ast.Name) -> any:
...
+        else:
+            return self.generic_visit(node)
+
     # Special Case
 
+    def visit_ScenicTry(self, node: s.ScenicTry):
+        statements = []
+        oldInTryInterrupt = self.inTryInterrupt
+
+        # TODO(shun): Do we need to create a dead copy? https://github.com/BerkeleyLearnVerify/Scenic/blob/eb7e7f7ad1879c70f0f6d9f5cff7b83afd23ef7a/src/scenic/syntax/translator.py#L1707-L1709
```
dfremont, line 273: You're right, this is no longer required: I'd forgotten that we now implement local variables of behaviors and modular scenarios as attributes (so that one can write `myScenario.someObject` for example; as a side effect this syntax also works for behaviors). I've removed the dead copy from the `main` branch and added tests for the cases that it used to be necessary for, just in case (in commit e0f3f0c0).

src/scenic/syntax/ast.py
```diff
@@ -10,6 +10,57 @@ def __init__(self, *args: any, **kwargs: any) -> None:
 
 
 # special statements
+
+
+class ScenicTry(AST):
```
dfremont, line 15: Let's call this `ScenicTryInterrupt` since in your implementation it isn't in fact used for ordinary `try-except` statements.

tests/syntax/test_compiler.py
```diff
@@ -7,6 +7,125 @@
...
 
 class TestCompiler:
+    def test_scenic_try(self):
+        node, _ = compileScenicAST(
+            ScenicTry(
+                body=[Expr(Call(Name("foo", Load()), args=[], keywords=[]))],
+                interrupt_when_handlers=[
+                    InterruptWhenHandler(
+                        Constant(True),
+                        [Expr(Call(Name("foo", Load()), args=[], keywords=[]))],
```
dfremont, line 17: Let's have the function be "bar" so that we can catch if the body and/or interrupt handlers are put in the wrong place.

dfremont (changes requested):
Looks good overall -- various minor comments above.

src/scenic/syntax/ast.py
shumbo: Fixed in e843c6c4af5948f148efb71414a917ad82b71f3c. I changed the name of the class to `TryInterrupt`.

src/scenic/syntax/compiler.py
shumbo: Fixed in 38b02303be4841a4ee19f9e54de5714d63c8ac9c

src/scenic/syntax/compiler.py
shumbo: Thanks for your investigation. Removed the comment in b65ffdb19489db98da9b53d756d0531c913204cf

src/scenic/syntax/ast.py
shumbo: I changed it to `TryInterrupt` (see https://github.com/BerkeleyLearnVerify/Scenic/pull/79#discussion_r950783059)

The reason I called it `ScenicTry` originally was to disambiguate it from Python's `ast.Try` but since we're now calling it `TryInterrupt` I don't think we need to add `Scenic` at the beginning. Let me know if you think we still should have the prefix.

tests/syntax/test_compiler.py
shumbo: Thanks for the catch. Fixed in 93a961e4a33b157247cf8364dac7f96d44520097

dfremont (approved):

PR #80, opened 2022-08-23
[Parser] implement requirement-like syntax

shumbo:
This PR implements Scenic constructs with the requirement-like syntax:

- `require always`
- `require eventually`
- `record`
- `record initial`
- `record final`
- `terminate when`
- `terminate simulation when`

These features need to preserve the syntax of the "body" of the call to be evaluated later, and the corresponding functions in veneer have similar signatures: `(reqId: int, body: () -> bool, lineno: int, name: str | None)`, with optional probability for plain requirements.

This PR adds grammar for the statements listed above to be able to parse them. When compiling these statements, visitors call the `createRequirementLike` helper function. It creates the call to the function while registering the syntax of the value to be evaluated later.

### DIFF
--- a/src/scenic/syntax/compiler.py
+++ b/src/scenic/syntax/compiler.py
@@ -820,7 +820,7 @@ class ScenicToPythonTransformer(ast.NodeTransformer):
         name: Optional[str] = None,
         prob: Optional[float] = None,
     ):
-        """Create a call to a function that implements requirement-like features, such as `record` and `terminate`.
+        """Create a call to a function that implements requirement-like features, such as `record` and `terminate when`.
 
         Args:
             functionName (str): Name of the requirement-like function to call. Its signature must be `(reqId: int, body: () -> bool, lineno: int, name: str | None)`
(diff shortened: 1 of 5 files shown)


src/scenic/syntax/compiler.py
```diff
@@ -233,22 +233,77 @@ def visit_Param(self, node: s.Param):
...
+
+    def createRequirementLike(
+        self,
+        functionName: str,
+        syntax: ast.AST,
+        lineno: int,
+        name: Optional[str] = None,
+        prob: Optional[float] = None,
+    ):
+        """Create a call to a function that implements requirement-like features, such as `record` and `terminate`.
```
dfremont, line 286: To avoid confusion, let's put "`terminate when`".

dfremont (approved):
Looks great -- thanks for refactoring this!

src/scenic/syntax/compiler.py
shumbo: Fixed this in 18a3e7202167a311d5fc80829585c82a98e24f27

PR #84, opened 2022-08-26
[Parser] implement `scenario` definitions

shumbo:
This PR adds `scenario` definitions.

### DIFF
--- a/src/scenic/syntax/compiler.py
+++ b/src/scenic/syntax/compiler.py
@@ -184,6 +184,9 @@ class ScenicToPythonTransformer(ast.NodeTransformer):
         self.behaviorLocals: set = set()
         "Set of variable names on the local scope of the behavior"
 
+        self.inCompose: bool = False
+        "True if the transformer is processing scenario"
+
         self.inTryInterrupt = False
         self.inInterruptBlock = False
         self.inLoop = False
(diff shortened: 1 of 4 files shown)


src/scenic/syntax/compiler.py
```diff
@@ -184,6 +184,9 @@ def __init__(self) -> None:
         self.behaviorLocals: set = set()
         "Set of variable names on the local scope of the behavior"
 
+        self.inCompose: bool = False
+        "True if the transformer is processing scenario"
```
dfremont, line 188: Let's be more precise in this comment, since the flag isn't true inside the `setup` block, only the `compose` block.

src/scenic/syntax/compiler.py
```diff
@@ -476,6 +479,63 @@ def visit_MonitorDef(self, node: s.MonitorDef):
...
+        # Find all locals of the scenario, which will be shared amongst the various blocks
+        allLocals = set()
+        if node.compose:
+            allLocals.update(LocalFinder.findIn(node.compose))
+        if node.setup:
+            allLocals.update(LocalFinder.findIn(node.setup))
+        oldBL = self.behaviorLocals
+        self.behaviorLocals = allLocals
+
+        guardCheckers = self.makeGuardCheckers(args, preconditions, invariants)
```
dfremont, line 504: Looks like this line got duplicated; only the copy on line 508 should exist.

src/scenic/syntax/compiler.py
```diff
@@ -476,6 +479,63 @@ def visit_MonitorDef(self, node: s.MonitorDef):
...
+
+        # Get preconditions and invariants
+        preconditions, invariants = self.separatePreconditionsAndInvariants(node.header)
+
+        setup = node.setup
+        compose = node.compose
+
+        # Find all locals of the scenario, which will be shared amongst the various blocks
+        allLocals = set()
+        if node.compose:
```
dfremont, line 497: Any reason not to just use `compose` and `setup` respectively for these kinds of conditions? This makes it look like we're saving the old value of `node.compose` in case it changes.

dfremont (changes requested):
Looks good -- just a few minor comments above.

src/scenic/syntax/compiler.py
shumbo: Fixed in 2c73958eb72971dcb12d86dd5329a5e1b35b039a

src/scenic/syntax/compiler.py
shumbo: Thanks for catching this. Fixed in 6f97e73e275f706e0910413aee4430e9a1c20748

src/scenic/syntax/compiler.py
shumbo: My intention was to use `compose` and `setup` for compiled AST nodes and use `node.compose` and `node.setup` to access the original nodes, but I must have mixed them up.

f7e299bc671ddec259cac0e8de2e659732226085

This should fix the problems. Now `node.compose` and `node.setup` are compiled to `compose` and `setup`, respectively, and then will be used to create a class definition.

dfremont (approved):

PR #85, opened 2022-08-27
[Parser] implement `override` statements

shumbo:
This PR implements `override` statements.

### DIFF
--- a/src/scenic/syntax/scenic.gram
+++ b/src/scenic/syntax/scenic.gram
@@ -567,7 +567,7 @@ scenic_compound_stmt:
     | scenic_monitor_def
     | scenic_scenario_def
     | scenic_try_interrupt_stmt
-
+    | scenic_override_stmt
 
 # SIMPLE STATEMENTS
 # =================
@@ -884,6 +884,14 @@ scenic_scenario_setup_block:
 
 scenic_scenario_compose_block:
     | "compose" &&':' b=block { b }
+
+# Override
+# --------
+
+scenic_override_stmt:
+    | 'override' e=primary ss=scenic_specifiers t=(NEWLINE { [] } | ',' NEWLINE INDENT x=scenic_new_block_body DEDENT { x }) {
+        s.Override(target=e, specifiers=ss + t)
+     }
 
 # Function definitions
 # --------------------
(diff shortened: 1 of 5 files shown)


src/scenic/syntax/scenic.gram
```diff
@@ -854,6 +854,14 @@ scenic_monitor_def_block:
 
 scenic_monitor_statements[list]: a=statement+ { list(itertools.chain.from_iterable(a)) }
 
+# Override
+# --------
+
+scenic_override_stmt:
+    | 'override' e=primary ss=scenic_specifiers NEWLINE { s.Override(target=e, specifiers=ss) }
```
shumbo, line 861: I tried to write a single rule that matches single-line and multiline specifiers, but I couldn't find a way to do it cleanly due to the implementation of Pegen.

```
'override' e=primary ss=scenic_specifiers t=(NEWLINE { [] } | ',' NEWLINE INDENT x=scenic_new_block_body DEDENT { x })
```

Something like this should work, but since `[]` is a falsy value in Python and Pegen depends on Python's truthy check to determine if it has found a matching string, the first case of `t` is never accepted. I can create a custom value that is truthy and convert that to the empty list inside curly braces; however, I figured writing separate rules is cleaner.

dfremont: Agreed, the grammar is more understandable this way.

src/scenic/syntax/scenic.gram
```diff
@@ -854,6 +854,14 @@ scenic_monitor_def_block:
 
 scenic_monitor_statements[list]: a=statement+ { list(itertools.chain.from_iterable(a)) }
 
+# Override
+# --------
+
+scenic_override_stmt:
```
dfremont, line 860: For the avoidance of future confusion, let's put a comment explaining that we're restricting `e` to a ` primary` rather than a full expression to avoid possible ambiguities.

dfremont (changes requested):
Looks great -- one tiny comment.

src/scenic/syntax/scenic.gram
shumbo: 41f9bd1f0432bacd33fb87521a19300163875fcc

Added a comment. Please let me know if it is not descriptive enough.

dfremont (approved):

PR #88, opened 2022-09-02
[Parser] generate parser before running tests

shumbo:
## Overview

This PR hooks `pytest_configure` to generate the parser before running tests. This ensures that the grammar is valid when running the test suite and test cases are executed with the updated parser.

The parser will be generated by default. Use the newly added option `--skip-pegen` to skip the process.

### Default behavior

In this PR, pytest generates the parser by default and we provide an option to skip it. We can choose to not generate the parser by default and add an option to turn it on.

Since the parser generation is pretty fast (around 500ms on my computer), I believe we can leave this on by default, but one might argue that the default should be off as the grammar might not change so often.

### What to execute

~Since Pegen does not expose APIs, we must use its CLI interface. We can directly call `pegen` (`python -m pegen`) or use `Makefile` in the project root.~

~In this PR, I chose to call `make` so I don't have to copy arguments to `conftest.py`. However, this requires `make` to run the test suite, which might not be available in some environments (especially on Windows).~

~Possible alternatives includes:~

- ~write the same command in both `Makefile` and `conftest.py`~
- ~create a separate Python script for generating the parser (by calling pegen internally) and use that from `conftest.py` (call the exposed function) and `Makefile` (call the script).~

### Update (9/9)

I tried to place a Python script that calls `pegen` in `src/scenic/syntax` but I realized that this approach doesn't work because `scenic.syntax` is invalid without the parser file and importing the build script raises an error.

I also realized that the options are not that long, so I decided to just write the build script one more time in `conftest.py`.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -1,6 +1,9 @@
 
 import os.path
+import subprocess
+from pathlib import Path
 from contextlib import contextmanager
+
 import pytest
 
 ## Fixtures for use in tests
@@ -27,13 +30,31 @@ def runLocally(request):
             os.chdir(oldDirectory)
     return manager
 
-## Command-line option to skip very slow tests
 
 def pytest_addoption(parser):
+    # option to skip very slow tests
     parser.addoption('--fast', action='store_true', help='skip very slow tests')
+    # option to skip parser generation
+    parser.addoption('--skip-pegen', action='store_true', help='generate the parser before running tests')
+
+class ParserGenerationException(BaseException):
+    pass
 
 def pytest_configure(config):
     config.addinivalue_line('markers', 'slow: mark test as very slow')
+
+    if not config.getoption("skip_pegen"):
+        result = subprocess.run(
+            [
+                "make",
+                "--always-make",
+            ],
+            cwd=Path(__file__).parent.parent,
+            capture_output=True,
+            text=True,
+        )
+        if result.returncode != 0:
+            raise ParserGenerationException(f"Failed to generate the parser: {result.stderr}")
 
 def pytest_collection_modifyitems(config, items):
     if config.getoption('--fast'):
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -50,8 +50,6 @@ def pytest_configure(config):
         parser = syntaxDir / "parser.py"
         result = subprocess.run(
             [
-                "poetry",
-                "run",
                 "python",
                 "-m",
                 "pegen",
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -35,7 +35,7 @@ def pytest_addoption(parser):
     # option to skip very slow tests
     parser.addoption('--fast', action='store_true', help='skip very slow tests')
     # option to skip parser generation
-    parser.addoption('--skip-pegen', action='store_true', help='generate the parser before running tests')
+    parser.addoption('--skip-pegen', action='store_true', help='skip generating the parser before running tests')
 
 class ParserGenerationException(BaseException):
     pass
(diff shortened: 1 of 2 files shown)


tests/conftest.py
```diff
@@ -27,14 +30,42 @@ def manager():
             os.chdir(oldDirectory)
     return manager
 
-## Command-line option to skip very slow tests
 
 def pytest_addoption(parser):
+    # option to skip very slow tests
     parser.addoption('--fast', action='store_true', help='skip very slow tests')
+    # option to skip parser generation
+    parser.addoption('--skip-pegen', action='store_true', help='generate the parser before running tests')
```
dfremont, line 38: `help` message seems to be backwards.

tests/conftest.py
```diff
@@ -27,14 +30,42 @@ def manager():
...
     config.addinivalue_line('markers', 'slow: mark test as very slow')
 
+    if not config.getoption("skip_pegen"):
+        projectRootDir = Path(__file__).parent.parent
+        syntaxDir = projectRootDir / "src" / "scenic" / "syntax"
+        grammar = syntaxDir / "scenic.gram"
+        parser = syntaxDir / "parser.py"
+        result = subprocess.run(
+            [
+                "poetry",
```
dfremont, line 53: Is this necessary? If the test suite is being run inside a virtual environment then that version of Python should get used automatically, I think (since the subprocess uses the same `PATH`). This code seems like it could run the subprocess in a different Python than the test suite, if you run the test suite from a virtualenv that isn't managed using Poetry.

tests/conftest.py
```diff
@@ -27,14 +30,42 @@ def manager():
...
+                grammar,
+                "-o",
+                parser,
+            ],
+            cwd=projectRootDir,
+            capture_output=True,
+            text=True,
+        )
+        if result.returncode != 0:
+            raise ParserGenerationException(f"Failed to generate the parser: {result.stderr}")
```
dfremont, line 67: Any reason for a new kind of exception here? We could just use `pytest.exit`.

dfremont (changes requested):
A couple questions but otherwise looks good.

tests/conftest.py
shumbo: Fixed 7becaf085edf0291de565e42301ab0cb88706dc1

tests/conftest.py
shumbo: Fixed faec3f12e3481185a78e9be1eaf7f944ecbbdd31

tests/conftest.py
shumbo: Fixed faec3f12e3481185a78e9be1eaf7f944ecbbdd31

tests/conftest.py
shumbo: Fixed in c204e3c8de2dc15d60bcb517ad6cb4efd4b96e52

dfremont: Let's set the return code too: `pytest.exit` uses 0 by default, whereas we should return 1 to indicate an error or even propagate the value from the subprocess.

dfremont (changes requested):
One last tweak but otherwise looks good.

tests/conftest.py
shumbo: I believe propagating the value from the subprocess makes sense and implemented it in c098877b505a5225ddb70fac990af0c4715730f3 (It seems that pegen always returns `1` for errors, making it technically the same as setting return code 1 for `pytest.exit`)

Screen Shot 2022-09-21 at 9 37 55

I tested my code against invalid grammar and confirmed that the status code is `1` after execution.

dfremont (approved):

PR #96, opened 2022-09-23
[Parser] Fix `test_errors.py`

shumbo:
This PR includes a couple of changes to pass `test_errors.py`.

## `context` decorator

One of the things that are tested in `test_errors.py` is whether statements are allowed in the context.

[Full list of statements with restrictions](https://shun-notes.notion.site/Scenic-statements-with-context-restriction-15db4bd4ed754e58b6e29bfdcf2bcc82)

These are hard to catch in grammar as there's only one `statement` rule that's used across different blocks. 

I propose the `context` decorator to catch those errors at the compiler level. The decorator takes an enum flag `Context` and the decorated visitor raises a syntax error if it is called on an illegal context.

### DIFF
--- a/src/scenic/syntax/ast.py
+++ b/src/scenic/syntax/ast.py
@@ -5,6 +5,8 @@ from typing import Optional, Union
 class AST(ast.AST):
     "Scenic AST base class"
 
+    _attributes = ("lineno", "col_offset", "end_lineno", "end_col_offset")
+
     def __init__(self, *args: any, **kwargs: any) -> None:
         super().__init__(*args, **kwargs)
 
--- a/src/scenic/syntax/compiler.py
+++ b/src/scenic/syntax/compiler.py
@@ -4,6 +4,7 @@ import itertools
 from typing import Callable, Literal, Optional, Tuple, List, Union
 
 import scenic.syntax.ast as s
+from scenic.core.errors import getText
 
 # exposed functions
 
@@ -11,6 +12,7 @@ import scenic.syntax.ast as s
 def compileScenicAST(
     scenicAST: ast.AST,
     *,
+    filename: Optional[str] = None,
     inBehavior: bool = False,
     inMonitor: bool = False,
     inCompose: bool = False,
@@ -19,6 +21,9 @@ def compileScenicAST(
 ) -> Tuple[Union[ast.AST, list[ast.AST]], List[ast.AST]]:
     """Compiles Scenic AST to Python AST"""
     compiler = ScenicToPythonTransformer()
+
+    # set filename
+    compiler.filename = filename
 
     # set optional flags
     compiler.inBehavior = inBehavior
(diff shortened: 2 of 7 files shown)


src/scenic/syntax/ast.py
```diff
@@ -5,6 +5,8 @@
 class AST(ast.AST):
     "Scenic AST base class"
 
+    _attributes = ("lineno", "col_offset", "end_lineno", "end_col_offset")
```
shumbo, line 8: `_attributes` were needed for `copy_location` and other utilities functions from the `ast` module to properly function.

https://github.com/python/cpython/blob/f6fd8aac13714ce17650eb4a648d5c08f0be53b4/Lib/ast.py#L178-L192

https://github.com/python/cpython/blob/f6fd8aac13714ce17650eb4a648d5c08f0be53b4/Lib/ast.py#L195-L227

src/scenic/syntax/compiler.py
```diff
@@ -193,6 +226,69 @@ def __init__(self) -> None:
         self.usedBreak = False
         self.usedContinue = False
 
+    def makeSyntaxError(self, msg, node: ast.AST) -> SyntaxError:
```
shumbo, line 229: construct a SyntaxError with the given error message and location of `node`

tests/syntax/test_errors.py
```diff
@@ -307,17 +308,16 @@ def runFile(path):
 
 @pytest.mark.parametrize('bug', (
     # BUGGY CODE                    ERROR CAUGHT DURING:
-    'x = 3 << 2',                   # token translation
```
dfremont, line 310: If there are still Python token types disallowed in Scenic, let's continue to have a test case like this one just in case such errors ever get caught differently than other syntax errors (you can change the comment of course).

tests/syntax/test_errors.py
```diff
@@ -307,17 +308,16 @@ def runFile(path):
...
     '4 = 2',                        # Python parsing
     '3 relative to',                # Python parsing
-    'Point at x y',                 # Python parsing (with offset past end of original line)
+    'new Point at x y',                 # Python parsing (with offset past end of original line)
```
dfremont, line 313: Please fix the alignment of these comments, and update the categories: this one should be "Scenic parsing" now, and the parenthesis no longer makes sense.

dfremont (changes requested):
Looks good -- just some minor comments.

tests/syntax/test_errors.py
shumbo: @dfremont Not sure if I understand your right. My understanding is that `<<` and other infix operators were disallowed in Scenic because they were used for substitution, but with the new parser, they can be used as plain Python operators. Are there tokens that should not be allowed? I'm not aware of one.

tests/syntax/test_errors.py
shumbo: Fixed in 4a046ec9ee90b764d367a17d55e7c4f20cedcdb5

tests/syntax/test_errors.py
dfremont: You're right, there are no tokens still disallowed, so this point is moot.

dfremont (approved):

PR #97, opened 2022-09-25
[Parser] implement `simulator` statements & fix simulator tests

shumbo:
This PR implements `simulator` statements and fixes tests under `./tests/simulator`.

### DIFF
--- a/src/scenic/core/object_types.py
+++ b/src/scenic/core/object_types.py
@@ -43,10 +43,6 @@ class Constructible(Samplable):
 		# find all defaults provided by the class or its superclasses
 		allDefs = collections.defaultdict(list)
 
-		# catch multiple inheritance of Scenic classes
-		if len(bases := [base.__name__ for base in cls.__bases__ if issubclass(base, Constructible)]) > 1:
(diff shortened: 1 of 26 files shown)


src/scenic/core/object_types.py
```diff
@@ -43,10 +43,6 @@ def __init_subclass__(cls):
 		# find all defaults provided by the class or its superclasses
 		allDefs = collections.defaultdict(list)
 
-		# catch multiple inheritance of Scenic classes
```
shumbo, line 46: We found that tests for `carla` and `lgsvl` do not follow the single-inheritance rule and decided to allow multiple subclasses.

src/scenic/syntax/compiler.py
```diff
@@ -449,7 +449,7 @@ def transformPropertyDef(self, node: s.PropertyDef):
                 ),
                 ast.Lambda(
                     args=selfArg,
-                    body=self.visit(node.value),
+                    body=node.value,
```
shumbo, line 452: `node.value` is visited in line 439. Removing this `visit` to avoid visiting the same node twice as some transformations are not idempotent.

dfremont (approved):

PR #98, opened 2022-09-26
[Parser] migrate `core` tests to new syntax

shumbo:
This PR migrates tests under `tests/core` and fixes the issue found through the migration.


dfremont (approved):
Looks great! Nice catch in `Constructible.dumpAsScenicCode` -- I would have totally forgotten that. Shows the value of having tests!

PR #109, opened 2023-03-04
Set heading action for the Newtonian simulator actors

abol-karimi:


Eric-Vin:
Hi Abel, that test that failed is an unfortunate edge case in our roads library that I don't think has anything to do with your changes. I'll rerun the test suite and see if it goes away.

Eric-Vin (approved):
Changes look good to me and fill the gap we currently have since we already support `setPosition`, `setVelocity`, etc...

dfremont:
When adding a new action to the driving domain, you need to add implementations to all simulator interfaces which support that domain. Looks like CARLA and LGSVL are missing in this PR.

abol-karimi:
> When adding a new action to the driving domain, you need to add implementations to all simulator interfaces which support that domain. Looks like CARLA and LGSVL are missing in this PR.

Since Carla already supports mutating the heading via set-transform, how about we remove the set-heading action from the driving domain?

dfremont:
I think it would be better to add it to the driving domain, since it makes sense for all simulators; I think `SetTransformAction` was meant as a temporary CARLA-specific hack (really we should provide something like it in the driving domain, but that can be a later PR). I guess since we've deprecated LGSVL you'd just have to implement `setHeading` for `CarlaActor`, which should be a simple matter of adapting the code in `SetTransformAction`.

abol-karimi:
Sounds good, I'll update my PR.

abol-karimi:
Since Carla does not have a dedicated setter for actor rotations, in a `SetHeadingAction` we would need to set the location as well too. So I propose that instead, we bring the `SetTransformAction` to the driving model, but with a different name, `SetPoseAction`, since `pose` is a standard term in robotics for combining position and orientation. If you agree, I'll write up the code.

abol-karimi:
I defined the `SetPoseAction` in the driving domain, together with its newtonian and carla implementations, but assuming that `2dMode` is active and the heading argument is a float. If we want to allow the driving domain to have the option of running in 3D mode, I'm not sure how to best design/implement it.

PR #115, opened 2023-05-03
Document temporal requirements

shumbo:
This PR adds documentation for temporal requirements.

Things to discuss:

- How do we want to update the dynamic scenarios tutorial?

### DIFF
--- a/docs/reference/dynamic_scenarios.rst
+++ b/docs/reference/dynamic_scenarios.rst
@@ -16,26 +16,24 @@ Behaviors are also suspended when they invoke a sub-behavior using :keyword:`do`
 When a behavior is first invoked, its preconditions are checked, and if any are not satisfied, the simulation is rejected, requiring a new simulation to be sampled. [#f1]_
 The behavior's invariants are handled similarly, except that they are also checked whenever the behavior is resumed (i.e. after taking an action and after a sub-behavior terminates).
 
-:term:`Monitors` and :keyword:`compose` blocks of :term:`modular scenarios` execute in the same way as behaviors, with the latter also including additional checks to see if any of their :keyword:`terminate when` conditions have been met or their safety properties violated.
+:term:`Monitors` and :keyword:`compose` blocks of :term:`modular scenarios` execute in the same way as behaviors, with :keyword:`compose` blocks also including additional checks to see if any of their :keyword:`terminate when` conditions have been met or their :term:`temporal requirements` violated.
 
 In detail, a single time step of a dynamic simulation is executed according to the following procedure:
 
 1. Execute all currently-running :term:`modular scenarios` for one time step.
    Specifically, for each such scenario:
 
-	a. Check if any of its safety properties are violated; if so, reject the simulation.
+	a. Check if any of its :term:`temporal requirements` have already been violated [#f2]_; if so, reject the simulation.
 
-	b. Check which liveness properties are satisfied; remember these for later.
+	b. Check if the scenario's time limit (if :keyword:`terminate after` has been used) has been reached; if so, go to step (e) below to stop the scenario.
 
-	c. Check if the scenario's time limit (if :keyword:`terminate after` has been used) has been reached; if so, go to step (f) below to stop the scenario.
+	c. If the scenario is not currently running a sub-scenario (with :keyword:`do`), check its invariants; if any are violated, reject the simulation. [#f1]_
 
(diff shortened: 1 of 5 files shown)


docs/reference/dynamic_scenarios.rst
```diff
@@ -65,7 +65,7 @@ In detail, a single time step of a dynamic simulation is executed according to t
 
 9. Update every :term:`dynamic property` of every object to its current value in the simulator.
 
-10. If the simulation is stopping for one of the reasons above, first check if any of the :keyword:`require eventually` conditions of any remaining scenarios were never satisfied: if so, reject the simulation.
+10. If the simulation is stopping for one of the reasons above, first check if any of the liveness properties of any remaining scenarios were never satisfied: if so, reject the simulation.
```
shumbo, line 68: I didn't understand what this meant but systematically replaced "require eventually conditions" with "liveness properties". I want to make sure this makes sense.

docs/reference/operators.rst
```diff
@@ -154,3 +154,51 @@ The midpoint of the corresponding edge of the Object’s bounding box, inheritin
...
+.. _next:
+
+next *condition*
+----------------
+Require the given condition to hold at the next time step of the dynamic scenario.
+
+.. _{condition} until {condition}:
+.. _until:
+
+*condition* until *condition*
```
shumbo, line 190: Is there a name for LHS and RHS of the until operator, like "hypothesis" and "conclusion" for `implies`?

docs/reference/dynamic_scenarios.rst
Eric-Vin: Makes sense to me! A liveness property is something that must occur in an execution, so "require eventually" is how we implement such a property.

docs/reference/operators.rst
Eric-Vin: The closest I can think of would be precondition/postcondition, but this doesn't seem to exactly meet their definitions in the contexts I've heard them used.

dfremont: I don't know of specific names for these -- what you have is clear, I think.

docs/reference/operators.rst
```diff
@@ -154,3 +154,51 @@ The midpoint of the corresponding edge of the Object’s bounding box, inheritin
...
+.. _{condition} implies {condition}:
+.. _implies:
+
+*hypothesis* implies *conclusion*
+---------------------------------
+Require the hypothesis to hold whenever the conclusion holds. This is a syntax sugar for :scenic:`not {hypothesis} or {conclusion}`.
+
+.. rubric:: References
+
+.. [B10] Bauer et al., :t:`Comparing LTL Semantics for Runtime Verification`, Journal of Logic and Computation. `[Online] <https://doi.org/10.1093/logcom/exn075>`_
```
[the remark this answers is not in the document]
dfremont: This is great. In fact I didn't know about the [title-reference role](https://docutils.sourceforge.io/docs/ref/rst/roles.html#title-reference), and will need to stick `:title:` in various other places in the docs!

docs/reference/operators.rst
```diff
@@ -154,3 +154,51 @@ The midpoint of the corresponding edge of the Object’s bounding box, inheritin
...
+-----------------------------
+Require the first condition to hold until the second condition holds, at which point the first condition is no longer required to hold.
+This operator implements the strong semantics and requires the second condition to hold at some point during the execution.
+
+.. _{condition} implies {condition}:
+.. _implies:
+
+*hypothesis* implies *conclusion*
+---------------------------------
+Require the hypothesis to hold whenever the conclusion holds. This is a syntax sugar for :scenic:`not {hypothesis} or {conclusion}`.
```
dfremont, line 200: Should be "require the conclusion to hold whenever the hypothesis holds".

docs/reference/dynamic_scenarios.rst
dfremont: We should be careful here: a property like `(always X) and (eventually Y)` is neither a safety nor a liveness property (violations of it don't necessarily have a finite prefix after which the property can't be satisfied, nor is it always possible to extend a finite trace to one where the property is satisfied), but it needs to be checked at the end of the simulation. I'll revise this text.

docs/reference/operators.rst
dfremont: Actually that sounds more like `always X implies Y`, so I'll rephrase.

dfremont:
I've pushed some revisions, including updating the dynamics tutorial. If it looks OK to you we can merge.

PR #122, opened 2023-05-17
Parse prefix temporal operators followed by `and` and `or`

shumbo:
The Python grammar (which we used as a base for Scenic grammar) is structured so it does not allow prefix operators after `and` or `or` without parenthesis. In Python, the only such prefix operator is `lambda`, and if you do 

```python
fn = False or lambda x: x
```

You might expect this to parse and assign the identity function to `fn`, but it just raises a syntax error.

This is problematic in Scenic because we have prefix temporal operators that we want to use together with `and` and `or`. For example, users might want to write

```scenic
require X or eventually Y
```

It failed to parse with the current grammar. This PR fixes the issue.

### DIFF
--- a/src/scenic/syntax/scenic.gram
+++ b/src/scenic/syntax/scenic.gram
@@ -1441,11 +1441,11 @@ scenic_implication (memo):
     | disjunction
 
 disjunction (memo):
-    | a=conjunction b=('or' c=conjunction { c })+ { ast.BoolOp(op=ast.Or(), values=[a] + b, LOCATIONS) }
+    | a=conjunction b=('or' c=(scenic_temporal_prefix | conjunction) { c })+ { ast.BoolOp(op=ast.Or(), values=[a] + b, LOCATIONS) }
     | conjunction
 
 conjunction (memo):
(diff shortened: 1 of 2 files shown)


src/scenic/syntax/scenic.gram
```diff
@@ -1441,11 +1441,11 @@ scenic_implication (memo):
     | disjunction
 
 disjunction (memo):
-    | a=conjunction b=('or' c=conjunction { c })+ { ast.BoolOp(op=ast.Or(), values=[a] + b, LOCATIONS) }
+    | a=conjunction b=('or' c=(scenic_temporal_prefix | conjunction) { c })+ { ast.BoolOp(op=ast.Or(), values=[a] + b, LOCATIONS) }
```
shumbo, line 1444: This change allows `scenic_temporal_prefix` to appear in the second operand and onward of `or`. The temporal prefix operators cannot appear in the first operand because `always x or y` should be parsed as `always (x or y)` and that part is already done by the rules above.

This results in ambiguous grammar because `x or always y or z` can be parsed as either `x or (always y or z)` (this is what we want) and `x or (always y) or z` (not what we want, but there is a derivation). However, Pegen uses whatever matches first, so in this case, we get the former. I added test cases to make sure the grammar can handle this case.

dfremont (approved):
Good catch!

Issue #120, opened 2023-05-15
Parallel Scene Generation


axelbr:
Hello!
Is it somehow possible to sample scenes in parallel? 
I am trying to reduce the time it takes to generate multiple scenes up front by parallelizing the sampling process over multiple processes.
Something similar to:
```python
def generate(scenario):
  scene, _ = scenario.generate()
  return scene

...

with Pool(len(scenarios)) as pool:
  scenes = pool.map(generate, scenarios)
```

Trying this will result in an AssertionError which originates from the assertion `not veneer.isActive()`. It seems like scenic relies on a bunch of global variables for the sampling process. Is there a way to encapsulate them and to run the sampling process in parallel?

Eric-Vin:
It is possible to sample multiple scenes in parallel by offloading the compilation, as well as the sampling, to a separate process. Here's some code that would do it:

```
def generate(path):
	scenario = scenic.scenarioFromFile(path=path)
	scene, _ = scenario.generate()
	return scene

with Pool(num_cores) as pool:
	scenes = pool.map(generate, [path for _ in range(num_scenarios)])
```

Note that you'll need to use `multiprocess`, not `multiprocessing`, as `pickle` is currently unable to pickle Scenic `Scene` objects while `dill` is.

These is some repeated work in the compilation/pruning of the scenarios, but hopefully this is negligible in your use case compared to the actual sampling. We are working on a better integration of parallel sampling into Scenic, which would avoid this repeated work.

Hope this helps!

axelbr:
Thanks a lot!
Yes, I had to use multiprocess as well to avoid the pickling problem.

axelbr:
And it is not an issue with the scenario specification. I tried several of your examples with it.

Eric-Vin:
Hmm, could you provide the Python program (using one of our example Scenic programs for simplicity) that causes the problem?

axelbr:
Sorry for the delay.
I am using Scenic version 2.1.0b4.
Here's some sample code that should reproduce the error:

```python
import scenic
from multiprocess.pool import Pool
from scenic.core.scenarios import Scene, Scenario
from scenic.simulators.carla import CarlaSimulator


def sample_scene(scenario: Scenario) -> Scene:
    # scenario = scenic.scenarioFromFile(scenario_path)
    scene, _ = scenario.generate()
    return scene


def sample_sequentially(scenario_path: str, num_scenes=10):
[...]

PR #123, opened 2023-05-17
Fix leak across multiple compilations of driving scenarios

dfremont:
Fixes the second issue raised in #120: compiling a scenario using the driving model two times and then trying to run a simulation failed with an error claiming an ordinary steering action was incompatible with a car.

This issue was caused by the `scenic.domains.driving.actions` module keeping a reference to the Scenic module `scenic.domains.driving.model` in order to access the `Steers` mixin: the latter module gets purged at the end of compilation, so the second compilation imports it again and gets a different copy of `Steers`. But the former module is not purged since it's an ordinary Python module, so the actions used in the scenario use a different version of `Steers` than the objects, causing the `isinstance` check used in `SteeringAction.canBeTakenBy` to fail.

### DIFF
--- a/src/scenic/domains/driving/actions.py
+++ b/src/scenic/domains/driving/actions.py
@@ -17,7 +17,39 @@ import math
 
 from scenic.core.vectors import Vector
 from scenic.core.simulators import Action
-import scenic.domains.driving.model as _model
+
+## Mixin classes indicating support for various types of actions.
+
+class Steers:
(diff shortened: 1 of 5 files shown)


Eric-Vin (approved):
Looks good to me (apart from the weird test failure which seems unrelated)

dfremont:
Agreed that the test failure is unrelated -- I will merge this to fix Axel's problem and look at the failure separately.

PR #125, opened 2023-05-20
2to3 converter maps "with heading" to "facing"

dfremont:
These were equivalent in Scenic 2, but the former is illegal in Scenic 3 (as `heading` is now final, derived from `parentOrientation` and `yaw` etc.).


dfremont:
(Note: the test failure is the same one we saw in #123; it should be fixed by 4c057e79ee28caedb2404a65e2eb1ca242901e8a).

Eric-Vin (approved):

PR #128, opened 2023-05-23
Only recognize temporal operators inside requirements

dfremont:
Thereby allowing normal use of the built-in `next` function, for example (outside of requirements).

We didn't do this before because the straightforward implementation would require having two copies of all of the expression rules (for normal expressions, and expressions inside requirements allowing temporal operators). The implementation in this PR avoids that by using a slightly-hacky parsing rule which allows too much (specifically parsing the temporal operators as function calls in illegal code like `(always (X)) > 5`) plus a check in the compiler for the illegal cases. This approach is the least painful I could come up with using straight PEG syntax; if Pegen later adds predicates or other metaprogramming features we could switch to those.

### DIFF
--- a/docs/internals/compiler.rst
+++ b/docs/internals/compiler.rst
@@ -135,12 +135,12 @@ easily distinguish Scenic-specific rules from those in the original Python gramm
 .. code-block:: pegen
 
    scenic_implication (memo):
-       | invalid_scenic_implication  # special rule to explain invalid uses of 'implies'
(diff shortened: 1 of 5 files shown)


shumbo (approved):
Thank you so much for working on this @dfremont. This looks great.

One minor thing is that, now that we only parse temporal operators inside requirements, we should be able to remove the following checks in the compiler.

https://github.com/BerkeleyLearnVerify/Scenic/blob/7cffad54bae2c01782767ab676754b99e8f3d098/src/scenic/syntax/compiler.py#L1746-L1770

PR #131, opened 2023-05-27
Trim Overlapping Polygons in Road Construction

Eric-Vin:
Small code fix that fixes the test crash we keep seeing by checking for overlapping road polygons and taking their difference followed by eroding them very slightly. Hopefully this does a good job of fixing the problem in general, though I'm not 100% sure this is the best spot for it.


Eric-Vin (commented):
It's my PR so I can't actually approve, but approved!

dfremont (approved):

Issue #132, opened 2023-05-29
Not able to run OAS_Scenario


sumitradagar:
![image](https://github.com/BerkeleyLearnVerify/Scenic/assets/37659902/b5235d44-623e-4ebe-b455-776d72203711)

dfremont:
Hello Sumitra, from the traceback it looks like you're using Python 3.7 -- could you please confirm?

sumitradagar:
Yes sir I'm using python 3.7

dfremont:
OK, could you please also send the exact command you ran that produced the error, and the backtrace you get if you run the command adding the `-b` option?

sumitradagar:
![image](https://github.com/BerkeleyLearnVerify/Scenic/assets/37659902/b77a4195-0b54-44e3-8879-a198080ea3d3)

dfremont:
Hi Sumitra, can you please run the command again using the `-b` option and send the backtrace? (Please copy-paste the text into a code block in your comment rather than including a screenshot, so that other people who get the same error can search for its text. Thanks!)

sumitradagar:
(mywork) C:\Users\Sumitra\work\Scenic>scenic examples\carla\OAS_Scenarios\oas_scenario_05.scenic -b
Beginning scenario construction...
  Loading cached version of road network...
  Loaded cached network in 0.26 seconds.
Traceback (most recent call last):
  File "C:\Users\Sumitra\work\Scenic\src\scenic\core\utils.py", line 22, in wrapper
    return self.__getattribute__(storageName)
  File "C:\Users\Sumitra\work\Scenic\src\scenic\core\object_types.py", line 533, in __getattribute__
    return object.__getattribute__(proxy, name)
AttributeError: 'Car' object has no attribute '_cached_visibleRegion'

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "C:\Program Files (x86)\Microsoft Visual Studio\Shared\Python37_64\lib\runpy.py", line 193, in _run_module_as_main
    "__main__", mod_spec)
  File "C:\Program Files (x86)\Microsoft Visual Studio\Shared\Python37_64\lib\runpy.py", line 85, in _run_code
    exec(code, run_globals)
  File "C:\Users\Sumitra\work\mywork\Scripts\scenic.exe\__main__.py", line 4, in <module>
    from scenic.__main__ import dummy
[...]

PR #134, opened 2023-06-01
Work around bpo-37838

dfremont:
This should fix the issue in #132 on old versions of Python.

I tested this by unconditionally patching `typing.get_type_hints`, but the PR only does the patching when the version of Python is old enough to suffer from bpo-37838. @Eric-Vin since I can't install an old-enough version under macOS, and our CI probably doesn't use one either, can you please see if you can do it on your machine?


dfremont:
I've tested that this fixes the issue in #132 on both Python 3.7.5 and 3.8.0.

Eric-Vin (approved):

PR #137, opened 2023-06-14
Revise simulator API

dfremont:
This PR makes a few changes to the simulator API. Most importantly, the initializer of `Simulation` now actually runs the simulation, so that the `run` method is unnecessary. Simulator-specific initialization is moved to the new `setup` method, which by default just creates all of the initial objects by calling `createObjectInSimulator` (so that the latter method is now used for *all* objects, not only objects created dynamically in the middle of a simulation).

I haven't yet tested this with CARLA or with LGSVL (the latter may be tough at this point).


Eric-Vin (approved):

src/scenic/simulators/carla/simulator.py
```diff
@@ -57,45 +57,53 @@ def __init__(self, carla_map, map_path, address='127.0.0.1', port=2000, timeout=
         self.record = record  # whether to use the carla recorder
         self.scenario_number = 0  # Number of the scenario executed
 
-    def createSimulation(self, scene, verbosity=0):
+    def createSimulation(self, scene, *, timestep, **kwargs):
+        if timestep != self.timestep:
```
Eric-Vin, line 61:
```suggestion
        if timestep is not None and timestep != self.timestep
```

PR #138, opened 2023-06-20
Sample checking generalization

Eric-Vin:
Generalizes how default requirements are encoded and how samples are checked. Also adds `generateBatch` method.


src/scenic/core/requirements.py
```diff
@@ -258,4 +237,141 @@ def __str__(self):
...
+    def __init__(self, objects, optional=True):
+        super().__init__(optional=optional)
+        self.objects = objects
+
+    def falsifiedByInner(self, sample):
+        objects = tuple(sample[obj] for obj in self.objects)
+        cm = trimesh.collision.CollisionManager()
+        for obj in objects:
+            if not obj.allowCollisions:
+                cm.add_object(str(obj), obj.occupiedSpace.mesh)
```
dfremont, line 298: The trimesh [docs](https://trimsh.org/trimesh.collision.html#trimesh.collision.CollisionManager.add_object) say this method will replace an existing object with the same name, which could easily happen (as the default name for an object is "unnamed Object"). Probably better to use `for i, obj in enumerate(objects)` above and then use `str(i)` as the name.

src/scenic/core/sample_checking.py
```diff
@@ -0,0 +1,108 @@
...
+from abc import ABC, abstractmethod
+import time
+
+from scenic.core.requirements import IntersectionRequirement
+
+class SampleChecker(ABC):
+    def __init__(self):
+        self.requirements = None
+
+    def addRequirements(self, requirements):
```
dfremont, line 12: Let's call this `setRequirements` since the current name suggests that you could call it multiple times and accumulate requirements.

src/scenic/core/sample_checking.py
```diff
@@ -0,0 +1,108 @@
...
+            metrics = (int(not rejected), time.time()-start)
+
+            self.updateMetrics(req, metrics)
+
+            if rejected:
+                return req.violationMsg
+
+        return None
+
+    def sortRequirements(self):
```
dfremont, line 86: This has no side effects, so it should be called `sortedRequirements`.

src/scenic/core/sample_checking.py
```diff
@@ -0,0 +1,108 @@
...
+
+    def addRequirements(self, requirements):
+        super().addRequirements(requirements)
+
+        self.buffers = {req:[(1,0)]*self.bufferSize for req in self.requirements}
+
+    def checkRequirementsInner(self, sample):
+        for req in self.sortRequirements():
+            # Evaluate the requirement with timing info.
+            start = time.time()
```
dfremont, line 74: Should use `time.perf_counter()` for these times since it's monotonic (and in theory `time()` can have very poor resolution). I'm guilty of using `time()` elsewhere -- we should probably go through and fix that!

src/scenic/core/sample_checking.py
```diff
@@ -0,0 +1,108 @@
...
+        while reqs and reqs[-1].optional:
+            reqs.pop()
+
+        return reqs
+
+    def updateMetrics(self, req, metrics):
+        """ Update the metrics for a given requirement"""
+        target_buffer = self.buffers[req]
+        target_buffer.pop()
+        target_buffer.insert(0, metrics)
```
dfremont, line 102: Better to use `collections.deque` with `maxlen=bufferSize`.

src/scenic/core/requirements.py
Eric-Vin: Good catch!

src/scenic/core/sample_checking.py
Eric-Vin: Ah very cool, I didn't know this existed!

src/scenic/core/sample_checking.py
Eric-Vin: Very neat! Adding it now.

src/scenic/core/sample_checking.py
```diff
@@ -0,0 +1,108 @@
...
+
+    def getWeightedAcceptanceProb(self, req):
+        target_buffer = self.buffers[req]
+        mean_accept = sum(metric[0] for metric in target_buffer)/self.bufferSize
+        mean_time = sum(metric[1] for metric in target_buffer)/self.bufferSize
```
dfremont, line 107: It's probably not worth the complexity for small buffers, but if you try larger ones later you might want to maintain the moving averages without a linear summation each time (see the `collections.deque` recipe for a relatively clean way to implement it).

src/scenic/core/scenarios.py
```diff
@@ -309,8 +323,48 @@ def generate(self, maxIterations=2000, verbosity=0, feedback=None):
...
+            `RejectionException`: if not enough valid samples are found in **maxIterations** iterations.
+        """
+        total_iterations = 0
+        scenes = []
+
+        for _ in range(numScenes):
+            try:
+                scene, iterations = self._generateInner(maxIterations, verbosity, feedback)
+                scenes.append(scene)
+                total_iterations += iterations
```
dfremont, line 355: Need to also decrease `maxIterations` by `iterations` if `maxIterations` is supposed to be a total over all the scenes (as stated in the docstring).

src/scenic/core/scenarios.py
```diff
@@ -331,7 +385,7 @@ def generate(self, maxIterations=2000, verbosity=0, feedback=None):
                 sample = Samplable.sampleAll(self.dependencies)
             except RejectionException as e:
                 optionallyDebugRejection(e)
-                rejection = e
+                rejection = f"Rejection exception: {e}"
```
dfremont, line 388: "Rejection exception" isn't meaningful to ordinary users, so I'd just keep this as `e` (the string form of the exception is supposed to be an explanatory string).

src/scenic/core/sample_checking.py
Eric-Vin: Good point, I'll keep it in mind!

src/scenic/core/scenarios.py
```diff
@@ -350,77 +404,44 @@ def generate(self, maxIterations=2000, verbosity=0, feedback=None):
...
+        # Object containment
+        for obj in self.objects:
+            container = self.containerOfObject(obj)
+            if not isinstance(container, AllRegion):
+                requirements.append(ContainmentRequirement(obj, container))
+
+        # Observing entity visibility
+        for obj in filter(lambda x: x._observingEntity is not None, self._instances):
+            requirements.append(VisibilityRequirement(obj._observingEntity,
+                obj, self.objects))
```
dfremont, line 430: Might as well save time later by filtering out known non-occluding objects here, like you do with `colliding_objects` above.

Eric-Vin: Good call.

tests/syntax/test_requirements.py
```diff
@@ -315,3 +315,42 @@ def test_static_intersection_violation_disabled():
...
         new Object at 1@0, with allowCollisions True
     """)
+
+## Tests for random properties that have special effects on constructing requirements
+def test_random_allowCollisions():
+    for _ in range(5):
+        sampleSceneFrom("""
+            new Object with allowCollisions Uniform(True, False)
+            new Object with allowCollisions Uniform(True, False)
+        """, maxIterations=10)
```
dfremont, line 325: Each iteration of the loop has about a 1 in 10^6 chance of failing, which is too high (this test may get run many times in CI, etc.). I'd increase `maxIterations` to 30. This doesn't slow the test down in any case.

I'm not sure what the outer loop is for: if the point is to ensure we get at least one round where more than 1 iteration is necessary, we'd need more repetitions to have a good probability of that. (But you could break out of the loop as soon as one of the calls uses more than 1 iteration.)

tests/syntax/test_requirements.py
```diff
@@ -315,3 +315,42 @@ def test_static_intersection_violation_disabled():
...
+            ego = new Object with visibleDistance 30,
+                at (0,0,1),
+                with width 5,
+                with length 5,
+                with height 5,
+                with pitch 45 deg,
+                with viewAngles (340 deg, 60 deg),
+                with rayDensity 5
+
+            seeing_obj = new Object at (0,10,5),
```
dfremont, line 343: Would be clearer to call this `seen_obj` (and change `name` below accordingly), since it's in fact `ego` that is doing the seeing.

tests/syntax/test_requirements.py
```diff
@@ -315,3 +315,42 @@ def test_static_intersection_violation_disabled():
...
+def test_random_allowCollisions():
+    for _ in range(5):
+        sampleSceneFrom("""
+            new Object with allowCollisions Uniform(True, False)
+            new Object with allowCollisions Uniform(True, False)
+        """, maxIterations=10)
+
+@pytest.mark.slow
+def test_random_occlusion():
+    for _ in range(5):
```
dfremont, line 329: As above, I'm not sure that just repeating the test 5 times is enough. It would probably also be good to add an assertion that in all returned scenes the wall has `occluding` false.

tools/benchmarking/sampleChecking/benchmark_sample_checking.py
```diff
@@ -0,0 +1,107 @@
...
+                p.start()
+                p.join(timeout=MAX_TIME)
+                p.kill()
+                trial_time = min(time.time()-start, MAX_TIME)
+
+                print(f"({sample_checker},{benchmark},{num_scenes},{trial_iter}): {trial_time:.2f}s")
+
+                total_time += trial_time
+
+            mean_time = total_time/TRIALS_PER
```
dfremont, line 83: Median would be more robust here. (You can use `statistics.median` on a list of the times.)

tools/benchmarking/sampleChecking/benchmark_sample_checking.py
```diff
@@ -0,0 +1,107 @@
...
+    sc_results = {}
+
+    for sample_checker in SAMPLE_CHECKERS:
+        results_val = {}
+        for benchmark, num_scenes in itertools.product(BENCHMARKS, SCENE_COUNT):
+            benchmark_name, benchmark_params = benchmark
+            total_time = 0
+            for trial_iter in range(TRIALS_PER):
+                p = Process(target=run_benchmark, args=[benchmark_name, benchmark_params, sample_checker, num_scenes])
+                start = time.time()
```
dfremont, line 73: Use `time.perf_counter()`

dfremont (changes requested):
Looks good overall. My main suggestion is to not include scenario compilation time in the benchmark timings; other comments above are minor.

dfremont (approved):

PR #153, opened 2023-07-07
Work around division by zero in colorsys

dfremont:
`colorsys.rgb_to_hls` divides by zero when given a color which is very nearly but not exactly white. Surprisingly, this doesn't seem to have been noticed before (at least I can't find anything online), so I'll file a bug. But in the mean time I'm adding this workaround, since the bug caused one of our tests to randomly fail.


Eric-Vin (approved):

Eric-Vin:
Code changes look good. The test failures look like the road polygons again, so we can either go ahead and merge or merge main into this branch (after the polygons fix is in) and rerun the CI.

dfremont:
OK, thanks. For future reference, I reported the `colorsys` bug at python/cpython#106498.

PR #156, opened 2023-07-10
Mesh Volume Intersects Heuristic

Eric-Vin:
New heuristic for uni-body mesh volume intersects checks. Also adds preprocessing to try to consolidate non-disjoint bodies in mesh volumes and mesh shapes, so that we can use the new heuristic more often.


src/scenic/core/regions.py
```diff
@@ -1185,6 +1189,20 @@ def intersects(self, other, triedReversed=False):
                 return surface_collision
 
             # PASS 4
+            # If we have 2 candidate points and both regions have only one body,
+            # we can just check if either region contains the candidate point of the
+            # other.
```
dfremont, line 1194: To make the reasoning clearer I'd add something like "(since surface intersections were ruled out above, so if the regions intersect then one contains the other)".

dfremont (changes requested):
Looks good, but let's fix the dining table and coffee table meshes too: the `vacuum.scenic` scenario now takes 15 seconds to compile instead of 1 second, which is a shame since the sampling time is almost 3x better. (The compile time is important too since people playing around with the examples will have to sit through it.) I left some other minor comments above.

src/scenic/core/utils.py
```diff
@@ -143,6 +144,35 @@ def loadMesh(path, filetype, compressed, binary):
...
+    if mesh.body_count < 2:
+        return mesh
+
+    mesh_bodies = mesh.split()
+
+    if not all(m.is_volume for m in mesh_bodies):
+        return mesh
+
+    try:
+        unified_mesh = trimesh.boolean.union(mesh_bodies, engine="scad")
```
dfremont, line 164: Actually, why are we using OpenSCAD here when the default for `MeshVolumeRegion.union` is Blender?

src/scenic/core/regions.py
```diff
@@ -1185,6 +1192,20 @@ def intersects(self, other, triedReversed=False):
...
+            # we can just check if either region contains the candidate point of the
+            # other. (This is because we previously ruled out surface intersections)
+            if (
+                s_candidate_point is not None
+                and o_candidate_point is not None
+                and self.mesh.body_count == 1
+                and other.mesh.body_count == 1
+            ):
+                return self.containsPoint(o_candidate_point) or other.containsPoint(
+                    s_candidate_point
```
dfremont, line 1205: This is a pretty awful output of `black`. If you put parentheses around everything, can you get it to do something like
```python
return (
    self.containsPoint(o_candidate_point)
    or other.containsPoint(s_candidate_point)
)
```
instead?

Eric-Vin: Unfortunately black won't let me change it :(

dfremont (approved):

PR #163, opened 2023-07-31
Deprecated LGSVL interface and Updated Docs.

Eric-Vin:


src/scenic/domains/driving/__init__.py
```diff
@@ -1,5 +1,7 @@
 """Domain for driving scenarios.
 
+This domain must be used with 2D compatibility mode.
```
dfremont, line 3:
```suggestion
This domain must be used in `2D compatibility mode`.
```

(Likewise for the identical notes in the other files. If the link doesn't connect automatically you might have to add `:ref:` in front, but that shouldn't be necessary.)

dfremont: Actually for the places where we expect to support 3D mode relatively soon, i.e. here and CARLA, let's put "must currently be used in" and add that we expect to add 3D support soon.

dfremont (changes requested):
Minor comments above. Two larger suggestions:
* Removing the LGSVL examples. Having them included is a bit misleading since you can't actually run them, and they'll still be in the repo in the 2.x branch.
* Adding a "Deprecated Simulators" list after the existing list at the top of the "Supported Simulators" page, with notes explaining when support was dropped for each simulator and why. That way people can easily see that we used to support LGSVL and figure out which version they have to roll back to if necessary. In the particular case of LGSVL/"SVL Simulator", we can say that the original simulator is no longer usable due to LG shutting down its cloud service, but that we might accept a PR targeting one of its forks.

dfremont (approved):

PR #164, opened 2023-08-16
Fix `_fields` on AST nodes

shumbo:
There were some errors in the `_fields` property on AST nodes. This PR fixes all inconsistencies in the field.


Eric-Vin (approved):
Looks good to me! (but I'm no parser expert)

dfremont (approved):

PR #167, opened 2023-08-25
Mesh Repair Function

Eric-Vin:


Eric-Vin:
Before and after on an Airsim drone repaired via the voxelization method. Note the before example is not actually watertight (which necessitates the repair), though it appears so. The voxelized version is water tight and a proper volume.

drone_before
drone_after

src/scenic/core/utils.py
```diff
@@ -216,6 +217,95 @@ def unifyMesh(mesh, verbose=False):
...
+                )
+
+            # Center new mesh
+            new_mesh.vertices -= new_mesh.bounding_box.center_mass
+
+            # Rescale mesh to original size
+            orig_scale = new_mesh.extents / dims
+            scale_matrix = numpy.eye(4)
+            scale_matrix[:3, :3] /= orig_scale
+            new_mesh.apply_transform(scale_matrix)
```
dfremont, line 289: This could result in non-uniform scaling, right? Wouldn't it be better to apply the inverse of the `scale_matrix` you used above? (Likewise for the translation.)

tests/core/test_utils.py
```diff
@@ -0,0 +1,52 @@
...
+import pytest
+import trimesh
+
+from scenic.core.utils import loadMesh, repairMesh
+
+
+@pytest.mark.slow
+def test_mesh_repair():
+    plane_mesh = loadMesh(
+        path=Path(".").parent.parent.parent
```
dfremont, line 13: You could define an `assets` fixture in `tests/conftest.py` which returns a `Path` object for the assets folder. Then you could just use `assets / "meshes" / "classic_plane.obj.bz2"` for the path and it would work regardless of where the test is located.

src/scenic/core/utils.py
```diff
@@ -216,6 +217,95 @@ def unifyMesh(mesh, verbose=False):
         return mesh
 
 
+def repairMesh(mesh, pitch=(1 / 2) ** 6, verbose=True):
```
dfremont, line 220: If we're going to recommend ordinary users use this function to fix their meshes, we'd better document the parameters and return value.

dfremont (changes requested):
Looks good overall -- just a couple minor comments above.

I'm not particularly happy with recommending people use a function in the `scenic.core.utils` module, since that module has various undocumented things meant for internal use only, but it's OK for now; during the region refactoring we should move it to `scenic.core.regions.meshes`.

src/scenic/core/utils.py
Eric-Vin: Hmm, I was thinking that it would be more important that we ensure the bounds are the same as the input mesh, even if it's slightly distorted. I haven't thought about the consequences of this in too much detail though.

tests/core/test_utils.py
Eric-Vin: Good idea!

Eric-Vin:
For some reason one of the CI jobs was taking 50 minutes, so I cancelled and restarted it.

src/scenic/core/utils.py
dfremont: Per our discussion, let's keep your approach so that the bounds are preserved, even if it means the mesh is slightly distorted. I was concerned the bounds could change dramatically if the voxelization drops thin parts of the mesh, but the default algorithm seems not to do that. In any case you do warn the user to check the result.

dfremont (approved):

PR #169, opened 2023-08-30
Move to Manifold for Boolean Operations

Eric-Vin:
Move from OpenSCAD/Blender to Manifold, which is faster, more reliable, and can (soon) be installed via PyPi.

This is a WIP, as I developed it using a locally built Manifold wheel and modified version of Trimesh. Once [this](https://github.com/mikedh/trimesh/pull/2018) PR and this [issue](https://github.com/elalish/manifold/issues/362) are implemented, I'll update this branch so that we can run CI and confirm everything works. 

I'll also need to update the installation instructions and documentation, and figure out how we can detect non 2-manifold meshes and present a clean error to users, as I believe Manifold will rightly error on those. Right now some of them are detected on our end, but others aren't and that leaves some undefined behavior.


Eric-Vin:
Confirmed that the tests pass with the updated CI workflows from this branch: https://github.com/BerkeleyLearnVerify/Scenic/actions/runs/7268486842

Eric-Vin:
Hmm, seems to exhibit an unrelated bug (and I believe it's the one we're working on patching in https://github.com/BerkeleyLearnVerify/Scenic/tree/OpendriveParser).

src/scenic/core/regions.py
```diff
@@ -3599,10 +3567,7 @@ class ViewRegion(MeshVolumeRegion):
       * Case 2.a    viewAngles[0] = 360 degrees     => Sphere
       * Case 2.b    viewAngles[0] < 360 degrees     => Sphere & CylinderSectionRegion
```
dfremont, line 3568: Numbering is off here.

src/scenic/core/utils.py
```diff
@@ -163,10 +163,15 @@ def loadMesh(path, filetype, compressed, binary):
 
 
 def unifyMesh(mesh, verbose=False):
-    """Attempt to merge mesh bodies, aborting if something fails.
+    """Attempt to merge mesh bodies, raising an error if something fails.
```
dfremont, line 166: Let's mention that it's `ValueError` (are there any other possibilities?) in case someone wants to catch it and do something else.

tests/core/test_utils.py
```diff
@@ -46,3 +46,21 @@ def test_mesh_repair(getAssetPath):
...
+        .union(trimesh.creation.icosphere(radius=3))
+        .difference(trimesh.creation.icosphere(radius=2))
+    )
+
+    # Manually append a box
+    bad_mesh = trimesh.util.concatenate(
+        nested_sphere, trimesh.creation.box(bounds=((0, 0, 0), (3, 5, 3)))
+    )
+
+    unifyMesh(bad_mesh)
```
dfremont, line 66: Would be nice to add some assertions here, e.g. that the unified mesh is a valid volume and has only one body.

dfremont (changes requested):
Looks great overall (hooray for simplifying the installation procedure!) -- just a few minor comments and our discussion offline about `unifyMesh`.

tests/core/test_utils.py
Eric-Vin: Good idea on the volume check, but unfortunately the unified mesh can actually still have two bodies and be valid (think two discrete spheres).

src/scenic/core/utils.py
Eric-Vin: Good point, added!

tests/core/test_utils.py
Eric-Vin: Good idea, though this one actually ends up with 3 bodies so I'll make the assertion reflect that.

Eric-Vin: Hmm, this is actually something we should discuss since this goes from 5 "bodies" (by face adjacency) to 3.

Eric-Vin:
Gah, this new failure looks to be related to [this](https://github.com/mikedh/trimesh/issues/2112). Looks like a [fix](https://github.com/mikedh/trimesh/pull/2113) is already in the pipeline though.

Eric-Vin:
Alright with the new fix and pinned Manifold version we should be good to merge.

dfremont (approved):

PR #171, opened 2023-08-31
Added Planar Error Case to Mesh Repair.

Eric-Vin:


dfremont (approved):

dfremont:
Note for future reference: this PR also fixes the bug in `repairMesh` which was causing the CI to hang sporadically.

Issue #168, opened 2023-08-30
Cannot run two scenarios back to back


abol-karimi:
Using Scenic's Python API, I cannot run a scenario immediately after another. For the second simulation, I get the following error:
```Traceback (most recent call last):
  File "tests/scenic2/loop.py", line 28, in <module>
    res2 = simulate()
  File "tests/scenic2/loop.py", line 19, in simulate
    sim_result = simulator.simulate(
  File "/home/carla/Scenic/src/scenic/core/simulators.py", line 76, in simulate
    simulation.run(maxSteps)
  File "/home/carla/Scenic/src/scenic/core/simulators.py", line 216, in run
    raise InvalidScenarioError(f'agent {agent} tried incompatible '
scenic.core.errors.InvalidScenarioError: agent unnamed Car (139820564322960) tried incompatible  action(s) (<scenic.domains.driving.actions.RegulatedControlAction object at 0x7f2a8315ad60>,)
```

Platform:

- Carla 0.9.14 docker
- Python 3.8.17 compiled from source
- Scenic 2.1.0b4

Here is the `loopy.py` script:

```
#!/usr/bin/env python3.8
import scenic

seconds = 20
timestep = .05
render = False

def simulate(scenic_file):
    scenario = scenic.scenarioFromFile(
                    scenic_file,
                    params={'timestep': timestep,
                            'render': render,
                            },
                    cacheImports=False
                    )
    scene, _ = scenario.generate(maxIterations=1)
    simulator = scenario.getSimulator()
    sim_result = simulator.simulate(
                        scene,
                        maxSteps=int(seconds // timestep),
                        maxIterations=1,
                        raiseGuardViolations=True
                        ) 
    return sim_result

res1 = simulate('tests/scenic2/dynamic1.scenic')
res2 = simulate('tests/scenic2/dynamic2.scenic')
```

and the two scenic files `dynamic1.scenic` and `dynamic2.scenic` specify the same scenario:

```
param carla_map = 'Town05'
carla_map = globalParameters.carla_map
param map = f'/home/carla/CarlaUE4/Content/Carla/Maps/OpenDrive/{carla_map}.xodr'
model scenic.simulators.carla.model

p0 = network.intersections[0].incomingLanes[0].centerline.pointAlongBy(0)

ego = Car at p0, facing roadDirection,
  with behavior FollowLaneBehavior(4)
```

abol-karimi:
[...]

PR #172, opened 2023-09-01
Fix multiple compilation of driving scenario bug; minor improvements

dfremont:
Fixes a variant of the issue in #123 which wasn't fixed by that PR and came up in #168. (Plus another couple minor things I improved along the way.)

This bug was enabled by the fact that it's still possible to import a Scenic module from a Python module if that Python module was imported during the compilation of a Scenic module. Completely banning imports of Scenic modules from Python seems to be nontrivial, so I'll save that for another PR.


Eric-Vin (approved):

Eric-Vin:
Restored the branch so that I can back-port it to 2.x.

PR #173, opened 2023-09-01
2.x Bug fix Back-ports

Eric-Vin:


Eric-Vin:
This passes tox on my machine, so we should be good.

tests/simulators/carla/test_carla.py
```diff
@@ -10,6 +10,27 @@ def test_basic(loadLocalScenario):
...
+    code = f"""
+        param map = r'{mapPath}'
+        param carla_map = 'Town01'
+        model scenic.simulators.carla.model
+        action = SetGearAction(0)
+        ego = Car
+        assert action.canBeTakenBy(ego)
+    """
+    for _ in range(2):
+        compileScenic(code, mode2D=True)
```
dfremont, line 31: Hmm, how did this work under 2.x? There shouldn't be any `mode2D` option. Did you try testing with CARLA installed?

dfremont (changes requested):
Looks good except for one small issue above -- but make sure to test with CARLA installed too.

tests/simulators/carla/test_carla.py
Eric-Vin: Good catch, fixed.

Eric-Vin:
Ah, good point on the Carla testing. I tested with that installed and fixed the relevant test, and we're now passing everything.

dfremont (approved):

PR #174, opened 2023-09-07
Overloaded Count to work for Scenes

Eric-Vin:


dfremont (changes requested):
Makes sense, although I would move `--count` out of the "Dynamic Simulations" section of the options (both in the `-h` list and in the documentation) since it is no longer specific to dynamic simulations.

Eric-Vin:
Good point, the changes have been made!

dfremont (approved):

PR #175, opened 2023-09-08
Pruning Improvements and Voxel Regions + Bugfix & Cleanup

Eric-Vin:
This PR adds various pruning improvements, including:

- New type of inradius for objects projected into the plane.
- Improvements to containment pruning, which allows for pruning efficiency equal to what was present in Scenic 2.0 for 2D scenes, even without 3D mode.
- Added a VoxelRegion class, and the ability to approximately erode a mesh. This enables pruning for objects contained in meshes, though not optimally. Better pruning comes at a higher compile time cost, enabled by increasing `PRUNING_PITCH` in `pruning.py`.
- Added a limited version of pruning based on visibility. This can be expanded in a future PR, but currently only prunes when an object must be visible from a fixed object.

**Performance Note**: This appears to add only minor compile time slowdowns to our current examples, so I think it's unlikely to cause a performance regression. Any performance loss would only be at compile time, and would be offset by greater sampling efficiency. In my tests, voxelization during pruning is the only added expensive operation and it takes roughly one second (per voxelization) with the current pitch.

### DIFF
--- a/src/scenic/core/object_types.py
+++ b/src/scenic/core/object_types.py
@@ -27,6 +27,7 @@ import shapely.affinity
 import trimesh
 
 from scenic.core.distributions import (
+    FunctionDistribution,
     MultiplexerDistribution,
     RandomControlFlowError,
     Samplable,
@@ -1255,67 +1256,68 @@ class Object(OrientedPoint):
     @cached_property
     def inradius(self):
         """A lower bound on the inradius of this object"""
-        # First check if all needed variables are defined. If so, we can
-        # compute the inradius exactly.
-        width, length, height, shape = self.width, self.length, self.height, self.shape
-        if not any(needsSampling(val) for val in (width, length, height, shape)):
-            shapeRegion = MeshVolumeRegion(
+
+        # Define a helper function that computes the actual inradius
+        def inradiusActual(width, length, height, shape):
+            return MeshVolumeRegion(
                 mesh=shape.mesh, dimensions=(width, length, height)
-            )
-            return shapeRegion.inradius
+            ).inradius
 
-        # If we have a uniform distribution over shapes and a supportInterval for each dimension,
-        # we can compute a supportInterval for this object's inradius
+        # Define a helper function that computes the support of the inradius,
+        # given the sub supports.
+        def inradiusSupport(width_s, length_s, height_s, shape_s):
+            # Unpack the dimension supports (and ignore the shape support)
+            min_width, max_width = width_s
+            min_length, max_length = length_s
+            min_height, max_height = height_s
 
-        # Define helper class
-        class InradiusHelper:
-            def __init__(self, support):
-                self.support = support
+            if None in [
+                min_width,
+                max_width,
+                min_length,
+                max_length,
+                min_height,
+                max_height,
+            ]:
+                # Can't get a bound on one or more dimensions, abort
+                return None, None
 
-            def supportInterval(self):
-                return self.support
+            min_bounds = np.array([min_width, min_length, min_height])
+            max_bounds = np.array([max_width, max_length, max_height])
 
-        # Extract bounds on all dimensions
-        min_width, max_width = supportInterval(width)
-        min_length, max_length = supportInterval(length)
-        min_height, max_height = supportInterval(height)
+            # Extract a list of possible shapes
+            if isinstance(self.shape, Shape):
+                shapes = [self.shape]
+            elif isinstance(self.shape, MultiplexerDistribution) and all(
+                isinstance(opt, Shape) for opt in self.shape.options
+            ):
+                shapes = self.shape.options
+            else:
+                # Something we don't recognize, abort
+                return None, None
 
-        if None in [min_width, max_width, min_length, max_length, min_height, max_height]:
-            # Can't get a bound on one or more dimensions, abort
-            return 0
+            # Get the inradius for each shape with the min and max bounds
+            min_distances = [
+                MeshVolumeRegion(mesh=shape.mesh, dimensions=min_bounds).inradius
+                for shape in shapes
+            ]
+            max_distances = [
+                MeshVolumeRegion(mesh=shape.mesh, dimensions=max_bounds).inradius
+                for shape in shapes
+            ]
 
-        min_bounds = np.array([min_width, min_length, min_height])
-        max_bounds = np.array([max_width, max_length, max_height])
+            distance_range = (min(min_distances), max(max_distances))
 
-        # Extract a list of possible shapes
-        if isinstance(shape, Shape):
-            shapes = [shape]
-        elif isinstance(shape, MultiplexerDistribution) and all(
-            isinstance(opt, Shape) for opt in shape.options
-        ):
-            shapes = shape.options
+            return distance_range
+
+        # Return either the inradius or a FunctionDistribution using the above helpers
+        args = toDistribution((self.width, self.length, self.height, self.shape))
+        if not isLazy(args):
+            return inradiusActual(*args)
         else:
-            # Something we don't recognize, abort
-            return 0
-
-        # Check that all possible shapes contain the origin
-        if not all(shape.containsCenter for shape in shapes):
-            # One or more shapes has inradius 0
-            return 0
-
-        # Get the inradius for each shape with the min and max bounds
-        min_distances = [
-            MeshVolumeRegion(mesh=shape.mesh, dimensions=min_bounds).inradius
-            for shape in shapes
-        ]
-        max_distances = [
-            MeshVolumeRegion(mesh=shape.mesh, dimensions=max_bounds).inradius
-            for shape in shapes
-        ]
-
-        distance_range = (min(min_distances), max(max_distances))
-
-        return InradiusHelper(support=distance_range)
+            return FunctionDistribution(
+                args=args, kwargs={}, func=inradiusActual, support=inradiusSupport
+            )
 
     @cached_property
     def planarInradius(self):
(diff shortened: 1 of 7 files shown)


Eric-Vin:
Ready to review, but need to do some bench-marking before we merge in.

src/scenic/core/object_types.py
```diff
@@ -1255,67 +1256,138 @@ def boundingBox(self):
...
-        min_length, max_length = supportInterval(length)
-        min_height, max_height = supportInterval(height)
+        # Return either the inradius or a FunctionDistribution using the above helpers
+        args = toDistribution((self.width, self.length, self.height, self.shape))
+        if not isLazy(args):
+            return inradiusActual(*args)
+        else:
+            return FunctionDistribution(
+                args=args, kwargs={}, func=inradiusActual, support=inradiusSupport
+            )
```
dfremont, line 1320: Rather than explicitly calling `toDistribution`, `isLazy`, and `FunctionDistribution` here, couldn't you just stick `@distributionFunction(support=inradiusSupport)` on `inradiusActual` and then `return inradiusActual(...)`?

src/scenic/core/object_types.py
```diff
@@ -1255,67 +1256,138 @@ def boundingBox(self):
...
+        args = toDistribution((self.width, self.length, self.shape))
+        if not isLazy(args):
+            return planarInradiusActual(*args)
+        else:
+            return FunctionDistribution(
+                args=args,
+                kwargs={},
+                func=planarInradiusActual,
+                support=planarInradiusSupport,
+            )
```
dfremont, line 1390: Same comment as above. Also maybe not for this PR, but it would be nice to consolidate the logic here, since it's quite similar between these two values. Maybe we can make a decorator or helper function for properties which are monotonic in the object dimensions.

src/scenic/core/pruning.py
```diff
@@ -135,8 +146,6 @@ def matchPolygonalField(heading, position):
 
 
 ### Pruning procedures
-
-
```
dfremont, line 139: I would like to leave at least 1 blank line after these headings, since they aren't comments referring to a single function but blocking out a whole section of the file.

src/scenic/core/pruning.py
```diff
@@ -428,3 +579,60 @@ def relativeHeadingRange(
...
+        for child in deps:
+            if dfs(child):
+                return True
+
+        # Set to completed
+        state[target] = 2
+
+        return False
+
+    return dfs(scenario)
```
dfremont, line 634: I don't think this will work: `dependencies(scenario)` will be empty, since `Scenario` is not a subclass of `LazilyEvaluable`. Why not do DFS from `B` and see if you ever reach `A`?

I see you have a test that's supposed to involve this check, so either I'm confused or that test doesn't actually require this code to work correctly.

src/scenic/core/regions.py
```diff
@@ -1417,6 +1418,7 @@ def intersect(self, other, triedReversed=False):
             # Other region is a mesh volume. We can extract the mesh to perform boolean operations on it
             other_mesh = other.mesh
 
+            # If dimensions are prov
```
dfremont, line 1421: Incomplete comment

src/scenic/core/regions.py
```diff
@@ -2055,6 +2062,152 @@ def evaluateInner(self, context):
...
 
+class VoxelRegion(Region):
+    """Region represented by a voxel grid in 3D space.
+
+    Args:
+        encoding: A numpy array encoding the voxel grid.
+        dimensions: An optional 3-tuple, with the values representing width, length, height
+          respectively. The voxel region will be scaled to have these dimensions. Default
+          value (1,1,1).
+        position: A position, which determines where the center of the region will be. Default
```
dfremont, line 2073: These arguments don't appear in the signature: it seems `voxelGrid` must be used.

src/scenic/core/regions.py
```diff
@@ -2055,6 +2062,152 @@ def evaluateInner(self, context):
...
+        name: An optional name to help with debugging.
+        lazy: Whether or not to be lazy about pre-computing internal values. Set this to True if this
+          VoxelRegion is unlikely to be used outside of an intermediate step in compiling/pruning.
+    """
+
+    def __init__(self, voxelGrid, orientation=None, name=None, lazy=False):
+        # Initialize superclass
+        super().__init__(name, orientation=orientation)
+
+        # Work around Trimesh caching bug
```
dfremont, line 2088: Would be good to put a link to the corresponding trimesh issue to help us remember to remove this when the bug is fixed.

src/scenic/core/regions.py
```diff
@@ -2055,6 +2062,152 @@ def evaluateInner(self, context):
...
+        if self._voxelGrid.encoding.is_empty:
+            raise ValueError("Tried to create an empty VoxelRegion.")
+
+        # Transform voxel grid points and extract scale
+        self.voxel_points = self._voxelGrid.points
+        self.scale = self._voxelGrid.scale
+
+        # Initialize KD-Tree for containment checking if not lazy
+        if not lazy:
+            self.kdTree
```
dfremont, line 2103: What's the rationale for forcing the computation here rather than just having it be computed the first time it's needed? (In other words, why provide `lazy` as an option rather than just having it always be true?)

src/scenic/core/regions.py
```diff
@@ -2055,6 +2062,152 @@ def evaluateInner(self, context):
...
+
+    @cached_property
```
dfremont, line 2105:
```suggestion
```

src/scenic/core/regions.py
```diff
@@ -2055,6 +2062,152 @@ def evaluateInner(self, context):
...
+        self._voxelGrid = trimesh.voxel.VoxelGrid(
+            voxelGrid.encoding, transform=voxelGrid.transform.copy()
+        )
+
+        # Check that the encoding isn't empty. In that case, raise an error.
+        if self._voxelGrid.encoding.is_empty:
+            raise ValueError("Tried to create an empty VoxelRegion.")
+
+        # Transform voxel grid points and extract scale
+        self.voxel_points = self._voxelGrid.points
```
dfremont, line 2098: Is some transformation actually supposed to happen here?

src/scenic/core/regions.py
```diff
@@ -2055,6 +2062,152 @@ def evaluateInner(self, context):
...
+        _, index = self.kdTree.query([point])
+        closest_point = self.voxel_points[index]
+
+        # Check voxel containment
+        voxel_low = closest_point - self.scale / 2
+        voxel_high = closest_point + self.scale / 2
+
+        return numpy.all(voxel_low <= point, axis=1) & numpy.all(
+            point <= voxel_high, axis=1
+        )
```
dfremont, line 2126: This returns an array containing a bool, not just a bool (better extend the test to check for that!). I think you can replace `[point]` with just `point` when calling `self.kdTree.query` and then get rid of the `axis` stuff in the last line.

src/scenic/core/regions.py
```diff
@@ -2055,6 +2062,152 @@ def evaluateInner(self, context):
...
+
+    def uniformPointInner(self):
+        # First generate a point uniformly in a box with dimensions
+        # equal to scale, centered at the origin.
+        base_pt = numpy.random.random_sample(3) - 0.5
+        scaled_pt = base_pt * self.scale
+        voxel_base = self.voxel_points[random.randrange(len(self.voxel_points))]
+        offset_pt = voxel_base + scaled_pt
+
+        # Then pick a random voxel point and add the base point to that point.
```
dfremont, line 2148: Comment in wrong spot (wording is confusing too)

src/scenic/core/scenarios.py
```diff
@@ -469,6 +469,7 @@ def _generateInner(self, maxIterations, verbosity, feedback):
             numpy.random.set_state(np_state)
 
             if rejection is not None:
+                # breakpoint()
```
dfremont, line 472:
```suggestion
```

tests/core/test_regions.py
```diff
@@ -521,6 +521,77 @@ def test_pointset_region():
...
+    vr2 = VoxelRegion(vg2)
+
+    assert vr2.size == pytest.approx((7 / 27) * 5 * 5 * 3)
+    assert vr2.dimensionality == 3
+
+
+def test_mesh_voxelization(getAssetPath):
+    plane_region = MeshVolumeRegion.fromFile(getAssetPath("meshes/classic_plane.obj.bz2"))
+    voxel_grid = plane_region.mesh.voxelized(max(plane_region.mesh.extents) / 100).fill()
+    vr = VoxelRegion(voxelGrid=voxel_grid)
```
dfremont, line 578: Maybe it would be better to test the `MeshVolumeRegion.voxelized` method here rather than using the raw `trimesh` methods.

tests/syntax/test_properties.py
```diff
@@ -128,3 +129,157 @@ def test_backRight():
...
+        hollow_mesh = trimesh.creation.box((1,1,1)).difference(
+            trimesh.creation.box((0.5,0.5,0.5)))
+        ego = new Object with width Range(1, 3),
+            with length Range(1, 3), with height Range(1, 3),
+            facing (Range(0, 360) deg, Range(0, 360) deg, Range(0, 360) deg),
+            with shape MeshShape(hollow_mesh)
+        """
+    )
+    ego = sampleEgo(scenario)
+    assert isinstance(scenario.objects[0].inradius, Distribution)
```
dfremont, line 189: This assertion seems fragile: future improvements might lead to obtaining the constant zero here (from the support interval, we can see that Scenic already "knows" the value has to be zero).

tests/syntax/test_pruning.py
```diff
@@ -21,6 +22,57 @@ def test_containment_in():
...
+    # Test both combined, in a slightly more complicated case.
+    # Specifically, there is a non vertical component to baseOffset
+    # that should be accounted for and the height is random.
+    scenario = compileScenic(
+        """
+        class TestObject:
+            baseOffset: (0.1, 0, self.height/2)
+
+        workspace = Workspace(PolygonalRegion([0@0, 2@0, 2@2, 0@2]))
+        ego = new Object in workspace, with height Range(0.1,0.5)
```
dfremont, line 65:
```suggestion
        ego = new TestObject on workspace, with height Range(0.1,0.5)
```

tests/syntax/test_pruning.py
```diff
@@ -21,6 +22,57 @@ def test_containment_in():
...
+    """
+    )
+    # Sampling should fail ~30.56% of the time, so
```
dfremont, line 68: Could you explain why this is the case? Since the non-vertical component of the offset is constant I would have thought pruning could take it into account exactly.

tests/syntax/test_pruning.py
```diff
@@ -106,3 +158,63 @@ def test_relative_heading_inconsistent():
...
+
+def test_visibility_pruning():
+    """Test visibility pruning in general.
+
+    The following scenarios are equivalent except for how they specify that foo
+    must be visible from ego. The size of the workspace and the visibleDistance
+    of ego are chosen such that without pruning the chance of sampling a valid
+    scene over 100 tries is 1-(1-Decimal(3.14)/Decimal(10000000000**2))**100 = ~1e-18.
+    Assuming the approximately buffered volume of the viewRegion has a 50% chance of
+    ejecting (i.e. it is twice as large as the true buffered viewRegion, which testing
```
dfremont, line 171:
```suggestion
    rejecting (i.e. it is twice as large as the true buffered viewRegion, which testing
```

tests/syntax/test_pruning.py
```diff
@@ -106,3 +158,63 @@ def test_relative_heading_inconsistent():
...
         )
+
+
+def test_visibility_pruning():
+    """Test visibility pruning in general.
+
+    The following scenarios are equivalent except for how they specify that foo
+    must be visible from ego. The size of the workspace and the visibleDistance
+    of ego are chosen such that without pruning the chance of sampling a valid
+    scene over 100 tries is 1-(1-Decimal(3.14)/Decimal(10000000000**2))**100 = ~1e-18.
```
dfremont, line 169:
```suggestion
    scene over 100 tries is 1-(1-Decimal(3.14)/Decimal(1e10**2))**100 = ~1e-18.
```

tests/syntax/test_pruning.py
```diff
@@ -106,3 +158,63 @@ def test_relative_heading_inconsistent():
...
+    in the viewRegion instead of at any point where the object intersects the view region.
+    Because of this, we want to see at least one sample where the position is outside
+    the viewRegion but the object intersects the viewRegion. The chance of this happening
+    is 1-((4/3)*math.pi*(1)**3)/((4/3)*math.pi*(1.1)**3) = ~25%, so by repeating the process
+    30 times we have a 1e-19 chance of not gettin a single point in this zone.
+    """
+    # requireVisible
+    scenario = compileScenic(
+        """
+        workspace = Workspace(RectangularRegion(0@0, 0, 10000000000, 10000000000))
```
dfremont, line 185:
```suggestion
        workspace = Workspace(RectangularRegion(0@0, 0, 1e10, 1e10))
```

tests/syntax/test_pruning.py
```diff
@@ -106,3 +158,63 @@ def test_relative_heading_inconsistent():
...
+    """
+    )
+    positions = [sampleParamP(scenario, maxIterations=100) for i in range(30)]
+    assert all(pos.distanceTo(Vector(0, 0, 0)) <= 1.1 for pos in positions)
+    assert any(pos.distanceTo(Vector(0, 0, 0)) >= 1 for pos in positions)
+
+    # visible
+    scenario = compileScenic(
+        """
+        workspace = Workspace(RectangularRegion(0@0, 0, 10000000000, 10000000000))
```
dfremont, line 199:
```suggestion
        workspace = Workspace(RectangularRegion(0@0, 0, 1e10, 1e10))
```

tests/syntax/test_pruning.py
```diff
@@ -106,3 +158,63 @@ def test_relative_heading_inconsistent():
...
+    Assuming the approximately buffered volume of the viewRegion has a 50% chance of
+    ejecting (i.e. it is twice as large as the true buffered viewRegion, which testing
+    indicates in this case has about a 10% increase in volume for this case), the chance
+    of not finding a sample in 100 iterations is 1e-31.
+
+    We also want to confirm that we aren't pruning too much, i.e. placing the position
+    in the viewRegion instead of at any point where the object intersects the view region.
+    Because of this, we want to see at least one sample where the position is outside
+    the viewRegion but the object intersects the viewRegion. The chance of this happening
+    is 1-((4/3)*math.pi*(1)**3)/((4/3)*math.pi*(1.1)**3) = ~25%, so by repeating the process
```
dfremont, line 179:
```suggestion
    per sample is 1 - (1 / 1.1)**3 = ~25%, so by repeating the process
```

tests/syntax/test_pruning.py
```diff
@@ -106,3 +158,63 @@ def test_relative_heading_inconsistent():
...
+    30 times we have a 1e-19 chance of not gettin a single point in this zone.
```
dfremont, line 180:
```suggestion
    30 times we have a 1e-19 chance of not getting a single point in this zone.
```

dfremont (changes requested):
Looks great overall. All my comments are minor except I think something's wrong with the cyclic dependency check in visibility pruning (see above).

src/scenic/core/object_types.py
Eric-Vin: Good idea! Much cleaner this way.

src/scenic/core/pruning.py
Eric-Vin: Ah good point, I'll re-add the spaces.

src/scenic/core/regions.py
Eric-Vin: Not sure what this was but it doesn't seem necessary. Removing it.

src/scenic/core/regions.py
Eric-Vin: Ah good catch, the docstring is out of date. I'll update it.

src/scenic/core/regions.py
Eric-Vin: Actually the issue has since been resolved but is only available in Trimesh >= 4.0. I'll remove the workaround since we're requiring that in this PR: https://github.com/BerkeleyLearnVerify/Scenic/pull/169

src/scenic/core/regions.py
Eric-Vin: This is to force computation at compile time instead of sample time (unless we expect this object not to need the kdTree, i.e. `lazy=True`).

src/scenic/core/regions.py
Eric-Vin: Good catch, this was originally a computed value but now is not.

src/scenic/core/regions.py
Eric-Vin: Whoops, a relic of old code. I'll clean up.

src/scenic/core/regions.py
Eric-Vin: Actually, I'll leave in the workaround and remove it once we require a high enough Trimesh version (to avoid CI failures)

src/scenic/core/regions.py
Eric-Vin: Good catch!

tests/core/test_regions.py
Eric-Vin: Good idea!

tests/syntax/test_properties.py
Eric-Vin: Good point, I'll remove this assertion and leave the one above since that really must be a distribution.

src/scenic/core/regions.py
Eric-Vin: Removing as discussed offline.

Eric-Vin:
Alright, should have all the changes you requested implemented!

dfremont (approved):

Issue #177, opened 2023-09-13
TypeError: a bytes-like object is required, not 'list'


Nancy22-NJ:
Hello, i have problem on encoding the scene as bytes. 
The following are the codes and its error message. 
Looking forward to your reply!
**data = scenario.sceneToBytes(scene)**
**TypeError: a bytes-like object is required, not 'list'**

Eric-Vin:
Hello,

Could you provide a code snippet that triggers this error? I'm able to serialize scenes on my machine so I'm unsure where the problem might be. Once I can reproduce the issue I'm happy to help further!

wlxxi:
**scenario = scenic.scenarioFromFile(task, params=dict, mode2D=True)
                    scene, _ = scenario.generate()
                    
                    data = scenario.sceneToBytes(scene)
                    with open(pathlib.Path(tempfile.gettempdir()) / 'record/test.scene', 'wb') as f:
                        f.write(data)**
I was trying to store this scene.

Eric-Vin:
I'm able to serialize one of our driving examples with this script, so I'm not sure what the problem is. Can you print out what scene is?

wlxxi:
> I'm able to serialize one of our driving examples with this script, so I'm not sure what the problem is. Can you print out what scene is?
`
param map = localPath('../assets/maps/CARLA/Town04.xodr')
param carla_map = 'Town04'
param weather = 'WetCloudyNoon'
model scenic.simulators.carla.model

## CONSTANTS
# Ego Car
EGO_MODEL = "vehicle.lincoln.mkz_2017"
EGO_SPEED = 60
EGO_BRAKING_THRESHOLD = 10

# Background Car
BC_MODEL = "vehicle.tesla.cybertruck"
param BC_V = [20, 40, 40, 20]
param BC_DIST = 40
BC_BRAKING_THRESHOLD = 10

BRAKE_ACTION = 1.0

## DEFINING BEHAVIORS
# EGO BEHAVIOR: Follow lane, and brake after passing a threshold distance to the leading car, terminate when collision take place
behavior EgoBehavior():
    try:
        do FollowLaneBehavior(EGO_SPEED)

    interrupt when withinDistanceToAnyCars(self, EGO_BRAKING_THRESHOLD):
         take SetBrakeAction(BRAKE_ACTION)


# BV BEHAVIOR: fix route, vary velocity
behavior BackgroundCarBehavior():

[...]

PR #178, opened 2023-09-13
Serialization Fixes for 3D

Eric-Vin:
It looks like some behavior isn't accounted for by our serialization library, specifically orientations from vector fields. I've added a test that reproduces that behavior and serialization functionality for `Orientation` objects.

### DIFF
--- a/tests/core/test_serialization.py
+++ b/tests/core/test_serialization.py
@@ -38,6 +38,12 @@ simpleScenario = """
     param baz = ({'a', 'b', 'c', 'd'}, frozenset({'e', 'f', 'g', 'h'}))
 """
 
+randomScenario = """
+    box = BoxRegion(dimensions=(5,5,5))
+    new Object in box, facing (Range(0,360) deg, Range(0,360) deg, Range(0,360) deg)
+    new Object on box.getSurfaceRegion()
+"""
+
 
 def skeleton(scene):
     # Simple picklable representation of the main parameters of the scene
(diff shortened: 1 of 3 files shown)


tests/core/test_serialization.py
```diff
@@ -38,6 +38,12 @@
     param baz = ({'a', 'b', 'c', 'd'}, frozenset({'e', 'f', 'g', 'h'}))
 """
 
+randomScenario = """
```
dfremont, line 41: If this scenario is only going to be used in one test I would move it into that test. If you want to keep it here I would at least give it a more informative name, since the `simpleScenario` above also has plenty of randomness in it.

tests/core/test_serialization.py
```diff
@@ -202,6 +208,14 @@ def test_simple_scene(self):
         assert scenario.sceneToBytes(scene2) == data
         assertSceneEquivalence(scene1, scene2)
 
+    def test_random_scene(self):
```
dfremont, line 211: Likewise, `test_random_scene` is misleading; I'd use `test_scene_random_orientation` or something like that.

dfremont (changes requested):
Looks good overall. Some minor comments above; also please add a line for `Orientation` objects to `test_object_with_encodeTo`.

Eric-Vin:
Tests have been moved/added.

Eric-Vin:
Should address Issue https://github.com/BerkeleyLearnVerify/Scenic/issues/177

dfremont:
Merging since the remaining test failure is unrelated.

dfremont (approved):

PR #182, opened 2023-09-24
Avoid copying objects each time step

dfremont:
We now use a single copy of each object throughout a dynamic simulation, mutating its dynamic properties, recomputing dynamic final properties, and explicitly clearing the caches for cached properties/methods (for those using Scenic's own internal decorators, of course: people using `functools.cached` etc. probably don't expect that to get cleared every step). As a result, we no longer need to run specifier resolution at each time step (previously necessary to ensure dynamic final properties were updated), which turned out to cause major slowdowns when an object had a property storing a large nested list. For such cases the major performance regression from Scenic 2 to 3 should be substantially reduced.

(I've removed my previous comment about the remaining overhead when recomputing dynamic finals: the latest version of this PR gets rid of all the extraneous work we can eliminate without special-casing known properties like `orientation` to remove type checks, etc.)

### DIFF
--- a/src/scenic/core/object_types.py
+++ b/src/scenic/core/object_types.py
@@ -45,7 +45,12 @@ from scenic.core.geometry import (
     normalizeAngle,
     pointIsInCone,
 )
-from scenic.core.lazy_eval import LazilyEvaluable, isLazy, needsLazyEvaluation
+from scenic.core.lazy_eval import (
+    LazilyEvaluable,
+    isLazy,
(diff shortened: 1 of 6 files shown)


dfremont:
@abol-karimi could you please check if this branch (`dynamic-props`) fixes the major slowdown you were seeing?

Eric-Vin (approved):
These changes look good to me, but I agree we should address the way types are inferred in the near future.

PR #183, opened 2023-09-25
Add Intersects Operator

Eric-Vin:
Add an operator for checking intersection of Objects/Regions. This will automatically use internal heuristics for fast checks as well.

### DIFF
--- a/docs/glossary.rst
+++ b/docs/glossary.rst
@@ -47,7 +47,7 @@ Glossary
 	footprint
 		The infinite extrusion of a 2D `region` in the positive and negative Z directions.
 		Testing containment of an `object` in a 2D region automatically uses its footprint, so that the object is considered contained if and only if its projection into the plane of the region is contained in the region.
-		Footprints are represented internally by instances of the `PolygonalFootprintRegion` class.
+		Footprints are represented internally by instances of the `PolygonalFootprintRegion` class, and can be accessed using the `fooprint` attribute.
 
 	global parameters
 		Parameters of a scene like weather or time of day which are not associated with any object.
--- a/docs/glossary.rst
+++ b/docs/glossary.rst
@@ -47,7 +47,7 @@ Glossary
 	footprint
 		The infinite extrusion of a 2D `region` in the positive and negative Z directions.
 		Testing containment of an `object` in a 2D region automatically uses its footprint, so that the object is considered contained if and only if its projection into the plane of the region is contained in the region.
-		Footprints are represented internally by instances of the `PolygonalFootprintRegion` class, and can be accessed using the `fooprint` attribute.
+		Footprints are represented internally by instances of the `PolygonalFootprintRegion` class, and can be accessed using the ``fooprint`` attribute.
 
 	global parameters
 		Parameters of a scene like weather or time of day which are not associated with any object.
--- a/docs/glossary.rst
+++ b/docs/glossary.rst
@@ -47,7 +47,7 @@ Glossary
 	footprint
 		The infinite extrusion of a 2D `region` in the positive and negative Z directions.
 		Testing containment of an `object` in a 2D region automatically uses its footprint, so that the object is considered contained if and only if its projection into the plane of the region is contained in the region.
-		Footprints are represented internally by instances of the `PolygonalFootprintRegion` class, and can be accessed using the ``fooprint`` attribute.
+		Footprints are represented internally by instances of the `PolygonalFootprintRegion` class, and can be accessed using the ``footprint`` attribute.
 
 	global parameters
 		Parameters of a scene like weather or time of day which are not associated with any object.
--- a/docs/reference/operators.rst
+++ b/docs/reference/operators.rst
@@ -71,7 +71,7 @@ Whether a position or `Object` lies in the `Region`; for the latter, the object 
 
 (*Object* | *region*) intersects (*Object* | *region*)
 ------------------------------------------------------
-Whether an `Object`/`Region` intersects another `Object`/`Region`, i.e. whether any portion of the occupied spaces intersect.
+Whether an `Object`/`Region` intersects another `Object`/`Region`, i.e. whether any portion of the occupied spaces intersect. It can often be useful to check intersection with the :term:`footprint` of a region, e.g. when checking whether a car intersects a given lane. In this case, one would write :scenic:`car intersects lane.footprint` instead of :scenic:`car intersects lane`. For more details, see :term:`footprint`.
 
 
 Orientation Operators
--- a/docs/reference/operators.rst
+++ b/docs/reference/operators.rst
@@ -71,7 +71,9 @@ Whether a position or `Object` lies in the `Region`; for the latter, the object 
 
 (*Object* | *region*) intersects (*Object* | *region*)
 ------------------------------------------------------
-Whether an `Object`/`Region` intersects another `Object`/`Region`, i.e. whether any portion of the occupied spaces intersect. It can often be useful to check intersection with the :term:`footprint` of a region, e.g. when checking whether a car intersects a given lane. In this case, one would write :scenic:`car intersects lane.footprint` instead of :scenic:`car intersects lane`. For more details, see :term:`footprint`.
+Whether an `Object`/`Region` intersects another `Object`/`Region`, i.e. whether any portion of the occupied spaces intersect.
+
+When working with 2D regions, it can be useful to check intersection with the :term:`footprint` of a region, e.g. when checking whether a car intersects a given lane. In this case, one would write :scenic:`car intersects lane.footprint` instead of :scenic:`car intersects lane`. For more details, see :term:`footprint`.
 
 
 Orientation Operators
--- a/docs/reference/operators.rst
+++ b/docs/reference/operators.rst
@@ -67,6 +67,12 @@ See the :ref:`Visibility System <visibility>` reference for a discussion of the 
 ----------------------------------
 Whether a position or `Object` lies in the `Region`; for the latter, the object must be completely contained in the region.
 
+.. _({Object} | {region}) intersects ({Object} | {region}):
+
+(*Object* | *region*) intersects (*Object* | *region*)
+------------------------------------------------------
+Whether an `Object`/`Region` intersects another `Object`/`Region`, i.e. whether any portion of the occupied spaces intersect.
+
 
 Orientation Operators
 =====================
--- a/docs/syntax_guide.rst
+++ b/docs/syntax_guide.rst
@@ -282,9 +282,11 @@ In the following tables, operators are grouped by the type of value they return.
    * - Boolean Operators
      - Meaning
    * - :sampref:`({Point} | {OrientedPoint}) can see ({vector} | {Object})`
-     - Whether or not a position or `Object` is visible from a `Point` or `OrientedPoint`.
+     - Whether or not a position or `Object` is visible from a `Point` or `OrientedPoint`
    * - :sampref:`({vector} | {Object}) in {region}`
-     -  Whether a position or `Object` lies in the region
+     - Whether a position or `Object` lies in the region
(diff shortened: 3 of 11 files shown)


docs/syntax_guide.rst
```diff
@@ -282,9 +282,11 @@ In the following tables, operators are grouped by the type of value they return.
    * - Boolean Operators
      - Meaning
    * - :sampref:`({Point} | {OrientedPoint}) can see ({vector} | {Object})`
-     - Whether or not a position or `Object` is visible from a `Point` or `OrientedPoint`.
+     - Whether or not a position or `Object` is visible from a `Point` or `OrientedPoint`
    * - :sampref:`({vector} | {Object}) in {region}`
-     -  Whether a position or `Object` lies in the region
+     - Whether a position or `Object` lies in the region
+   * - :sampref:`({Object} | {region}) intersects ({Object} | {region})`
+     - Whether an `Object`/`Region` intersects with an `Object`/`Region`.
```
dfremont, line 289:
```suggestion
     - Whether an `Object`/`Region` intersects an `Object`/`Region`.
```

src/scenic/core/object_types.py
```diff
@@ -1066,22 +1066,41 @@ def distanceTo(self, point):
...
 
             self_poly = self._boundingPolygon
             other_poly = other._boundingPolygon
             return self_poly.intersects(other_poly)
 
-        if isLazy(self.occupiedSpace) or isLazy(other.occupiedSpace):
+        # For an object that is a box and flat with a polygonal region, we can
+        # also take a fast route.
+        if self._isPlanarBox and (isinstance(other, PolygonalRegion) and other.z == 0):
+            return PolygonalRegion(polygon=self._boundingPolygon).intersects(other)
```
dfremont, line 1089: Creating a `PolygonalRegion` seems unnecessary here. Can't we just do `self._boundingPolygon.intersects(other.polygons)` (after checking for laziness)?

src/scenic/syntax/ast.py
```diff
@@ -1250,3 +1250,13 @@ def __init__(self, left: ast.AST, right: ast.AST, *args: any, **kwargs: any) ->
         self.left = left
         self.right = right
         self._fields = ["left", "right"]
+
+
+class IntersectsOp(AST):
+    __match_args__ = ("left", "right")
+
+    def __init__(self, left: ast.AST, right: ast.AST, *args: any, **kwargs: any) -> None:
```
dfremont, line 1258:
```suggestion
    def __init__(self, left: ast.AST, right: ast.AST, *args: Any, **kwargs: Any) -> None:
```

(I see this file uses `any` all over the place, but that's wrong: `any` is a built-in function, whereas we want `Any` from the `typing` module. We can make another PR to fix those, or you can find-and-replace them here.)

tests/syntax/test_operators.py
```diff
@@ -503,6 +503,56 @@ def test_object_in_region_3d():
     assert p == (True, True, False, False)
 
 
+# Intersects
+def test_intersects_obj_obj():
+    p = sampleParamPFrom(
+        """
+    obj1 = new Object at (-1,0,0.1), with allowCollisions True
```
dfremont, line 510: The Scenic code should be indented.

tests/syntax/test_operators.py
```diff
@@ -503,6 +503,56 @@ def test_object_in_region_3d():
...
+    obj2 = new Object at (1,0,0), with allowCollisions True
+    obj3 = new Object with width 10, with length 10, with height 10, with allowCollisions True
```
dfremont, line 512: Could you please add a case where neither the center nor the corners of the objects lie in each other, but they still intersect?

src/scenic/core/object_types.py
```diff
@@ -1066,22 +1066,41 @@ def distanceTo(self, point):
...
                 return False
 
             self_poly = self._boundingPolygon
             other_poly = other._boundingPolygon
             return self_poly.intersects(other_poly)
 
-        if isLazy(self.occupiedSpace) or isLazy(other.occupiedSpace):
+        # For an object that is a box and flat with a polygonal region, we can
+        # also take a fast route.
+        if self._isPlanarBox and (isinstance(other, PolygonalRegion) and other.z == 0):
```
dfremont, line 1088: Shouldn't the `other.z == 0` be something like `abs(self.position.z - other.z) <= self.height/2`?

tests/syntax/test_operators.py
```diff
@@ -503,6 +503,56 @@ def test_object_in_region_3d():
...
+    )
+    assert p == (False, False, True, True)
+
+
+def test_intersects_fast_paths():
+    p = sampleParamPFrom(
+        """
+    obj1 = new Object at (0.25,0,0), with allowCollisions True
+    obj2 = new Object at (-0.25,0,0), with allowCollisions True
+    reg = RectangularRegion((0,0,0), 0, 10, 10)
```
dfremont, line 549: Should add cases where the objects are not at `z=0`, and cases where the object intersects the region but is not contained in it, plus where the object doesn't intersect the region at all (both for being too far out of the correct plane and for being in the right plane but not intersecting the polygon).

dfremont (changes requested):
I'm not sure the object-polygon fast path is correct -- see below. Other than that just a few minor comments.

src/scenic/core/object_types.py
Eric-Vin: Hmm, actually I don't think we need to check z at all here, since 2D regions are infinite over the z axis. Does that sound right?

Eric-Vin: Actually, those are the footprints. Should intersection use footprint?

src/scenic/syntax/ast.py
Eric-Vin: Do we even need typing information here? If the type can be any then I feel like it's cleaner to just leave it blank and let any analysis automatically assume it can be any type.

src/scenic/core/object_types.py
Eric-Vin: As we discussed, fixed to your solution for now.  For reference, we plan to discuss adding an operator that assumes footprint.

src/scenic/core/object_types.py
Eric-Vin: Good point, making that change.

Eric-Vin:
Made the relevant changes, but converting this to a draft PR while we discuss a similar operator that uses the footprint of regions as discussed.

src/scenic/syntax/ast.py
dfremont: We don't use the typing information. I assume @shumbo put it in in case we later run a type checker on the implementation. I think the best practice is to put `Any` for arguments which are known to really be able to take on any type (or at least a sufficiently broad range of types that there's no easy way to specify them more precisely) rather than leaving the arguments unannotated: the latter can be mistaken for "we just haven't gotten around to annotating this yet".

Eric-Vin: Sounds good, I've put it on the todo list!

docs/glossary.rst
```diff
@@ -47,7 +47,7 @@ Glossary
 	footprint
 		The infinite extrusion of a 2D `region` in the positive and negative Z directions.
 		Testing containment of an `object` in a 2D region automatically uses its footprint, so that the object is considered contained if and only if its projection into the plane of the region is contained in the region.
-		Footprints are represented internally by instances of the `PolygonalFootprintRegion` class.
+		Footprints are represented internally by instances of the `PolygonalFootprintRegion` class, and can be accessed using the ``fooprint`` attribute.
```
dfremont, line 50:
```suggestion
		Footprints are represented internally by instances of the `PolygonalFootprintRegion` class, and can be accessed using the ``footprint`` attribute.
```

docs/reference/operators.rst
```diff
@@ -67,6 +67,12 @@ See the :ref:`Visibility System <visibility>` reference for a discussion of the
 ----------------------------------
 Whether a position or `Object` lies in the `Region`; for the latter, the object must be completely contained in the region.
 
+.. _({Object} | {region}) intersects ({Object} | {region}):
+
+(*Object* | *region*) intersects (*Object* | *region*)
+------------------------------------------------------
+Whether an `Object`/`Region` intersects another `Object`/`Region`, i.e. whether any portion of the occupied spaces intersect. It can often be useful to check intersection with the :term:`footprint` of a region, e.g. when checking whether a car intersects a given lane. In this case, one would write :scenic:`car intersects lane.footprint` instead of :scenic:`car intersects lane`. For more details, see :term:`footprint`.
```
dfremont, line 74: Let's put the footprint discussion in its own paragraph, starting with something like "When working with 2D regions, it can be useful...". That way people who aren't using 2D regions don't have to read that paragraph.

docs/glossary.rst
Eric-Vin: Good catch, thanks!

src/scenic/core/object_types.py
```diff
@@ -1117,22 +1117,44 @@ def distanceTo(self, point):
...
+        # For an object that is a box and flat with a polygonal region, we can
+        # also take a fast route.
+        if self._isPlanarBox and (
+            isinstance(other, PolygonalRegion)
+            and abs(self.position.z - other.z) <= self.height / 2
+        ):
+            return self._boundingPolygon.intersects(other.polygons)
+
+        ## Default Case
+        # Extract other's occupied space if it's a region
```
dfremont, line 1146:
```suggestion
        # Extract other's occupied space if it's an object
```

src/scenic/core/object_types.py
```diff
@@ -1117,22 +1117,44 @@ def distanceTo(self, point):
 
     @cached_method
     def intersects(self, other):
-        """Whether or not this object intersects another object"""
-        # For objects that are boxes and flat, we can take a fast route
-        if self._isPlanarBox and other._isPlanarBox:
+        """Whether or not this object intersects another object or region"""
+        ## Type Checking ##
+        if not isinstance(other, (Object, Region)):
```
dfremont, line 1122: This won't work for distributions over objects or over regions, and I think it's possible to get into this function with those (it doesn't look like the veneer function calling this does any checks). We should either make this a `distributionFunction` or add some extra checks. (The former would be a little heavy-handed since it would trigger on objects with any random properties at all, even if they have static poses, but the latter would be more complex to implement.)

docs/reference/operators.rst
Eric-Vin: Good idea

src/scenic/core/object_types.py
Eric-Vin: Hmm good point. I'll add a fix.

tests/syntax/test_operators.py
```diff
@@ -503,6 +503,105 @@ def test_object_in_region_3d():
...
 
 
+# Intersects
+def test_intersects_obj_obj():
+    p = sampleParamPFrom(
+        """
+        obj1 = new Object at (-1,0,0.1), with allowCollisions True
+        obj2 = new Object at (1,0,0), with allowCollisions True
+        obj3 = new Object with width 10, with length 10, with height 10, with allowCollisions True
+        param p = tuple([obj1 intersects obj2, obj1 intersects obj3, obj2 intersects obj3])
```
dfremont, line 513: There's no longer any need to write `tuple([...])` rather than just `(...)` (the old tests did that because we repurposed the tuple notation for vectors prior to Scenic 2).

dfremont (changes requested):
Looks great, just one more tiny comment above.

tests/syntax/test_operators.py
Eric-Vin: Good point, I went ahead and fixed this test and all the others in the file.

dfremont (approved):

PR #184, opened 2023-09-26
Updated Documentation/Type Checking for On Specifier

Eric-Vin:


docs/reference/specifiers.rst
```diff
@@ -110,12 +111,12 @@ on *region*
...
 
-If :prop:`position` is not already specified with priority 1, positions the *base* of the object uniformly at random in the given `Region`, offset by :prop:`contactTolerance` (to avoid a collision).
+If :prop:`position` is not already specified with priority 1, positions the *base* of the object uniformly at random in the given `Region`, on the :prop:`onSurface` of the given `Object`, or with the base of the object at the given vector. The position is always offset by half of :prop:`contactTolerance` (to avoid a collision).
 The base of the object is determined by adding the object's :prop:`baseOffset` to its :prop:`position`.
 
-Note that while :specifier:`on` can be used with `Region`, `Object` and `Vector`, it cannot be used with a distribution containing anything other than `Region`. When used with an object the base of the object being placed is placed on the target object's `onSurface` and when used with a vector the base of the object being placed is set to that position.
+Note that while :specifier:`on` can be used with `Region`, `Object` and `Vector`, it cannot be used with a distribution containing anything other than `Region`.
```
dfremont, line 117: Is this right? I don't see what goes wrong with a distribution over vectors, and indeed that seems like a totally reasonable use case (e.g. `new Object on Uniform(*my_favorite_base_points)`).

docs/reference/specifiers.rst
```diff
@@ -110,12 +111,12 @@ on *region*
...
 The base of the object is determined by adding the object's :prop:`baseOffset` to its :prop:`position`.
 
-Note that while :specifier:`on` can be used with `Region`, `Object` and `Vector`, it cannot be used with a distribution containing anything other than `Region`. When used with an object the base of the object being placed is placed on the target object's `onSurface` and when used with a vector the base of the object being placed is set to that position.
+Note that while :specifier:`on` can be used with `Region`, `Object` and `Vector`, it cannot be used with a distribution containing anything other than `Region`. 
 
-If instead :prop:`position` has already been specified with priority 1, then its value is modified by projecting it onto the given region.
+If instead :prop:`position` has already been specified with priority 1, then its value is modified by projecting it onto the given region. Not that this modifying version of the specifier does not accept a vector.
```
dfremont, line 119:
```suggestion
If instead :prop:`position` has already been specified with priority 1, then its value is modified by projecting it onto the given region (or the :prop:`onSurface` of the given object). Note that this modifying version of the specifier does not accept a vector.
```

docs/syntax_guide.rst
```diff
@@ -208,7 +208,7 @@ Additional specifiers for the :prop:`position` and :prop:`orientation` propertie
      - Positions the object uniformly at random in the given Region
    * - :sampref:`contained in {region}`
      - Positions the object uniformly at random entirely contained in the given Region
-   * - :sampref:`on {region}`
+   * - :sampref:`on ({region} | {Object} | {vector})`
      - Positions the base of the object uniformly at random in the given Region, or modifies the position so that the base is in the Region.
```
dfremont, line 212: Need to update the description too. Making the description short might be hard (since this needs to fit in a table without too much scrolling), so you could break it into two or more entries.

dfremont (changes requested):
Looks fine overall but I made a few suggestions.

docs/reference/specifiers.rst
Eric-Vin: Good point, I'll change the semantics/description.

Eric-Vin:
Made some updates to the docs and typechecking.

dfremont (approved):

PR #187, opened 2023-09-30
Refactor dynamics.py; fix message for rejected simulations

dfremont:
Breaks `dynamics.py` into a package with several files; also fixes the crash that would happen with `verbosity` 2 and up when a simulation is rejected.


Eric-Vin (approved):
Looks good to me!

PR #188, opened 2023-10-01
Mesh Loading Upgrades

Eric-Vin:
Fixes an issue where Trimesh returns a Scene instead of a mesh. Also adds a clearer error message when a shape is not a valid volume, as before we would try to unify it and trigger an assertion.


Eric-Vin:
Putting a hold on this commit while I debug it further. It appears this sometimes returns a list of meshes....

Eric-Vin:
Never mind, it appears this is correct.

PR #191, opened 2023-10-04
[WIP] Better Simulator Error Messages with No Objects

Eric-Vin:


Eric-Vin:
As discussed, we're changing tack on this one to support this use case.

PR #194, opened 2023-10-07
Fix drivable region

dfremont:
`Network.roadRegion` can contain some points slightly outside of `Network.laneRegion`, because roads are formed as unions of their lanes with buffering+erosion to remove small gaps. Since we compute `drivableRegion` using `laneRegion` (unioned with `intersectionRegion`), this means that points at a distance from `drivableRegion` greater than the network tolerance can still have a road at them (according to `Network.roadAt` etc.). This caused a test failure in PR #178.

This PR adds `roadRegion` to `drivableRegion` to remove the discrepancy above. I've also robustified some of the road network tests that have failed recently so that hopefully in the future they will fail closer to when the actual bug is introduced rather than taking several CI runs to first turn up.

### DIFF
--- a/src/scenic/core/regions.py
+++ b/src/scenic/core/regions.py
@@ -2643,10 +2643,14 @@ class PolygonalRegion(Region):
     @distributionFunction
     def unionAll(regions, buf=0):
         regions = tuple(regions)
-        if not all([r.z == regions[0].z for r in regions]):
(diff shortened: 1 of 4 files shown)


Eric-Vin (approved):
Looks good, I patched unionAll as requested so all the tests should pass now.

PR #195, opened 2023-10-11
Sensors

theilem:
Sensors implementation with a Carla example

(EDIT by @dfremont: this PR has been greatly expanded since @theilem's initial draft. @lola831 added documentation and support for MetaDrive and I extended the `record` statement. Thanks Mirco for pioneering this work!)


dfremont:
Hi Mirco, the CI is failing since we now enforce certain formatting requirements. See [here](https://scenic-lang.readthedocs.io/en/latest/developing.html#getting-started) for how to get the formatters to run automatically when you commit; for a one-time fix for the files you've already committed, you can run `make format` in the root directory of the repo.

Speaking of CI, it would be great if you could add some tests for the new functionality to the test suite (maybe in `tests/syntax/test_dynamics.py`, next to the test for the `record` statement). We can't test the CARLA implementation right now, but your test could implement a dummy sensor (e.g. one that returns the ground-truth position) just so we can check that the overall flow works correctly.

PR #197, opened 2023-10-31
3.12 Updates

Eric-Vin:
Various updates for 3.12 compatibility.

This PR incorporates changes from [this](https://github.com/we-like-parsers/pegen/pull/95) PR.

NOTE: This PR should be SQUASHED & MERGED


pyproject.toml
```diff
@@ -35,7 +35,7 @@ dependencies = [
 	"networkx >= 2.6",
 	"numpy ~= 1.24",
 	"opencv-python ~= 4.5",
-	"pegen >= 0.2",
+	"pegen @ git+https://github.com/we-like-parsers/pegen.git",
```
dfremont, line 38: Hmm, this seems a bit risky: does `pegen` make any claims about the stability of their `main` branch? We could ask when they plan to make a new release, and if it isn't imminent we could pin to a specific commit temporarily.

Eric-Vin: Strongly agree, let's discuss offline.

src/scenic/syntax/scenic.gram
```diff
@@ -1101,6 +1237,17 @@ try_stmt[ast.Try]:
     | 'try' &&':' b=block ex=except_block+ el=[else_block] f=[finally_block] {
         ast.Try(body=b, handlers=ex, orelse=el or [], finalbody=f or [], LOCATIONS)
      }
+    | 'try' &&':' b=block ex=except_star_block+ el=[else_block] f=[finally_block] {
```
dfremont, line 1240: We'll have to decide whether we want to allow `except*` (from Python 3.11) in `try-interrupt` statements. Offhand I can't see any use for that, so I'm fine disallowing it as in the current grammar. But then we should probably add a sentence to the end of the documentation for the `try` statement as this is a further difference between Scenic's `try` and Python's.

dfremont (changes requested):
Looks good other than two quick points above.

pyproject.toml
Eric-Vin: Fixed with new Pegen 0.3.0.

src/scenic/syntax/scenic.gram
Eric-Vin: Good call, I've added a note to the docs.

dfremont (approved):

PR #210, opened 2024-01-11
Added Require Monitor Error

Eric-Vin:
Added error when "require MONITOR" is used instead of "require monitor MONITOR", since I imagine this will be a common error that is nasty to debug.


dfremont (approved):
Good catch -- I thought we already had a specific error for this but apparently not!

PR #213, opened 2024-01-30
Added PathRegion Default Orientation

Eric-Vin:


Eric-Vin:
Huh, it seems my local installation of black disagrees with the one in the CI.

src/scenic/core/regions.py
```diff
@@ -2397,11 +2397,21 @@ class PathRegion(Region):
     Args:
         points: A list of points defining a single polyline.
         polylines: A list of list of points, defining multiple polylines.
+        orientation:
```
dfremont, line 2400: Need documentation here (which should mention which orientation is used if this argument is not provided, as well as how you can disable the default orientation entirely by passing `None`).

src/scenic/core/regions.py
```diff
@@ -2397,11 +2397,21 @@ class PathRegion(Region):
...
-    def __init__(self, points=None, polylines=None, tolerance=1e-8, name=None):
-        super().__init__(name)
+    def __init__(
+        self, points=None, polylines=None, tolerance=1e-8, orientation=True, name=None
+    ):
+        if orientation is True:
+            orientation = VectorField("Path", self.defaultOrientation)
+            self.usingDefaultOrientation = True
+        else:
+            self.usingDefaultOrientation = False
```
dfremont, line 2411: Probably better to use `_usingDefaultOrientation` since it's an implementation detail.

Eric-Vin: I copied this over from PolyLine region, should I swap it there as well?

tests/syntax/test_regions.py
```diff
@@ -167,6 +167,19 @@ def test_mesh_region_distribution():
     )
 
 
+def test_path_region_default_orientation():
+    ego = sampleEgoFrom(
+        """
+        import math
+        region = PathRegion(points=[(0,0,0), (1,1,math.sqrt(2)), (2,2,2*math.sqrt(2))])
+        ego = new Object in region
+    """
```
dfremont, line 176: Weird indentation here (should match opening quotes).

src/scenic/core/regions.py
dfremont: Yes. `PolylineRegion` dates from very early days when we had minimal documentation and I wasn't careful about this sort of thing.

tests/syntax/test_regions.py
Eric-Vin: It appears black insists on this style, as it won't even let me change it manually. Perhaps there's a setting we can tune?

src/scenic/core/regions.py
```diff
@@ -886,7 +886,11 @@ def sampleGiven(self, value):
             dimensions=value[self.dimensions],
             position=value[self.position],
             rotation=value[self.rotation],
-            orientation=value[self.orientation],
+            orientation=(
+                True
+                if self.__dict__.get("_usingDefaultOrientation", False)
+                else value[self.orientation]
+            ),
```
dfremont, line 893: What's wrong with just `value[self.orientation]` here? If `self.orientation` is `True` it will get passed through `value` unchanged.

src/scenic/core/regions.py
```diff
@@ -1779,21 +1786,23 @@ class MeshSurfaceRegion(MeshRegion):
         onDirection: The direction to use if an object being placed on this region doesn't specify one.
     """
 
-    def __init__(self, *args, **kwargs):
-        super().__init__(*args, **kwargs)
+    def __init__(self, *args, orientation=True, **kwargs):
+        if orientation is True:
+            orientation = VectorField(
+                "DefaultSurfaceVectorField", lambda pos: self.getFlatOrientation(pos)
```
dfremont, line 1792:
```suggestion
                "DefaultSurfaceVectorField", self.getFlatOrientation
```

dfremont (changes requested):
Looks great, just two nitpicks above. Thanks for fixing all the misaligned quotes!

src/scenic/core/regions.py
Eric-Vin: Hmm, but wouldn't it get converted to the default vector field on line 1790?

dfremont: I get it now -- never mind!

dfremont (approved):

PR #214, opened 2024-02-02
Visible/Non Visible Specifier Changes

Eric-Vin:
The `visible`/`non visible` specifiers now specify position considering the whole occupied space rather than just the position. For example, this means `visible` can place objects that have a corner visible, and `not visible` can place objects that are behind walls.


Eric-Vin:
Going to wait for https://github.com/BerkeleyLearnVerify/Scenic/pull/213 to be merged in, so I can merge in main and fix the formatting check.

docs/reference/specifiers.rst
```diff
@@ -192,8 +192,11 @@ visible [from (*Point* | *OrientedPoint*)]
...
 
-Also optionally specifies :prop:`position` to be a uniformly random point in the :term:`visible region` of the ego, or of the given Point/OrientedPoint if given.
-Note that the position set by this specifier is slightly stricter than simply adding a requirement that the ego :keyword:`can see` the object: the specifier makes the *center* of the object (its :prop:`position`) visible, while the :keyword:`can see` condition will be satisfied even if the center is not visible as long as some other part of the object is visible.
+Also optionally specifies :prop:`position` to be uniformly random over all points that could result in a visible object (note that the above requirement will ensure the object is in fact visible).
+
+.. versionchanged:: 3.0
+
+    This specifier now specifies position uniformly randomly over all points that could result in a visible object. This accounts for objects who's position might be out of the visible region, but who have a portion of their occupied space visible (e.g. a corner that is visible), which would never be generated with the previous semantics.
```
dfremont, line 199:
```suggestion
    This specifier now specifies :prop:`position` uniformly randomly over all points that could result in a visible object. This allows for objects whose :prop:`position` might be out of the visible region, but which have a portion of their occupied space visible (e.g. a corner that is visible). With the previous semantics, such configurations would never be generated because the *center* of the object was required to be visible.
```

docs/reference/specifiers.rst
```diff
@@ -209,8 +212,12 @@ not visible [from (*Point* | *OrientedPoint*)]
 
 Requires that this object is *not* visible from the ego or the given `Point`/`OrientedPoint`.
 
-Similarly to :sampref:`visible [from ({Point} | {OrientedPoint})]`, this specifier can position the object uniformly at random in the *non-visible* region of the ego.
-However, it depends on :prop:`regionContainedIn`, in order to restrict the non-visible region to the :term:`container` of the object being created, which is hopefully a bounded region (if the non-visible region is unbounded, it cannot be uniformly sampled from and an error will be raised).
+Similarly to :sampref:`visible [from ({Point} | {OrientedPoint})]`, this specifier can optionally position the object uniformly at random over all points that could result in a non-visible object (note that the above requirement will ensure the object is in fact not-visible).
```
dfremont, line 215:
```suggestion
Similarly to :sampref:`visible [from ({Point} | {OrientedPoint})]`, this specifier can optionally position the object uniformly at random over all points that could result in a non-visible object (note that the above requirement will ensure the object is in fact not visible).
```

docs/reference/specifiers.rst
```diff
@@ -209,8 +212,12 @@ not visible [from (*Point* | *OrientedPoint*)]
...
+
+.. versionchanged:: 3.0
+
+    This specifier now specifies position uniformly randomly over all points that could result in a non-visible object. This accounts for objects who's position might be out of the visible region, but who have a portion of their occupied space visible (e.g. a corner that is visible), which could be generated with the previous semantics (and now will not).
```
dfremont, line 219: Please make similar edits as to the comment above for `visible`.

src/scenic/core/regions.py
```diff
@@ -1729,6 +1729,11 @@ def inradius(self):
         else:
             return region_distance
 
+    @cached_property
+    @distributionFunction
+    def circumradius(self):
+        hypot(*(self.mesh.extents / 2))
```
dfremont, line 1735: Hmm, does this get used anywhere? There's no `return`, so this always returns `None`. Either this should be removed or tests for it should be added to the mesh region tests.

src/scenic/core/regions.py
```diff
@@ -1742,6 +1747,56 @@ def voxelized(self, pitch, lazy=False):
...
+    def _erodeOverapproximate(self, maxErosion, pitch):
+        """Compute an overapproximation of this region eroded.
+
+        Erode as much as possible, but no more than maxErosion, outputting
+        a VoxelRegion. Note that this can sometimes return a larger region
+        than the original mesh
+        """
+        # Compute a voxel overapproximation of the mesh. Technically this is not
+        # an overapproximation, but one dilation with a rank 3 structuring unit
+        # with connectivity 3 is. To simplify, we just erode one less time than
```
dfremont, line 1760:
```suggestion
        # with connectivity 3 is. To simplify, we just erode one fewer time than
```

src/scenic/syntax/veneer.py
```diff
@@ -1612,10 +1613,29 @@ def VisibleFrom(base):
...
+
+            position = Region.uniformPointIn(buffered_vr)
+
+        return {"position": position, "_observingEntity": base}
+
     return Specifier(
         "Visible/VisibleFrom",
         {"position": 3, "_observingEntity": 1},
-        {"position": Region.uniformPointIn(base.visibleRegion), "_observingEntity": base},
+        DelayedArgument({"width", "length", "height"}, helper),
```
dfremont, line 1638: This should not depend on `width`, etc. in 2D mode. (Unlikely to cause problems, but no reason to break backwards compatibility here.) I'd check for 2D mode in the specifier rather than inside the delayed argument.

tests/syntax/test_specifiers.py
```diff
@@ -508,9 +508,9 @@ def test_visible():
     for i in range(30):
         scene = sampleScene(scenario, maxIterations=10)
         ego, base = scene.objects
-        assert ego.position.distanceTo(base.position) <= 10
-        assert ego.position.x >= base.position.x
-        assert ego.position.y >= base.position.y
+        assert ego.position.distanceTo(base.position) <= 10 + math.sqrt(3 * 0.5**2)
+        assert ego.position.x >= base.position.x - math.sqrt(3 * 0.5**2)
+        assert ego.position.y >= base.position.y - math.sqrt(3 * 0.5**2)
```
dfremont, line 513: Nitpick: the repetition pains me -- can you just compute the radius once right after the scenario and use it here? Likewise for the other tests.

tests/syntax/test_specifiers.py
```diff
@@ -525,7 +525,9 @@ def test_visible_from_point():
     )
     for i in range(20):
         scene = sampleScene(scenario, maxIterations=10)
-        assert scene.egoObject.position.distanceTo(Vector(300, 200)) <= 2
+        assert scene.egoObject.position.distanceTo(Vector(300, 200)) <= 2 + math.sqrt(
+            3 * 0.5**2
```
dfremont, line 529: This is an awful black-ism: if moving to `2 + radius` doesn't fix it, please add `pt = Vector(300, 200)` above the loop or something similar so we can keep this mundane assertion a 1-liner. Likewise for the other test below.

src/scenic/core/regions.py
Eric-Vin: Good point, removed it.

src/scenic/syntax/veneer.py
Eric-Vin: Resolved by new approach where we move this logic to pruning.

docs/new.rst
```diff
@@ -30,6 +30,8 @@ Backwards-incompatible syntax changes:
 	  Classes providing a default value for :prop:`heading` should instead provide a default value for :prop:`parentOrientation`.
 	  Code like :specifier:`with heading 30 deg` should be replaced with the more idiomatic :specifier:`facing 30 deg`.
 
+	* The :specifier:`visible` and :specifier:`not visible` now specify position uniformly randomly over all points that could result in a visible/not visible object (accounting for occlusion). This is different from previous versions, where the specifiers would simply place the center of an object either inside or outside the view region of the observing object.
```
dfremont, line 33:
```suggestion
	* The :specifier:`visible` and :specifier:`not visible` specifiers now specify :prop:`position` uniformly randomly over all points that could result in the object being visible/not visible (accounting for occlusion). This is different from previous versions, where the specifiers would simply place the center of an object either inside or outside the view region of the observing object.
```
Actually I wonder if a higher-level description would be better, to make it clearer that the specifiers morally mean the same thing as before. Something like "The [specifiers] now take into account occlusion and the shapes of objects. In previous versions, they only checked whether the *center* of the object was visible/not visible, ignoring occlusion." People could still click through to the documentation of the specifier for the more precise definition. (We could even add "See :specifier:`visible` for details." if that would help.)

Also this should be under "backwards-incompatible semantics changes", not syntax changes.

docs/reference/specifiers.rst
```diff
@@ -196,7 +196,11 @@ Also optionally specifies :prop:`position` to be uniformly random over all point
...
+	This specifier now specifies :prop:`position` uniformly randomly over all points that could result in a visible object. This allows for objects whose :prop:`position` might be out of the visible region, but which have a portion of their occupied space visible (e.g. a corner that is visible). With the previous semantics, such configurations would never be generated because the *center* of the object was required to be visible.
+
+.. note::
+	
+	As an implementation detial, position is initially set to be sample from `AllRegion` (or the workspace if one has been set). Scenic will then attempt to further restrict the sampled via various pruning techniques, but sometimes this is not possible. If this occurs and Scenic has not been able to further restrict the sampled region from `AllRegion`, an error will be raised at compile time. The simplest way to remedy this is by setting a workspace or specifying position with a higher priority using a different specifier.
```
dfremont, line 203:
```suggestion
	As an implementation detail, :prop:`position` is initially set to be sampled from `everywhere` (or the :term:`workspace` if one has been set). Scenic will then attempt to further restrict the sample region via various pruning techniques, but sometimes this is not possible. If this occurs and Scenic has not been able to further restrict the sampled region from `everywhere`, an error will be raised at compile time. The simplest way to remedy this is by setting a workspace or specifying :prop:`position` with a higher priority using a different specifier.
```

docs/reference/specifiers.rst
```diff
@@ -212,12 +216,11 @@ not visible [from (*Point* | *OrientedPoint*)]
...
+Similarly to :sampref:`visible [from ({Point} | {OrientedPoint})]`, this specifier can optionally position the object uniformly at random over all points that could result in a non-visible object (note that the above requirement will ensure the object is in fact not visible).
 
 .. versionchanged:: 3.0
 
-    This specifier now specifies position uniformly randomly over all points that could result in a non-visible object. This accounts for objects who's position might be out of the visible region, but who have a portion of their occupied space visible (e.g. a corner that is visible), which could be generated with the previous semantics (and now will not).
-
+	This specifier now specifies :prop:`position` uniformly randomly over all points that could result in a non-visible object. This allows for objects whose :prop:`position` might be out of the visible region, but which have a portion of their occupied space visible (e.g. a corner that is visible). With the previous semantics, such configurations would sometimes be generated because the *center* of the object was required to be non-visible
```
dfremont, line 223:
```suggestion
	This specifier now specifies :prop:`position` uniformly randomly over all points that could result in a non-visible object. This disallows objects whose :prop:`position` is out of the visible region, but which have a portion of their occupied space visible (e.g. a corner that is visible). With the previous semantics, such configurations would sometimes be generated because only the *center* of the object was required to be non-visible.
```

src/scenic/core/distributions.py
```diff
@@ -796,23 +801,27 @@ def scalar(thing):
...
         if result is NotImplemented and self.reverse:
-            assert len(rest) == 1
+            assert len(rest) == 1 and len(kwargs) == 0
             rop = getattr(rest[0], self.reverse)
             result = rop(first)
         if result is NotImplemented and self.symbol:
             raise TypeError(
                 f"unsupported operand type(s) for {self.symbol}: "
-                f"'{type(first).__name__}' and '{type(rest[0]).__name__}'"
+                f"'{type(first).__name__}', '{type(rest[0]).__name__}'"
```
dfremont, line 814: Why was this changed? This error message comes from Python (try `4+"a"`).

src/scenic/core/regions.py
```diff
@@ -2200,6 +2224,199 @@ def dilation(self, iterations, structure=None):
...
+        # Determine which faces should be added for each voxel in our extracted surface.
+        point_face_mask_list = []
+
+        def index_in_bounds(index):
+            return all((0, 0, 0) <= index) and all(index < dense_encoding.shape)
+
+        def actual_face(index):
+            return (
+                not index_in_bounds(target_index)
+                or not dense_encoding[tuple(target_index)]
```
dfremont, line 2254: Pretty confusing to reference a local which hasn't been defined yet -- maybe pass it as an argument?

src/scenic/core/regions.py
```diff
@@ -2200,6 +2224,199 @@ def dilation(self, iterations, structure=None):
...
+            )
+            if actual_face(target_index):
+                face_mask[4] = True
+
+            # Bottom
+            target_index = numpy.asarray(
+                [base_index[0], base_index[1], base_index[2] - 1]
+            )
+            if actual_face(target_index):
+                face_mask[5] = True
```
dfremont, line 2303: What about something like this?
```
offsets = (
    [1, 0, 0],
    [-1, 0, 0],
    [0, 1, 0],
    [0, -1, 0],
    [0, 0, 1],
    [0, 0, -1],
)
for i, offset in enumerate(offsets):
    if actual_face(base_index + offset):
        face_mask[i] = True
```

src/scenic/core/regions.py
```diff
@@ -2200,6 +2224,199 @@ def dilation(self, iterations, structure=None):
...
+
+            point_face_mask_list.append((base_center, face_mask))
+
+        # Construct triangles for mesh
+        triangles = []
+
+        for base_center, face_mask in point_face_mask_list:
```
dfremont, line 2310: Since `point_face_mask_list` isn't used after this, why not just combine these 2 loops, eliminate `face_mask`, and directly append to `triangles`?

src/scenic/core/regions.py
```diff
@@ -2720,6 +2937,10 @@ def footprint(self):
     def containsPoint(self, point):
         return self.footprint.containsPoint(point)
 
+    @distributionFunction
+    def _trueContainsPoint(self, point):
+        return point.z == self.z and self.containsPoint
```
dfremont, line 2942:
```suggestion
        return point.z == self.z and self.containsPoint(point)
```
Better add a test which would have caught this!

src/scenic/core/pruning.py
```diff
@@ -236,14 +240,26 @@ def pruneContainment(scenario, verbosity):
...
                 container = container.buffer(-maxErosion)
             elif isinstance(container, MeshVolumeRegion):
-                # We can attempt to erode a voxel approximation of the MeshVolumeRegion.
-                eroded_container = container._erodeOverapproximate(
-                    maxErosion, PRUNING_PITCH
-                )
+                current_pitch = PRUNING_PITCH
+                eroded_container = None
+
+                while eroded_container is None:
```
dfremont, line 246: Don't we need to increase the pitch each iteration to avoid an infinite loop?

src/scenic/core/pruning.py
```diff
@@ -380,79 +396,125 @@ def pruneVisibility(scenario, verbosity):
...
+        ):
             continue
 
-        newBase = base
+        currBase = base
+        currDist = pir_dist
+        currPos = position
 
-        # Define a helper function to buffer an oberver's visibleRegion, resulting
+        # Define a helper function to attempt buffer an oberver's visibleRegion, resulting
```
dfremont, line 419:
```suggestion
        # Define a helper function to attempt to buffer an observer's visibleRegion, resulting
```

src/scenic/core/pruning.py
```diff
@@ -380,79 +396,125 @@ def pruneVisibility(scenario, verbosity):
...
+                        if isinstance(buffered_container, VoxelRegion):
+                            buffered_container = buffered_container.mesh
+
+                        current_pitch = min(2 * current_pitch, 1)
+
+                    assert buffered_container is not None
+
+                    return buffered_container
+            else:
+                return viewRegion
```
dfremont, line 448: Isn't this case unsound? If we intersect with the raw view region we could lose some valid positions where the object would be visible. If we really cannot buffer the view region we should not attempt to prune.

src/scenic/core/pruning.py
```diff
@@ -380,79 +396,125 @@ def pruneVisibility(scenario, verbosity):
...
 
-        # Define a helper function to buffer an oberver's visibleRegion, resulting
+        # Define a helper function to attempt buffer an oberver's visibleRegion, resulting
         # in a region that contains all points that could feasibly be the position
-        # of obj, if it is visible from the observer.
+        # of obj, if it is visible from the observer. If possible buffer exactly, otherwise
+        # try to buffer approximately, and if that is also not feasible just return the viewRegion.
         def bufferHelper(viewRegion):
-            return viewRegion._bufferOverapproximate(obj.radius / 2, PRUNING_PITCH)
+            buffer_quantity = obj.radius / 2 + maxDistance
```
dfremont, line 424: Seems like this should be just `obj.radius`, not half of that (the position could be as far off as `obj.radius` and still have part of the object be in the visible region).

src/scenic/core/pruning.py
```diff
@@ -565,41 +630,30 @@ def percentagePruned(base, newBase):
...
 
-def checkCyclical(A, B):
-    """Check for a potential circular dependency
-
-    Returns True if the scenario would have a circular dependency
-    if A depended on B.
-    """
-    state = collections.defaultdict(lambda: 0)
+def checkConditionedCycle(A, B):
+    """Returns true if A depends on B"""
```
dfremont, line 634: It looks to me like this won't work if (after conditioning) `A` is `B` or directly depends on it (the latter since you don't check if `target_dep` itself is `B`, only its dependencies). Presumably these cases don't arise with the current pruning methods, but we might as well make this function satisfy the description in the docstring in case we use it differently later.

tests/core/test_regions.py
```diff
@@ -613,6 +613,50 @@ def test_empty_erosion():
...
+    reg1 = BoxRegion(position=(0, 0, 0), dimensions=(0.5, 1, 1))
+    reg2 = AllRegion("all")
+    reg3 = CircularRegion((0, 0, 0), 1)
+    reg4 = reg3.footprint
+
+    regions = (reg1, reg2, reg3, reg4)
+
+    intersection_region = IntersectionRegion(*regions)
+
+    def attempt_sample(num_samples):
```
dfremont, line 626: Since this is used in 2 tests and potentially more in the future, why not move it to the top of the file as a utility function? You could call it `sample_ignoring_rejections` or something.

tests/syntax/test_pruning.py
```diff
@@ -208,39 +207,49 @@ def test_visibility_pruning():
...
+        ego = new Object at (0,0,0), with visibleDistance 1
+        foo = new Object on workspace, with requireVisible True,
+            with shape SpheroidShape(dimensions=(0.2,0.2,0.2)),
+            with baseOffset (0,0,{baseOffsetVal}), with contactTolerance 0
+        param p = foo.position
         """
     )
+    positions = [sampleParamP(scenario, maxIterations=100) for i in range(30)]
+    assert all(pos.distanceTo(Vector(0, 0, 0)) <= 1.1 for pos in positions)
+    assert any(pos.distanceTo(Vector(0, 0, 0)) >= 1 for pos in positions)
```
dfremont, line 224: My math may be off, but I'm getting a probability of about 1.6e-05 for this assertion to fail, which is too large (I usually shoot for 1e-18).

tests/syntax/test_specifiers.py
```diff
@@ -505,12 +505,44 @@ def test_visible():
         "             with visibleDistance 10, with viewAngle 90 deg\n"
         "ego = new Object visible"
     )
+    radius = math.sqrt(3 * 0.5**2)
```
dfremont, line 508: `math.hypot(0.5, 0.5, 0.5)` would be clearer here I think.

tests/syntax/test_specifiers.py
```diff
@@ -505,12 +505,44 @@ def test_visible():
...
+        assert ego.position.distanceTo(base.position) <= 10 + radius
+        assert ego.position.x >= base.position.x - radius
+        assert ego.position.y >= base.position.y - radius
+
+
+def test_visible_2d():
+    scenario = compileScenic(
+        "ego = new Object at 100 @ 200, facing -45 deg,\n"
+        "             with visibleDistance 10, with viewAngle 90 deg\n"
+        "ego = new Object visible",
```
dfremont, line 521: New/modified tests should use multiline strings (this old code dates to before `compileScenic` could properly remove the indentation resulting from their use).

dfremont (changes requested):
Looks good overall: a few comments above, but mostly minor.

One other thing that I couldn't leave as a comment: the docs for the `visible` and `not visible` specifiers should now list `regionContainedIn` as a dependency. (Arguably it's an implementation detail, not really part of the semantics, but it is a real dependency so we should probably document it.)

src/scenic/core/distributions.py
Eric-Vin: Huh, I have no idea. I'll change it back.

src/scenic/core/regions.py
Eric-Vin: Good point. I thought we would need to do further filtering but it turned out to be unnecessary.

src/scenic/core/regions.py
Eric-Vin: Good catch, fixed and added test!

src/scenic/core/pruning.py
Eric-Vin: Good catch! I copied this over from the erosion function since we can just do a best effort there. I'll remove it and add an assertion, but the good news is that this should never happen (since we have the approximate fast path).

src/scenic/core/pruning.py
Eric-Vin: Good catch!

src/scenic/core/pruning.py
```diff
@@ -565,41 +630,30 @@ def percentagePruned(base, newBase):
...
-            return False
-        elif state[target] == 1:
-            return True
+    A = conditionedVal(A)
+    B = conditionedVal(B)
 
-        # Set to in-process
-        state[target] = 1
+    while unseen_deps:
+        target_dep = unseen_deps.pop(0)
```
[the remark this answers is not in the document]
Eric-Vin: I wasn't aware of the time complexity difference, thanks for pointing it out! I assumed it was a doubly linked list, but I see that's what `Deque()` is for. I'll make the swap to what you suggested.

src/scenic/core/pruning.py
Eric-Vin: Good catch, I made a change that should fix it.

tests/syntax/test_specifiers.py
Eric-Vin: Sure, I swapped all occurrences in the file.

tests/syntax/test_pruning.py
Eric-Vin: Fixed as discussed over Slack.

tests/syntax/test_pruning.py
```diff
@@ -176,68 +176,73 @@ def test_visibility_pruning():
...
     scenario = compileScenic(
         f"""
         workspace = Workspace(RectangularRegion(0@0, 0, 1e10, 1e10))
-        ego = new Object at (0,0,0), with visibleDistance 1
-        foo = new Object on workspace, visible,
-            with shape SpheroidShape(dimensions=(0.2,0.2,0.2)),
+        ego = new Object at (0,0,0), with visibleDistance 1, with allowCollisions True
+        foo = new Object on workspace,
+            with shape SpheroidShape(dimensions=(2,2,2)),
+            with allowCollisions True, visible,
```
dfremont, line 239: The `visible` is hard to notice here: I'd move it back to being right after `on workspace`.

tests/syntax/test_specifiers.py
```diff
@@ -501,11 +501,13 @@ def test_beyond_from_3d():
...
 def test_visible():
     scenario = compileScenic(
-        "ego = new Object at 100 @ 200, facing -45 deg,\n"
-        "             with visibleDistance 10, with viewAngle 90 deg\n"
-        "ego = new Object visible"
+        """
+        ego = new Object at 100 @ 200, facing -45 deg,
+                     with visibleDistance 10, with viewAngle 90 deg
+        ego = new Object visible
+    """
```
dfremont, line 508:
```suggestion
        """
```

src/scenic/core/regions.py
```diff
@@ -2145,6 +2224,158 @@ def dilation(self, iterations, structure=None):
...
+            base_center = surface_centers[i]
+
+            for i, offset in enumerate(offsets):
+                if actual_face(base_index + offset):
+                    if i == 0:
+                        triangles.append(
+                            [
+                                base_center + [hpitch, hpitch, -hpitch],
+                                base_center + [hpitch, hpitch, hpitch],
+                                base_center + [hpitch, -hpitch, hpitch],
```
dfremont, line 2276: Can you put these sign patterns into an array and then dispense with the need for all the `if` statements? Other than the sign patterns the code is identical in all 6 cases, I think.

dfremont (approved):
Looks great -- thanks Eric for persevering through this long review process!

PR #217, opened 2024-03-04
feat: adding code cov for PyTest including statement & branch coverage

abanuelo:
Adding [`pytest-cov`](https://pypi.org/project/pytest-cov/), a plugin on top of pytest for code coverage along with coverage tool in pyproject.toml:

- shows branches in test files
- show percentage of statements and branches covered
- lines of code missed or needing of testing
- sets a minimum average code coverage threshold of 80% for all test files
- Create new GitHub Workflow for coverage


.coveragerc
```diff
@@ -1,7 +1,11 @@
 # .coveragerc to control coverage.py
 [run]
+branch = True
 omit =
```
abanuelo, line 4: Let me know what other files to omit from the coverage reports

Eric-Vin: For now I think simulator interfaces that can't be run from the test suite should probably be excluded (until we get the CI set up to test them).

Eric-Vin:
Looks great to me! Just have two question: 

- Should we run the example tests as well when calculating coverage? I'm thinking probably not, since the test suite itself should ideally cover all behavior, but I know there's some behavior in certain interfaces.
- Will this calculate coverage for scenic interface files? I know some of them have quite complicated behavior, such as the ground surface generation for Webots.

.coveragerc
abanuelo: Ack! `src/scenic/simulators/*` has been excluded from the overall code coverage report!

abanuelo:
> * late coverage for scenic interface file

1. Eventually, yes! `examples/` was originally excluded from the CI and this PR does not change that. I think we can consider re-adding it to both CI and coverage when they stabilize. 
2. No. Everything within the `src/scenic/simulators/*` has been excluded from coverage.

Eric-Vin:
Whoops, sorry! I meant to say Scenic model files, not the interfaces. Those are `.scenic` files though so I'm not sure if we could even use the same infrastructure to measure coverage for those.

abanuelo:
Correct, I don't think we could use the same code coverage frame work since technically it looks for code coverage around `.py` files. I can look into potentially extending the existing coverage library to find line/branch coverage for `.scenic` files but that would require some work. 

The question here is, how beneficial is line coverage for overall scenic simulation files? My first impression is we just care about expected inputs and outputs of those simulations to be correct and less about line coverage.

.coveragerc
```diff
@@ -1,7 +1,11 @@
...
-    src/scenic/simulators/webots/WBT*.py
+    tests/*
+    */__init__.py
```
dfremont, line 6: What's the reason for excluding the `__init__.py` files? If they are empty, I would think they would have 100% coverage anyway, and if they aren't empty we should probably compute coverage for them, since they do get run when Scenic is used.

.github/workflows/run-coverage.yml
```diff
@@ -0,0 +1,60 @@
...
+      run: |
+        python -m pip install --upgrade pip
+
+    - name: Install Scenic and dependencies
+      run: |
+        python -m pip install -e ".[${{ matrix.extras }}]"
+
+    - name: Run and report code coverage
+      run: |
+        coverage run -m pytest
```
dfremont, line 59: I think if you use `coverage` directly it doesn't capture subprocesses (in the coverage report I don't see any entry for `__main__.py`, presumably because all the tests for it use subprocesses). Probably better to use `pytest --cov`.

dfremont (changes requested):
Looks good -- just a couple minor comments above.

The textual coverage report is pretty hard to read, with the long lists of uncovered lines. Is there a way to display only the per-file statistics and not the details? For actually seeing which lines weren't covered people will probably want to use the HTML report anyway, which this job doesn't generate. (Or are you going to set up Codecov to display a more detailed report in a separate PR?)

.coveragerc
abanuelo: I just checked many of the `__init__.py` files and they were either empty or had some comments. Many had zero lines of coverage to actually report. And yes they would generate 100% coverage, but it would actually skew the average coverage percentage generated across all files. I am happy to re-add as needed!

.github/workflows/run-coverage.yml
abanuelo: Agreed, I did notice a slight coverage percentage difference between `pytest --cov` and `coverage` and reading up on the differences, I believe you are right on the ability to capture subprocesses. I will adjust accordingly!

abanuelo:
This PR is just setting the infrastructure in place to run code coverage. We will need another PR to integrate with CodeCov to export the report as an `.xml` where it can then nicely print out the coverage for the repo! I will defer those changes as needed!

.coveragerc
abanuelo: Showing skew for future runs and leaving in `__init__.py`.

.coveragerc
```diff
@@ -1,7 +1,10 @@
 # .coveragerc to control coverage.py
 [run]
+branch = True
 omit =
-    src/scenic/simulators/webots/WBT*.py
+    tests/*
+    src/scenic/simulators/*
```
dfremont, line 6: Actually let's include `src/scenic/simulators/newtonian` and `src/scenic/simulators/utils` (all other submodules of `scenic.simulators` can be excluded though).

.github/workflows/run-coverage.yml
```diff
@@ -56,5 +56,10 @@ jobs:
 
     - name: Run and report code coverage
       run: |
-        coverage run -m pytest
-        coverage report -m
\ No newline at end of file
+        pytest --cov=. --cov-report json
```
dfremont, line 59: I think `--cov=blah` overrides the `source` option in the pyproject.toml file, so if you want to specify the desired folder(s) in that file we should just put `--cov` here (with no equals sign or path).

On that note, I see from Codecov that it's including the `docs` folder, which makes sense given that we put `.` here. We definitely don't want to include `docs`, so maybe it would be better to just set `source` to `src` and not have to exclude `docs`, `tests`, etc.

Eric-Vin (approved):

dfremont (approved):

PR #220, opened 2024-03-14
fix: fixing syntax for issue template

abanuelo:
Fixed issues include:

1. `ISSUE_TEMPLATE` being within the `.github/workflows/` directory when it should really be within `.github/`. The same applies for `PULL_REQUEST_TEMPLATE.md`. 
3. Some of the `validates: required: true` in the `.yml` files were not properly tabbed. I fixed it and should resolve specific templates not appearing within the Issues tab.
4. I added a `1-*`, `2-*`, giving GitHub instructions on how to order to the issue templates.


.github/ISSUE_TEMPLATE/config.yaml
```diff
@@ -0,0 +1,5 @@
+blank_issues_enabled: true
+contact_links:
+  - name: Questions
+    url: https://forms.gle/uUhQNuPzQrvvBFJX9
+    about: Send your questions via this Google Form
```
abanuelo, line 5: The `url` field does not allow for `mailto:<email>` and requires something of the format `https://`. That being the case I just created a basic Google Form that asks for the users 

1. name
2. email
3. question

If we decide not to move forward with this template I can remove it entirely and re-add it once the forum has been established. Please let me know!

dfremont: Sounds good.

dfremont (approved):

PR #224, opened 2024-03-25
fix: ignoring src/scenic/simulators in codecov report

abanuelo:
### Description
PR #223 merge is blocked by decreased code coverage directly resulting from adding code to carla simulator.py.

### Issue Link
None, adhoc fix needed to merge in PR #223 

### Checklist

### DIFF
--- a/codecov.yml
+++ b/codecov.yml
@@ -15,6 +15,8 @@ codecov:
     layout: "reach, diff, flags, files"
     behavior: default
   require_ci_to_pass: true
+  ignore:
+    - "src/scenic/simulators/**"
 cli:
   plugins:
     pycoverage:
--- a/codecov.yml
+++ b/codecov.yml
@@ -17,6 +17,8 @@ codecov:
   require_ci_to_pass: true
   ignore:
     - "src/scenic/simulators/**"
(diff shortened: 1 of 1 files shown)


codecov.yml
```diff
@@ -15,6 +15,8 @@ codecov:
     layout: "reach, diff, flags, files"
     behavior: default
   require_ci_to_pass: true
```
abanuelo, line 17: nit: we can also remove this to ensure it never blocks CI, let me know your thoughts here!

abanuelo: Resolving conversation as we have agreed to keep it!

codecov.yml
```diff
@@ -15,6 +15,8 @@ codecov:
...
+  ignore:
+    - "src/scenic/simulators/**"
```
dfremont, line 19: Is there a clean way to include `simulators/newtonian` and `simulators/utils` without explicitly listing all the other subfolders?

abanuelo: Yes I can apply a `!` operator! I am going to make that change and let it percolate to see if it has any incompatibilities with `codecov.yml` formatting!

dfremont (approved):
Looks great!

PR #227, opened 2024-04-01
feat: adding dynamic scenario and object creation tests for CARLA

abanuelo:
### Description

Adding the following three test cases:
1. Car applying throttle increases speed
2. Car applying break will bring the car to a halt
3. Can create all Carla objects. For this third one, there was an error with the `BusStop` class within `model.scenic` that needed to be captured. Also note that these tests are automatically skipped as there are memory leak errors with Carla at the moment!

### Checklist


abanuelo:
# Important!!!!

1. In order to trigger the dynamic scenario tests for CARLA we need to have an environment variable exported called `CARLA_ROOT`. As part of the simulator CI, this environment variables will be exported as necessary.
2. The `pyproject.toml` file has been updated such that the only precondition to install the `carla` pypi package is if we have  python version below 3.10. Python 3.11 is giving Carla some problems: https://github.com/carla-simulator/carla/issues/7328

tests/simulators/carla/test_carla.py
```diff
@@ -8,6 +9,412 @@
...
+
+@pytest.mark.skip(reason="Skipping test due to Carla memory leak issues")
+@pytest.mark.parametrize("modelName", walkerModels)
+def test_pedestrian_blueprints(getCarlaSimulator, launchCarlaServer, modelName):
+    pytest.importorskip("carla")
+    simulator, town, mapPath = getCarlaSimulator("Town01")
+    model_blueprint(simulator, mapPath, town, "Pedestrian", modelName)
+
+
+@flaky(max_runs=5, min_passes=1)
```
abanuelo, line 362: **Note:** Introducing `flaky` to re-run dynamic scenario tests multiple times for `CARLA`. Reason being is that when starting the CARLA server via a subprocess via the `launchCarlaServer` pytest fixture, it could be segfault due to the memory leak errors that CARLA is currently experiencing.

pyproject.toml
```diff
@@ -64,7 +65,7 @@ test-full = [  # like 'test' but adds dependencies for optional features
 	"scenic[test]",       # all dependencies from 'test' extra above
 	"scenic[guideways]",  # for running guideways modules
 	"astor >= 0.8.1",
-	'carla >= 0.9.12; python_version <= "3.8" and (platform_system == "Linux" or platform_system == "Windows")',
+	'carla >= 0.9.12; python_version <= "3.10"',
```
Eric-Vin, line 68: Just wondering, why did we change this?

dfremont: I confirmed removing the platform checks breaks installing Scenic with the `test-full` extra on macOS (`pip install ".[test-full]"`), since there is no `carla` wheel for macOS. Better put them back. (Note that we check for the existence of the `carla` package at runtime, so having the install-time checks here doesn't lead to any problems with Scenic expecting `carla` and failing to import it.)

tests/conftest.py
```diff
@@ -52,6 +54,47 @@ def loader(relpath, **kwargs):
...
 
+def checkCarlaPath():
+    CARLA_ROOT = os.environ.get("CARLA_ROOT")
+    if not CARLA_ROOT:
+        pytest.skip("CARLA_ROOT env variable not set.")
+    return CARLA_ROOT
+
+
+@pytest.fixture
+def launchCarlaServer():
```
dfremont, line 65: Let's move all these CARLA-specific things into `tests/simulators/carla/conftest.py`.

tests/conftest.py
```diff
@@ -52,6 +54,47 @@ def loader(relpath, **kwargs):
...
+        f"bash {CARLA_ROOT}/CarlaUE4.sh -RenderOffScreen", shell=True
+    )
+    # NOTE: CARLA server takes time to start up
+    time.sleep(3)
+    yield
+    while carla_process.poll() is None:
+        carla_process.kill()
+        time.sleep(0.1)
+    # NOTE: CARLA server hangs. The first `kill` usually works,
+    # but for safe measure we needed to kill hanging zombie processes
```
dfremont, line 77: I don't understand this: `poll()` is not supposed to return a non-`None` value until the process has terminated, so how can the process still be running when we get here?

I'm also not sure if there's any value in repeatedly sending SIGKILL. We could just send SIGKILL, wait a few seconds, and then send SIGTERM (with `terminate()`) if the process is still open.

tests/conftest.py
```diff
@@ -52,6 +54,47 @@ def loader(relpath, **kwargs):
...
+        carla_process.kill()
+        time.sleep(0.1)
+    time.sleep(5)
+
+
+@pytest.fixture
+def getCarlaSimulator():
+    from scenic.simulators.carla import CarlaSimulator
+
+    base = Path(__file__).parent.parent / "assets" / "maps" / "CARLA"
```
dfremont, line 88:
```suggestion
def getCarlaSimulator(getAssetPath):
    from scenic.simulators.carla import CarlaSimulator

    base = getAssetPath("maps/CARLA")
```

tests/simulators/carla/test_carla.py
```diff
@@ -8,6 +9,412 @@
...
+            param map = r'{mapPath}'
+            param carla_map = '{town}'
+            param time_step = 1.0/10
+
+            model scenic.simulators.carla.model
+            ego = new {modelType} with blueprint '{modelName}'
+            terminate after 1 steps
+        """
+    scenario = compileScenic(code, mode2D=True)
+    scene, _ = scenario.generate(maxIterations=10000)
```
dfremont, line 31:
```suggestion
    scene = sampleScene(scenario)
```
The `sampleScene` helper does some additional checks, and I don't think there's any reason to allow more than 1 iteration (these scenarios should not cause any rejections).

tests/simulators/carla/test_carla.py
```diff
@@ -8,6 +9,412 @@
...
+    else:
+        assert obj.blueprint == modelName
+
+
+from scenic.simulators.carla.blueprints import carModels
+
+
+@pytest.mark.skip(reason="Skipping test due to Carla memory leak issues")
+@pytest.mark.parametrize("modelName", carModels)
+def test_car_blueprints(getCarlaSimulator, launchCarlaServer, modelName):
```
dfremont, line 45:
```suggestion
def test_car_blueprints(getCarlaSimulator, modelName):
```
Likewise for the other tests.

tests/conftest.py
abanuelo: Ack! Will do that accordingly!

tests/simulators/carla/test_carla.py
```diff
@@ -8,6 +9,412 @@
...
+    pytest.importorskip("carla")
```
dfremont, line 46: Let's not keep repeating this check. We have a whole folder for CARLA tests, so let's use it: you can make `test_blueprints.py`, etc. to organize things better. Then you can use `pytest.skip` to skip entire modules which require CARLA when it's not available. (Alternatively, once the crashing problem is fixed we could make `getCarlaSimulator` a module-scoped fixture.)

abanuelo: I was thinking of doing that originally but was unsure if we wanted to restrict all tests into a single file. Will do!

tests/simulators/carla/test_carla.py
abanuelo: I think the `launchCarlaServer` needs to stay in order to bring up the CARLA server. I tested on the `trurl` machine to confirm and that holds!

dfremont: That's surprising, doesn't the `getCarlaSimulator` fixture use `launchCarlaServer` internally?

abanuelo: I could that I suppose, let me try to add that in!

tests/simulators/carla/test_blueprints.py
```diff
@@ -0,0 +1,226 @@
...
+            terminate after 1 steps
+        """
+    scenario = compileScenic(code, mode2D=True)
+    scene = sampleScene(scenario)
+    simulation = simulator.simulate(scene)
+    obj = simulation.objects[0]
+    assert obj.blueprint == modelName
+
+
+@pytest.mark.parametrize("modelName", carModels)
```
Eric-Vin, line 60: Can we compress these to one parameterized test?

tests/simulators/carla/test_actions.py
```diff
@@ -0,0 +1,102 @@
...
+def checkCarlaPath():
+    CARLA_ROOT = os.environ.get("CARLA_ROOT")
+    if not CARLA_ROOT:
+        pytest.skip("CARLA_ROOT env variable not set.")
+    return CARLA_ROOT
+
+
+def launchCarlaServer():
+    CARLA_ROOT = checkCarlaPath()
+    subprocess.Popen(f"bash {CARLA_ROOT}/CarlaUE4.sh -RenderOffScreen", shell=True)
```
dfremont, line 24: I'd worry about leaving CARLA open indefinitely (we might test other simulators and only have the resources to run one at a time). I'd suggest having `getCarlaSimulator` both start and stop CARLA, and give it [package scope](https://docs.pytest.org/en/7.1.x/how-to/fixtures.html#scope-sharing-fixtures-across-classes-modules-packages-or-session) so that the fixture will be shut down after all the CARLA tests are done.

tests/simulators/carla/test_actions.py
```diff
@@ -0,0 +1,102 @@
...
+    simulator, town, mapPath = getCarlaSimulator("Town01")
+    code = f"""
+        param map = r'{mapPath}'
+        param carla_map = '{town}'
+        param time_step = 1.0/10
+
+        model scenic.simulators.carla.model
+
+        behavior DriveThenApplyThrottle():
+            do FollowLaneBehavior() for 2 seconds
```
dfremont, line 64: I think it would be better for this test to not depend on `FollowLaneBehavior`: it's not so clear that the assertion below is really testing `SetThrottleAction`. Just applying the throttle for a few time steps should be enough.

tests/simulators/carla/test_actions.py
```diff
@@ -0,0 +1,102 @@
...
+            take SetThrottleAction(0.9)
+        
+        p = new Point at (369, -326)
+        ego = new Car at p, with behavior DriveThenApplyThrottle
```
dfremont, line 68:
```suggestion
        ego = new Car at (369, -326), with behavior DriveThenApplyThrottle
```

tests/simulators/carla/test_actions.py
```diff
@@ -0,0 +1,102 @@
...
+        record ego.speed as CarSpeed
+        terminate after 4 seconds
```
dfremont, line 70: Any particular reason to make this test slower? I think you could just run the simulation for a few time steps (e.g. `terminate after 2 steps`) and assert the speed is zero (or almost zero) at the beginning and significantly nonzero at the end.

tests/simulators/carla/test_actions.py
```diff
@@ -0,0 +1,102 @@
...
+    simulator, town, mapPath = getCarlaSimulator("Town01")
+    code = f"""
+        param map = r'{mapPath}'
+        param carla_map = '{town}'
+        param time_step = 1.0/10
+
+        model scenic.simulators.carla.model
+
+        behavior DriveThenBrake():
+            do FollowLaneBehavior() for 2 seconds
```
dfremont, line 89: Likewise here I'd rather just use plain constant throttle for some time to build up speed (tested in the previous test) and then apply full braking for the rest of the scenario.

tests/simulators/carla/test_actions.py
```diff
@@ -0,0 +1,102 @@
...
+            take SetBrakeAction(1)
+
+        p = new Point at (369, -326)
+        ego = new Car at p, with behavior DriveThenBrake
+        record ego.speed as CarSpeed
```
dfremont, line 94: If you only need the last value of the speed, you can do `record final ego.speed as CarSpeed` and get a plain number.

tests/simulators/carla/test_actions.py
```diff
@@ -0,0 +1,102 @@
...
+        terminate after 4 seconds
+    """
+    scenario = compileScenic(code, mode2D=True)
+    scene = sampleScene(scenario)
+    simulation = simulator.simulate(scene)
+    records = simulation.result.records["CarSpeed"]
+    threshold = 3
+    assert int(records[-1][1]) < threshold
```
dfremont, line 102: It would be good to document where this threshold comes from. Ideally you can run a scenario with the initial constant throttle suggested above and then set throttle to zero, to confirm that the speed stays above this threshold if there's no braking being applied (only friction).

tests/simulators/carla/test_blueprints.py
dfremont: Yes, that would be great. If Pytest doesn't allow one parameter to depend on another (which wouldn't surprise me), I think it would be fine to only parametrize over the type of object and have a loop inside the test which iterates over the specific blueprints.

dfremont (changes requested):
Looks good overall -- a few suggestions above.

tests/simulators/carla/test_blueprints.py
abanuelo: Ack! I created a dict from which I loop over to further parameterize the `test_blueprints.py` file!

tests/simulators/carla/test_actions.py
abanuelo: Ack! Added cleanup for the carla server subprocess along with adding the `package` scope for the fixture!

tests/simulators/carla/test_actions.py
abanuelo: Great catch! I have adjusted the test to just apply Throttle! It makes things a lot clearer!

tests/simulators/carla/test_actions.py
abanuelo: Ack!

tests/simulators/carla/test_actions.py
abanuelo: Ack! I have created an ApplyThrottle behavior followed immediately by a full braking behavior to make this test a little more clear!

tests/simulators/carla/test_actions.py
abanuelo: Sure I changed the simulation for a total of 50 steps to best illustrate the acceleration applied from throttle along with the full braking case!

tests/simulators/carla/test_actions.py
abanuelo: Great catch! Ack!

tests/simulators/carla/test_actions.py
abanuelo: Now with the requested revisions we just can directly check if the final speed is just zero so no need to document this threshold!

tests/simulators/carla/test_actions.py
```diff
@@ -0,0 +1,133 @@
...
+        if not isCarlaServerRunning():
+            time.sleep(1)
+
+    # extra 3 seconds to ensure server startup
+    time.sleep(3)
+    return
+
+
+def terminateCarlaServer():
+    result = subprocess.run("lsof -i :2000", capture_output=True, text=True, shell=True)
```
dfremont, line 41: This looks somewhat risky and fragile. How about just having `launchCarlaServer` return the `Popen` object and then calling `terminate()` on that here? It looks like you're supporting having CARLA already running when `pytest` starts but I'm not sure if we want to do that given that we're unconditionally killing CARLA here. It's probably better to have our fixture completely manage CARLA, both starting and stopping it (in which case there may be no need to have these helper functions and you could just inline them into the fixture).

tests/simulators/carla/test_actions.py
```diff
@@ -0,0 +1,133 @@
...
+
+@pytest.fixture(scope="package")
+def getCarlaSimulator(request):
+    if not isCarlaServerRunning():
+        launchCarlaServer()
+
+    def cleanup():
+        terminateCarlaServer()
+
+    request.addfinalizer(cleanup)
```
dfremont, line 68: Do you need to use an explicit finalizer here? I think it would work equally well to use a [yield fixture](https://docs.pytest.org/en/7.1.x/how-to/fixtures.html#yield-fixtures-recommended), i.e., change the `return _getCarlaSimulator` to `yield _getCarlaSimulator` and then add the call to `terminateCarlaServer()` after that.

tests/simulators/carla/test_actions.py
dfremont: But 50 time steps of 0.1 seconds is even longer than 4 seconds. Also the assertions haven't changed.

tests/simulators/carla/test_actions.py
```diff
@@ -0,0 +1,133 @@
...
+        model scenic.simulators.carla.model
+
+        behavior DriveWithThrottle():
+            while True:
+                take SetThrottleAction(1)
+
+        behavior Brake():
+            while True:
+                take SetThrottleAction(0)
+                take SetBrakeAction(1)
```
dfremont, line 119:
```suggestion
                take SetThrottleAction(0), SetBrakeAction(1)
```
(You want these to be simultaneous, not alternating.)

tests/simulators/carla/test_actions.py
```diff
@@ -0,0 +1,133 @@
...
+
+        ego = new Car at (369, -326), with behavior DriveThenBrake
+        record final ego.speed as CarSpeed
+        terminate after 50 steps
+    """
+    scenario = compileScenic(code, mode2D=True)
+    scene = sampleScene(scenario)
+    simulation = simulator.simulate(scene)
+    records = simulation.result.records["CarSpeed"]
+    assert records == 0.0
```
dfremont, line 133:
```suggestion
    finalSpeed = simulation.result.records["CarSpeed"]
    assert finalSpeed == pytest.approx(0.0)
```

dfremont (changes requested):
A few more comments (and one which I've unresolved from before).

tests/simulators/carla/test_actions.py
abanuelo: Ack!

tests/simulators/carla/test_actions.py
abanuelo: Ack!

tests/simulators/carla/test_actions.py
abanuelo: Ack!

tests/simulators/carla/test_actions.py
abanuelo: Since the CARLA server is launched from another bash script we need to trigger `killall` for relevant processes. But coupling starting and stopping into the fixture has been performed!

tests/simulators/carla/test_actions.py
abanuelo: Played around with timesteps that made sense for the simulations. Reducing timesteps for `test_throttle` and `test_brake`.

tests/simulators/carla/test_actions.py
```diff
@@ -0,0 +1,118 @@
...
+        sock.settimeout(1)
+        try:
+            sock.connect((host, port))
+            return True
+        except (socket.timeout, ConnectionRefusedError):
+            return False
+
+
+@pytest.fixture(scope="package")
+def getCarlaSimulator():
```
dfremont, line 37:
```suggestion
def getCarlaSimulator(getAssetPath):
```

tests/simulators/carla/test_actions.py
```diff
@@ -0,0 +1,118 @@
...
+
+        for _ in range(30):
+            if isCarlaServerRunning():
+                break
+            time.sleep(1)
+
+        # Extra 5 seconds to ensure server startup
+        time.sleep(5)
+
+    base = Path(__file__).parent.parent.parent.parent / "assets" / "maps" / "CARLA"
```
dfremont, line 53:
```suggestion
    base = getAssetPath("maps/CARLA")
```

tests/simulators/carla/test_actions.py
```diff
@@ -0,0 +1,118 @@
...
+    simulator, town, mapPath = getCarlaSimulator("Town01")
+    code = f"""
+        param map = r'{mapPath}'
+        param carla_map = '{town}'
+        param time_step = 1.0/10
+
+        model scenic.simulators.carla.model
+
+        behavior DriveWithThrottle():
+            take SetThrottleAction(1)
```
dfremont, line 76:
```suggestion
            while True:
                take SetThrottleAction(1)
```

tests/simulators/carla/test_actions.py
```diff
@@ -0,0 +1,118 @@
...
+
+        ego = new Car at (369, -326), with behavior DriveThenBrake
+        record final ego.speed as CarSpeed
+        terminate after 6 steps
+    """
+    scenario = compileScenic(code, mode2D=True)
+    scene = sampleScene(scenario)
+    simulation = simulator.simulate(scene)
+    finalSpeed = simulation.result.records["CarSpeed"]
+    assert finalSpeed == pytest.approx(0.0, abs=1e-1)
```
dfremont, line 118: I'm assuming you've checked that this assertion fails if you change `SetBrakeAction(1)` to `SetBrakeAction(0)` above (simulating `SetBrakeAction` not working) -- if you haven't, please do.

dfremont (changes requested):
Looking good: just a couple last tweaks.

tests/simulators/carla/test_actions.py
abanuelo: Can confirm that this test fails if that's the case:

```
getCarlaSimulator = <function getCarlaSimulator.<locals>._getCarlaSimulator at 0x72a2228f6560>

    def test_brake(getCarlaSimulator):
        simulator, town, mapPath = getCarlaSimulator("Town01")
        code = f"""
            param map = r'{mapPath}'
            param carla_map = '{town}'
            param time_step = 1.0/10
    
            model scenic.simulators.carla.model
    
            behavior DriveWithThrottle():
                while True:
                    take SetThrottleAction(1)
    
            behavior Brake():
                while True:
                    take SetThrottleAction(0), SetBrakeAction(0)
    
            behavior DriveThenBrake():
                do DriveWithThrottle() for 2 steps
                do Brake() for 4 steps
    
            ego = new Car at (369, -326), with behavior DriveThenBrake
            record final ego.speed as CarSpeed
            terminate after 6 steps
        """
        scenario = compileScenic(code, mode2D=True)
        scene = sampleScene(scenario)
        simulation = simulator.simulate(scene)
        finalSpeed = simulation.result.records["CarSpeed"]
>       assert finalSpeed == pytest.approx(0.0, abs=1e-1)
E       assert 0.3394579134514629 == 0.0 ± 1.0e-01
E         comparison failed
E         Obtained: 0.3394579134514629
E         Expected: 0.0 ± 1.0e-01

tests/simulators/carla/test_actions.py:117: AssertionError
=========================== short test summary info ============================
FAILED tests/simulators/carla/test_actions.py::test_brake - assert 0.3394579134514629 == 0.0 ± 1.0e-01

```

tests/simulators/carla/test_actions.py
abanuelo: In order to use this fixture, we need to be update to `session` to avoid the following error:
```
ScopeMismatch: You tried to access the function scoped fixture getAssetPath with a package scoped request object, involved factories:
tests/simulators/carla/test_actions.py:37:  def getCarlaSimulator(getAssetPath)
tests/conftest.py:43:  def getAssetPath()
```

tests/simulators/carla/test_actions.py
abanuelo: See comment above!

tests/simulators/carla/test_actions.py
abanuelo: Ack!

dfremont (approved):
Looks great!

PR #231, opened 2024-04-09
feat: adding webots dynamic scenario tests

abanuelo:
### Description

Adding a webots dynamic scenario test where an object is created above a plane and then drops until it hits the floor. The test checks the expected start and ending z values for the object's position. Note, that webots runs headlessly via a python subprocess. This subprocess then creates a `results.txt` file which contains the starting and ending z positions. Below is an illustration of the test case running on our machine:

https://github.com/BerkeleyLearnVerify/Scenic/assets/32311654/dbade0af-69fc-48a9-b4c5-ea84c27d68ab

### Checklist


abanuelo:
# Important!!!! 
In order to for dynamic webot tests to work without a display port when you either:
1.  `ssh` into the instance and trigger the webots tests manually 
2. run the ci that triggers simulation tests on the instance

it is **VERY** important you specify the display port prior to running the tests!
```
export DISPLAY=:1 
```

If not you will generate the following warning which will automatically fail all webots dynamic scenario tests:
```
Warning: Could not find the Qt platform plugin "offscreen" in ""
Fatal: This application failed to start because no Qt platform plugin could be initialized. Reinstalling the application may fix this problem.

Available platform plugins are: xcb.

/usr/local/bin/webots: line 105: 12490 Aborted                 (core dumped) "$webots_home/bin/webots-bin" "$@"
```

If you are using the DCV client to run the test (as illustrated in the PR description above), no need to export `DISPLAY` since its automatically captured by the DCV session!

tests/simulators/webots/dynamic/webots_data/controllers/scenic_supervisor/scenic_supervisor.py
```diff
@@ -0,0 +1,30 @@
...
+
+supervisor = Supervisor()
+simulator = WebotsSimulator(supervisor)
+
+path = supervisor.getCustomData()
+print(f"Loading Scenic scenario {path}")
+scenario = scenic.scenarioFromFile(path)
+
+scene, _ = scenario.generate()
+simulation = simulator.simulate(scene, verbosity=2)
```
Eric-Vin, line 24: Maybe add a large upper bound to the time limit so if the test fails, it does so gracefully?

lola831 (approved):
I've addressed the comment, added the timeout, and tested the changes locally to ensure everything works as expected.

Issue #228, opened 2024-04-04
Adding default `carla_map` value if `map` present in .scenic files


abanuelo:
### Describe the feature and motivation

When running some examples within the `examples/carla/` directory, there were some files such as `car.scenic` that were throwing errors as a result of a missing `param carla_map` in file. By default, the `carla_map` is set to `None`. The recommended feature enhancement would be making this value set to the town being loaded in from the `param map` variable if `carla_map` is missing from the `.scenic` file.

For example:
```
param map = localPath('../../assets/maps/CARLA/Town01.xodr')
model scenic.simulators.carla.model

ego = new Car
```

would implicitly assume that `param carla_map='Town01'`, from the `map` parameter.

### Issue Submission Checklist

- [X] I checked the open issues, forum, etc. and have not found any solution

abanuelo:
Just for replication purposes, if you were to run the `badlyParkedCarPullingIn.scenic` example by running:
```
scenic examples/driving/badlyParkedCarPullingIn.scenic --2d --simulate --model scenic.simulators.carla.model --count 1
```
and comment out [this line](https://github.com/BerkeleyLearnVerify/Scenic/blob/0232e6f8ac3d34e7c2e47ced3e29532815dd7a91/examples/driving/badlyParkedCarPullingIn.scenic#L2) defining the `carla_map` param, it will yield a simulation that looks as follows

https://github.com/BerkeleyLearnVerify/Scenic/assets/32311654/c88b0b67-92dc-4392-8ba8-4d13810ff3e1

PR #237, opened 2024-04-17
fix: defaulting carla_map to map .xodr path in scenic files

abanuelo:
### Description

When `carla_map` is undefined in a `.scenic` file it defaults to `None` when it should default to the `map` town if the `map` param is present in the file. 

### Issue Link

#228 

### Checklist

### DIFF
--- a/src/scenic/simulators/carla/model.scenic
+++ b/src/scenic/simulators/carla/model.scenic
@@ -70,7 +70,8 @@ except ModuleNotFoundError:
     class _CarlaVehicle: pass
     class _CarlaPedestrian: pass
 
-param carla_map = None
+map_town = ((str(globalParameters.map).split('/'))[-1]).split('.')[0]
+param carla_map = map_town if globalParameters.map else None
 param address = '127.0.0.1'
 param port = 2000
 param timeout = 10


src/scenic/simulators/carla/model.scenic
```diff
@@ -70,7 +70,8 @@ except ModuleNotFoundError:
     class _CarlaVehicle: pass
     class _CarlaPedestrian: pass
 
-param carla_map = None
+map_town = ((str(globalParameters.map).split('/'))[-1]).split('.')[0]
```
dfremont, line 73: I don't think this will work with Windows paths (using backslashes as separators). Maybe something like `pathlib.Path(globalParameters.map).stem`?

src/scenic/simulators/carla/model.scenic
```diff
@@ -70,7 +70,8 @@ except ModuleNotFoundError:
...
+param carla_map = map_town if globalParameters.map else None
```
dfremont, line 74: What's the reason for the check here? If `map` is the empty string, an error will have already been raised when we tried to open the map file.

dfremont (changes requested):
The idea is good -- just a couple comments above, and we also need a test for the new feature. The test can just compile a very simple scenario like
```
param map = ...
model scenic.simulators.carla.model
```
and test that the `carla_map` parameter gets set properly in the compiled `Scenario`.

src/scenic/simulators/carla/model.scenic
abanuelo: Ack! I remove the redundant checks for setting `map_town` here!

src/scenic/simulators/carla/model.scenic
abanuelo: Ack! Giving this a go!

Eric-Vin (approved):
Looks good to me, though we should check with @dfremont to see if he'd rather we just ensure the defaults are consistent.

abanuelo:
@dfremont I have added the assertion to check if the `carla_map` parameter is being properly set within the test suite. Please let me know if this suffices!

tests/simulators/carla/test_carla.py
```diff
@@ -8,6 +8,18 @@
 )
 
 
+def test_map_param_parse(getAssetPath):
+    pytest.importorskip("carla")
```
dfremont, line 12: This test doesn't require the `carla` package, so let's remove this line. (Presumably this restriction is why we're getting the coverage warning below.)

dfremont (changes requested):
Looks good except the new test isn't actually running since it says it requires `carla`. Should be good to go after fixing that.

tests/simulators/carla/test_carla.py
abanuelo: Ack! Will make that change now!

dfremont (approved):

PR #240, opened 2024-04-20
Color tweaks.

Eric-Vin:
### Description
Updated color specification process to ensure we have a unified internal representation and ensure we pass colors to Trimesh in the correct form. Tested on test suite and examples.

### Issue Link
N/A

### Checklist

### Additional Notes
N/A

### DIFF
--- a/src/scenic/core/object_types.py
+++ b/src/scenic/core/object_types.py
@@ -497,6 +497,13 @@ class Constructible(Samplable):
                     "Color property contains value not between 0 and 1 (inclusive)."
                 )
 
+            if len(value) == 3:
+                value = (value[0], value[1], value[2], 1)
+            if len(value) == 4:
+                pass
+            else:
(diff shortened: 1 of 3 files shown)


src/scenic/core/object_types.py
```diff
@@ -497,6 +497,13 @@ def _specify(cls, context, prop, value):
                     "Color property contains value not between 0 and 1 (inclusive)."
                 )
 
+            if len(value) == 3:
+                value = (value[0], value[1], value[2], 1)
+            if len(value) == 4:
+                pass
+            else:
```
dfremont, line 504: Why not just `elif len(value) != 4:`?

Eric-Vin:
Unfortunately I haven't found a way to open the Trimesh viewer when running `pytest`, at least on my machine. I can definitely add the length test though.

dfremont (approved):
Looks good!

PR #241, opened 2024-04-22
Orientation Composition Fixes

Eric-Vin:
### Description
Fixed orientations being composed backwards. Also added several tests and updated some documentation.

### Issue Link
N/A

### Checklist

### Additional Notes
N/A

### DIFF
--- a/src/scenic/core/vectors.py
+++ b/src/scenic/core/vectors.py
@@ -324,6 +324,12 @@ class Orientation:
 
     # will be converted to a distributionMethod after the class definition
     def __mul__(self, other) -> Orientation:
+        """To compose rotations, i.e. apply rotation A followed by rotation B,
+        one should compute A*B. Note that as rotations are represented intrinsically,
+        A should come first.
+        See https://en.wikipedia.org/wiki/Davenport_chained_rotations#Conversion_to_extrinsic_rotations
+        for more details.
+        """
         if type(other) is not Orientation:
             return NotImplemented
         # Preserve existing orientation objects when possible to help pruning.
@@ -339,7 +345,7 @@ class Orientation:
             other = Orientation._fromHeading(other)
         elif type(other) is not Orientation:
             return NotImplemented
-        return other * self
+        return self * other
 
     @distributionMethod
     def __radd__(self, other) -> Orientation:
@@ -347,7 +353,7 @@ class Orientation:
             other = Orientation._fromHeading(other)
         elif type(other) is not Orientation:
             return NotImplemented
-        return self * other
+        return other * self
 
     def __repr__(self):
         return f"Orientation.fromEuler{tuple(self.eulerAngles)!r}"
@@ -362,7 +368,7 @@ class Orientation:
         That is, considering ``self`` as the parent orientation, find the Euler angles
         expressing the given orientation.
         """
-        local = orientation * self.inverse
+        local = self.inverse * orientation
         return local.eulerAngles
 
     @distributionFunction
--- a/src/scenic/core/vectors.py
+++ b/src/scenic/core/vectors.py
@@ -324,9 +324,11 @@ class Orientation:
 
     # will be converted to a distributionMethod after the class definition
     def __mul__(self, other) -> Orientation:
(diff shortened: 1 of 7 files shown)


src/scenic/core/vectors.py
```diff
@@ -324,6 +324,12 @@ def _inverseRotation(self):
 
     # will be converted to a distributionMethod after the class definition
     def __mul__(self, other) -> Orientation:
+        """To compose rotations, i.e. apply rotation A followed by rotation B,
```
dfremont, line 327: This doesn't have the right format for a docstring (a la [PEP 257](https://peps.python.org/pep-0257/), which the documentation builder expects): the current discussion should be in a second paragraph, with a 1-line summary at the top separated by a blank line. Like:
```
"""Apply a rotation to this orientation, yielding a new orientation.

As we represent orientations as intrinsic rotations, rotation A followed by rotation B is
given by the quaternion product A*B, not B*A. See...
"""
```

src/scenic/core/vectors.py
```diff
@@ -362,7 +368,7 @@ def localAnglesFor(self, orientation) -> typing.Tuple[float, float, float]:
         That is, considering ``self`` as the parent orientation, find the Euler angles
         expressing the given orientation.
         """
-        local = orientation * self.inverse
+        local = self.inverse * orientation
```
dfremont, line 373: This formula isn't obvious, so let's add a unit test for this function (in `tests/core/test_vectors.py`). The following works:
```
parent = Orientation.fromEuler(pi/4, pi/4, 0)
target = Orientation.fromEuler(-3*pi/4, pi/4, pi/4)
angles = parent.localAnglesFor(target)
assert angles == pytest.approx((-3*pi/4, pi/2, 0))
```
By waving my hands around I can convince myself the resulting angles do in fact rotate the parent onto the target. Actually we might as well test that relationship more robustly:
```
for i in range(100):
    parent = Orientation.fromEuler(*[random.uniform(-pi, pi) for _ in range(3)])
    target = Orientation.fromEuler(*[random.uniform(-pi, pi) for _ in range(3)])
    local = Orientation.fromEuler(*parent.localAnglesFor(target))
    assert target.approxEq(parent * local)
```

tests/syntax/test_operators.py
```diff
@@ -688,6 +688,21 @@ def test_orientation_relative_to_orientation():
     assert ego.orientation.approxEq(Orientation.fromEuler(math.pi / 2, math.pi / 2, 0))
 
 
+def test_orientation_relative_to_orientation2():
+    ego = sampleEgoFrom(
+        """
+        ego = new Object facing (Orientation.fromEuler(-135 deg, 45 deg, 0)
+        relative to Orientation.fromEuler(90 deg, 0, 0))
+    """
```
dfremont, line 696:
```suggestion
            relative to Orientation.fromEuler(90 deg, 0, 0))
        """
```

tests/syntax/test_specifiers.py
```diff
@@ -1069,6 +1070,20 @@ def test_facing_vf_3d():
     )
 
 
+def test_facing_equivalence():
+    p = sampleParamPFrom(
+        """
+        a = new OrientedPoint facing (Orientation.fromEuler(-135 deg, 45 deg, 0)
+        relative to Orientation.fromEuler(90 deg, 0, 0))
```
dfremont, line 1077:
```suggestion
            relative to Orientation.fromEuler(90 deg, 0, 0))
```

src/scenic/core/vectors.py
Eric-Vin: Good idea! Test added.

dfremont: Can you please add the first test too? It's not subsumed by the second test since it's checking the specific angles returned, not just the round-trip.

Eric-Vin: Whoops, sure thing.

dfremont (approved):
Looks good! I confirmed objects with non-global parent orientations now correctly apply `yaw`, etc. in the built-in visualizer and in CARLA (with the appropriate coordinate conversion fixes -- I'll send those separately).

PR #242, opened 2024-04-23
Add Qiangcheng Wu to credits

dfremont:
### Description

Adds Qiangcheng Wu to the credits, per Eddie.

### Checklist


ek65 (approved):

ek65:
Codecov's error is pointing out that test cases are not added which is not necessary in this case.

PR #244, opened 2024-04-23
feat: adding coordinate system to docs

abanuelo:
### Description

As discussed, there is a need to have a single page describing the Scenic coordinate system. 

### Issue Link

N/A

### Checklist

### DIFF
--- a/.github/workflows/check-formatting.yml
+++ b/.github/workflows/check-formatting.yml
@@ -5,21 +5,24 @@ on:
   pull_request:
   workflow_call:
 
+permissions:
+  contents: read
+
 jobs:
   check:
     runs-on: ubuntu-latest
     steps:
       - name: Checkout code
-        uses: actions/checkout@v3
+        uses: actions/checkout@eef61447b9ff4aafe5dcd4e0bbf5d482be7e7871
         with:
           persist-credentials: false
 
       - name: Run black to check formatting
-        uses: psf/black@stable
+        uses: psf/black@8a737e727ac5ab2f1d4cf5876720ed276dc8dc4b
         with:
           version: "25.1.0"
 
       - name: Run isort to check import order
-        uses: isort/isort-action@v1
+        uses: isort/isort-action@24d8a7a51d33ca7f36c3f23598dafa33f7071326
         with:
           isort-version: "5.12.0"
--- a/.github/workflows/run-coverage.yml
+++ b/.github/workflows/run-coverage.yml
@@ -17,6 +17,9 @@ on:
         description: Git ref on which to run the tests.
         type: string
 
+permissions:
+  contents: read
+
 jobs:
   coverage:
     strategy:
@@ -29,21 +32,21 @@ jobs:
 
     steps:
     - name: Checkout given ref
-      uses: actions/checkout@v3
+      uses: actions/checkout@eef61447b9ff4aafe5dcd4e0bbf5d482be7e7871
       if: inputs.ref != ''
       with:
         ref: ${{ inputs.ref }}
         persist-credentials: false
 
     - name: Checkout current branch
-      uses: actions/checkout@v3
+      uses: actions/checkout@eef61447b9ff4aafe5dcd4e0bbf5d482be7e7871
       if: inputs.ref == ''
       with:
         ref: ${{ github.ref }}
         persist-credentials: false
 
     - name: Set up Python ${{ matrix.python-version }}
-      uses: actions/setup-python@v4
+      uses: actions/setup-python@42375524e23c412d93fb67b49958b491fce71c38
       with:
         python-version: ${{ matrix.python-version }}
         cache: 'pip'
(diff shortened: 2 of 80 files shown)


docs/tutorials/coordinates.rst
```diff
@@ -0,0 +1,51 @@
...
+    * Reference on Orientations: :ref:`Orientation`.
+    * Reference on Heading: :ref:`Heading`
+
+Local Coordinate System
+-----------------------
+If you want to specify an objects coordinate system relative to other objects, there are specifiers for the :prop:`orientation` to do so. 
+
+    1. The :specifier:`on {region}` specifier specifies :prop:`parentOrientation` whenever the region in question has a :term:`preferred orientation`: a `vector field` (another primitive Scenic type) which defines an orientation at each point in the region.
+    2. The :specifier:`ahead of` specifier only specifies :prop:`parentOrientation` *optionally*, giving it a new default value: if you want a different value, you can override that default by explicitly writing :specifier:`with parentOrientation {value}`.
+    3. The general operator :scenic:`{X} relative to {Y}` can interpret vectors and orientations as being in a variety of local coordinate systems
```
Eric-Vin, line 43: We should add more detail here and/or link tot he specifier page, since this specifier can be particularly confusing. Specifically, tuple syntax (e.g. (3,3,3)) are interpreted as Vectors, and can give un-intiuitive results. See [here](https://scenic-lang.readthedocs.io/en/latest/reference/operators.html#direction-relative-to-direction) for more details.

docs/tutorials/coordinates.rst
```diff
@@ -0,0 +1,51 @@
...
+    * Tutorial on Orientations :ref:`orientations_tutorial`
+    * Reference on Orientations: :ref:`Orientation`.
+    * Reference on Heading: :ref:`Heading`
+
+Local Coordinate System
+-----------------------
+If you want to specify an objects coordinate system relative to other objects, there are specifiers for the :prop:`orientation` to do so. 
+
+    1. The :specifier:`on {region}` specifier specifies :prop:`parentOrientation` whenever the region in question has a :term:`preferred orientation`: a `vector field` (another primitive Scenic type) which defines an orientation at each point in the region.
+    2. The :specifier:`ahead of` specifier only specifies :prop:`parentOrientation` *optionally*, giving it a new default value: if you want a different value, you can override that default by explicitly writing :specifier:`with parentOrientation {value}`.
```
Eric-Vin, line 42: Maybe we can generalize this? I think most position specifiers can optionally specify parentOrientation at level 3, and on does it at level 2.

Eric-Vin (commented):
Looks great overall! A few comments:

- We should probably adjust the [Orientation](https://scenic-lang.readthedocs.io/en/latest/reference/data.html#orientation) entry now that we have this page, maybe cleaning it up a little and linking to the coordinate systems page.
-  It doesn't look like this page is available in the sidebar. Maybe we should add it under Language Reference?
- We might want to instruct users on how to manually compose rotations, i.e. that A followed by B is `A*B`, which is the opposite of typical quaternions since we're using intrinsic angles. See [here](https://en.wikipedia.org/wiki/Davenport_chained_rotations#Conversion_to_extrinsic_rotations) for more details. @dfremont any thoughts on this?

docs/tutorials/coordinates.rst
lola831: I’ve updated the "Local Coordinate System" section in the coordinates documentation to add more detail on this issue. Now, there’s a note explaining that tuple syntax (e.g. (3, 3, 3)) is interpreted as a Vector and can yield unintuitive results. I’ve also added a reference link to the specifiers page for additional context.

docs/tutorials/coordinates.rst
lola831: I made the suggested changes in commit f8abfdf2a40ea272b97cee85fd1326924d0515dc

docs/reference/coordinates.rst
```diff
@@ -26,26 +26,33 @@ Below is an illustration of the global coordinate system
...
+This corresponds to the :prop:`heading` property in older versions of Scenic. :prop:`heading` represents yaw in the global XY plane. Scenic represents headings in radians, measured anticlockwise from North, so that a heading of 0 is due North and a heading of π/2 is due West.
 
 For more information on orientations and heading please consult:
 
     * Tutorial on Orientations :ref:`orientations_tutorial`
     * Reference on Orientations: :ref:`Orientation`.
     * Reference on Heading: :ref:`Heading`
 
+.. note::
+    When manually composing rotations in Scenic, remember that if rotation **A** is applied first, followed by rotation **B**, the resulting orientation is computed as **A * B**. This multiplication order is reversed relative to the typical extrinsic rotation convention, reflecting Scenic’s use of intrinsic Euler angles.
```
Eric-Vin, line 38: This is great, thanks Lola! Maybe we should link to [this](https://en.wikipedia.org/wiki/Davenport_chained_rotations#Conversion_to_extrinsic_rotations) Wikipedia page here as well? It might be helpful for users. @dfremont any thoughts?

docs/tutorials/coordinates.rst
Eric-Vin: Looks great!

docs/tutorials/coordinates.rst
Eric-Vin: Looks great!

Eric-Vin (commented):
Looks great! Just one minor comment. We'll probably want @dfremont 's thoughts as well before we merge.

docs/reference/coordinates.rst
lola831: I’ve added the Wikipedia link to the note as suggested.

Issue #176, opened 2023-09-10
Incorrect roadDirection


abol-karimi:
Example script:
```
param carla_map = 'Town05'
carla_map = globalParameters.carla_map
param map = f'/home/carla/CarlaUE4/Content/Carla/Maps/OpenDrive/{carla_map}.xodr'
param weather = 'CloudySunset'
model scenic.simulators.carla.model

intersection = network.elements['intersection396']
lane = intersection.incomingLanes[4]
point = new OrientedPoint at lane.centerline[0]

ego = new Car following roadDirection from point for 0

for d in (10, 20, 30, 40):
  car = new Car following roadDirection from point for d
```
which I run in python with:
```
import scenic
scenario = scenic.scenarioFromFile('road_direction.scenic',
                                   mode2D=True)
scene, _ = scenario.generate(maxIterations=1)
sim_result = scenario.getSimulator().simulate(scene, maxSteps=600)
```
produces this scene:
![Wrong_roadDirection](https://github.com/BerkeleyLearnVerify/Scenic/assets/25490389/d22f1915-f9ab-499c-a2c6-57d333587d29)

Platform:

- Carla 0.9.14 docker
- Python 3.8.17 compiled from source
- Scenic 74a3ebdad896e70e8bf777c267f3adb36edf91a6

Eric-Vin:
Hi Abel, I'll take a look at this. It's possible that the map information is incorrect, but it could also be an issue on Scenic's side. Can you provide a seed to reproduce this, ideally using the internal visualizer? Something like:

abol-karimi:
`scenic road_direction.scenic --2d` produces correct results (with any seeds), for example:

![roadDirection](https://github.com/BerkeleyLearnVerify/Scenic/assets/25490389/b9100326-12f5-43c3-95ff-5d0f9559c353)

so it seems that the bug is in the transformation to Carla's coordinates.

Eric-Vin:
Hmm interesting. I'll investigate.

abol-karimi:
[...]

PR #245, opened 2024-04-23
Fix orientation conversion for CARLA

dfremont:
### Description

Fixes the conversions between Scenic orientations and CARLA rotations so that they work for arbitrary 3D orientations.
### Issue Link

#176 was probably a combination of this bug and the broader Scenic bug fixed in #241; this PR fixes the scenario provided in #176.
### Checklist

### Additional Notes

Once the CARLA CI is set up we can add tests for this conversion; there isn't really anything we can add to the main test suite, unfortunately.

For future reference, the discussion of CARLA's conventions for rotations in #176 was wrong (CARLA's documentation is absolutely unhelpful here): in the course of fixing this issue I determined by experiment that CARLA uses the same ZXY intrinsic rotation sequence as Scenic, except that the Z rotation (yaw) is left-handed and the global orientation points East. I've added a comment to this effect in the code.

### DIFF
--- a/src/scenic/simulators/carla/utils/utils.py
+++ b/src/scenic/simulators/carla/utils/utils.py
@@ -33,7 +33,9 @@ def scenicToCarlaLocation(pos, z=None, world=None, blueprint=None):
 
 
 def scenicToCarlaRotation(orientation):
-    pitch, yaw, roll = orientation.r.as_euler("XZY", degrees=True)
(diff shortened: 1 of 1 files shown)


abanuelo:
@dfremont regarding: 
> Once the CARLA CI is set up we can add tests for this conversion; there isn't really anything we can add to the main test suite, unfortunately.

What tests did you have in mind for the CI?

Eric-Vin (approved):
Confirmed that this fixed the bug occurring in Issue #176

Issue #230, opened 2024-04-07
Scenic only generates one scene per call


Kai-X-Org:
### System Details

1. Python 3.11.0
2. Scenic 3.0.0b2
3. Apple M1 Pro MacOS Ventura 13.4
4. No simulator

### Detailed Description

When running a scenic file from the command line without setting the --count flag, Scenic only generates one scene. After closing the window popped up by clicking the red x, Scenic terminates rather than generate another scene.

### Steps To Reproduce

I simply ran `ego = new Object` in a Scenic file from the command line.

### Issue Submission Checklist

- [X] I am reporting an issue, not asking a question
- [X] I checked the open issues, forum, etc. and have not found any solution
- [x] I have provided all necessary code, etc. to reproduce the issue

ek65:
Hi, @Kai-X-Org, I could not replicate your error on my end. 
Could you try running a scenic program in examples/webots/vacuum/vacuum_simple.scenic and see if you run into the same issue?

abanuelo:
@Eric-Vin @Kai-X-Org I have confirmed that this error exists. Namely, when you try to close out the pygame display it only generates a single simulation and when you try to close it out, it prevents you from doing so. 

My script:
```
model scenic.simulators.newtonian.model
ego = new Object
```

My command:
```
scenic examples/carla/test.scenic \
    --2d       \
    --simulate \
    --model scenic.simulators.newtonian.model -b
```

Here is a video illustrating the process. Note, the only way to quit the simulation is to have a `KeyboardInterrupt` which is not ideal because it does not allow for subsequent simulations to continue.

https://github.com/BerkeleyLearnVerify/Scenic/assets/32311654/1e75b7f6-ebed-4cc5-a833-393840ae28e8

abanuelo:
Note, I think this may be the intended behavior for the Cocoa implementation that is used in MacOS environments by `pyglet`. More details can be reported in this open issue: https://github.com/mikedh/trimesh/issues/2131

lola831:
This issue is now resolved by [PR #330](https://github.com/BerkeleyLearnVerify/Scenic/pull/330).

PR #248, opened 2024-04-25
fix: Apple Silicon devices can generate multiple scenarios

abanuelo:
### Description

For this fix, I noticed that when if we use `pyglet>=1.5.28` in our `pyproject.toml` it will use default to shutting down the generated Scenario immediately after the [`.show()` is called to render the trimesh Scene](https://github.com/BerkeleyLearnVerify/Scenic/blob/d846a15bced1b5fd3030101116d6eb5af89d4a34/src/scenic/core/scenarios.py#L234). To resolve this I played with installing earlier versions of `pyglet` and landed on `1.5.26` as a good alternative to allow us to create multiple scenarios on Apple silicon devices, as is the expected behavior. Below is a video illustrating the fix:

https://github.com/BerkeleyLearnVerify/Scenic/assets/32311654/4c439741-985d-44b4-80f5-5978f785d3c6

### Issue Link

#230 

### Checklist

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -40,7 +40,7 @@ dependencies = [
 	"pillow ~= 9.1",
 	'pygame >= 2.1.3.dev8, <3; python_version >= "3.11"',
 	'pygame ~= 2.0; python_version < "3.11"',
-	"pyglet ~= 1.5",
+	"pyglet == 1.5.26",
 	"python-fcl >= 0.7",
 	"Rtree ~= 1.0",
 	"rv-ltl ~= 0.1",


Eric-Vin:
Hmm, I'm hesitant to pin a specific version since we might block downstream users who are using Scenic as a library. Is this a bug fix where we can set a floor instead?

abanuelo:
@dfremont and I discussed the proposed fix in this PR and he suggested we make the cieling for `pyglet` 1.5.26 but if `trimesh` does upgrade its library to support `>2.0` `pyglet` we can adjust accordingly.

lola831:
With PR [#330](https://github.com/BerkeleyLearnVerify/Scenic/pull/330) now updating our pyglet dependency to allow version 2.0.18—which fixes the Apple Silicon window termination issue—we no longer need these workarounds.

PR #251, opened 2024-04-26
Add Tox plugin to work around cyclic dependency on verifai

dfremont:
### Description

The `tox` environments with the `test-full` extra (`py38-extras` etc.) recently broke, since they install `verifai`, whose cyclic dependency on `scenic` causes issues with the way that `tox` installs dependencies (see discussion [here](https://github.com/tox-dev/tox/discussions/3273)). Until we get a better solution, this PR adds a Tox plugin which works around the problem.

### Issue Link

n/a

### Checklist

### DIFF
--- a/toxfile.py
+++ b/toxfile.py
@@ -1,0 +1,17 @@
+import tox.plugin
+
+
+@tox.plugin.impl
+def tox_on_install(tox_env, arguments, section, of_type):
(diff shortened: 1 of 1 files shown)


Eric-Vin:
Looks good to me, though I hope we can get a permanent solution soon.

Eric-Vin (approved):

Issue #222, opened 2024-03-21
Bug Report: Unable to override behavior if no behavior is given initially


beyazit-yalcinkaya:
### System Details

1. Python 3.9.10
2. Scenic 3.0.0b2
3. 2.3 GHz Quad-Core Intel Core i7 macOS Sonoma 14.3.1 (23D60)
4. Newtonian

### Detailed Description

If a behavior is not specified for an agent initially, its behavior cannot be overridden afterward by another scenario.

### Steps To Reproduce

1. To reproduce the issue run the following python script and scenic scenario.
Scenario:
```
""" Scenario Description
Based on 2019 Carla Challenge Traffic Scenario 03.
Leading vehicle decelerates suddenly due to an obstacle and 
ego-vehicle must react, performing an emergency brake or an avoidance maneuver.
"""
param map = localPath('../../../tests/scenic/Town01.xodr')
# param map = localPath('../../tests/scenic/Town01.xodr')
param carla_map = 'Town01'

model scenic.simulators.newtonian.driving_model

#CONSTANTS
EGO_SPEED = 10
THROTTLE_ACTION = 0.6
BRAKE_ACTION = 1.0
EGO_TO_OBSTACLE = Range(-20, -15)
EGO_BRAKING_THRESHOLD = 11

behavior FollowTrajectoryBehaviorAvoidCollision(target_speed, trajectory):
    try:
        do FollowTrajectoryBehavior(target_speed=target_speed, trajectory=trajectory)

    interrupt when withinDistanceToAnyObjs(self, EGO_BRAKING_THRESHOLD):
        take SetBrakeAction(BRAKE_ACTION)

scenario SubScenario1(target):
    setup:
        # print(f"X = {X}")
        # print(f"EGO_TO_OBSTACLE = {EGO_TO_OBSTACLE}")
        override target with behavior FollowLaneBehavior(EGO_SPEED)

scenario SubScenario2(target, direction):
    setup:
        # print(f"X = {X}")
        # print(f"EGO_TO_OBSTACLE = {EGO_TO_OBSTACLE}")
        current_lane = target.lane
        straight_manuevers = filter(lambda i: i.type == direction, current_lane.maneuvers)
        maneuver = None
        if len(straight_manuevers) > 0:
            maneuver = Uniform(*straight_manuevers)
        assert maneuver
[...]

PR #252, opened 2024-04-27
Behavior Override Fix

Eric-Vin:
### Description
Fixes the bug in Issue https://github.com/BerkeleyLearnVerify/Scenic/issues/222 , where one had issues overriding behavior if it was set to `None`. Also fixed a bug where the `lastActions` property wasn't being set properly, and added various tests.

### Issue Link
https://github.com/BerkeleyLearnVerify/Scenic/issues/222

### Checklist

### Additional Notes
N/A

### DIFF
--- a/src/scenic/core/simulators.py
+++ b/src/scenic/core/simulators.py
@@ -331,7 +331,6 @@ class Simulation(abc.ABC):
         self.result = None
         self.scene = scene
         self.objects = []
-        self.agents = []
         self.trajectory = []
         self.records = defaultdict(list)
         self.currentTime = 0
@@ -441,6 +440,10 @@ class Simulation(abc.ABC):
             if maxSteps and self.currentTime >= maxSteps:
                 return TerminationType.timeLimit, f"reached time limit ({maxSteps} steps)"
 
+            # Clear lastActions for all objects
+            for obj in self.objects:
+                obj.lastActions = tuple()
+
             # Compute the actions of the agents in this time step
             allActions = OrderedDict()
             schedule = self.scheduleForAgents()
@@ -472,11 +475,14 @@ class Simulation(abc.ABC):
                 # Save actions for execution below
                 allActions[agent] = actions
 
+                # Log lastActions
+                agent.lastActions = actions
+
             # Execute the actions
             if self.verbosity >= 3:
                 for agent, actions in allActions.items():
                     print(f"      Agent {agent} takes action(s) {actions}")
-                    agent.lastActions = actions
+
             self.actionSequence.append(allActions)
             self.executeActions(allActions)
 
@@ -521,8 +527,6 @@ class Simulation(abc.ABC):
 
         # Add the new object to our lists.
         self.objects.append(obj)
-        if obj.behavior:
-            self.agents.append(obj)
 
         # Enable dynamic proxy for the object so that any mutations will not
         # affect the original object (e.g. if the simulator sets some of its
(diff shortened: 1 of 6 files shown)


tests/syntax/test_modular.py
```diff
@@ -809,6 +811,91 @@ def test_override_behavior():
...
+    scene, _ = generateChecked(scenario, 1)
+    assert scene.objects[0].prop == 1
+    sampleActionsFromScene(
+        scene,
+        maxIterations=1,
+        maxSteps=1,
+        singleAction=True,
+        asMapping=False,
+        timestep=1,
+    )
```
dfremont, line 865:
```suggestion
    sampleActionsFromScene(scene)
```
(Writing out all the parameters is confusing since you don't actually change any of them from their default values. The defaults are set to be what we want for most tests, and only the ones which need to be changed are written out in other tests.)

tests/utils.py
```diff
@@ -92,7 +92,7 @@ def sampleEgoActions(
         asMapping=False,
         timestep=timestep,
     )
-    return [actions[0] for actions in allActions]
+    return [actions[0] if actions else None for actions in allActions]
```
dfremont, line 95: What is this change for? Shouldn't we always return whatever action(s) came out of `sampleActions` here? (The latter already converts an empty tuple to `None` if `singleAction` is enabled.)

dfremont (changes requested):
Looks good, just a few nitpicks.

tests/utils.py
Eric-Vin: Hmm, I added this because at this stage actions was either empty or `None` so `actions[0]` was causing an exception.

src/scenic/core/simulators.py
```diff
@@ -294,7 +294,10 @@ class Simulation(abc.ABC):
         timestep (float): Length of each time step in seconds.
         objects: List of Scenic objects (instances of `Object`) existing in the
             simulation. This list will change if objects are created dynamically.
-        agents: List of :term:`agents` in the simulation.
+        agents: List of :term:`agents` in the simulation. An agent is any object that has
+            or had a behavior at any point in the simulation. The agents list may have objects
+            appended to the end as the simulation progresses (if a non-agent object has its
+            behavior overriden), but once an object is in the agents list its position is fixed.
```
dfremont, line 300:
```suggestion
            behavior overridden), but once an object is in the agents list its position is fixed.
```

tests/syntax/test_modular.py
```diff
@@ -809,6 +810,77 @@ def test_override_behavior():
...
+        """
+        scenario Main():
+            setup:
+                ego = new Object with prop 1
+            compose:
+                do Sub1()
+        scenario Sub1():
+            setup:
+                override ego with prop 2
+                override ego with behavior Bar
```
dfremont, line 849:
```suggestion
                override ego with prop 2, with behavior Bar
```
This is clearer, I think -- otherwise someone might think the test is about some bad interaction between multiple `override` statements.

tests/syntax/test_modular.py
```diff
@@ -809,6 +810,77 @@ def test_override_behavior():
...
```
dfremont, line 870:
```suggestion
                override ego with prop 2, with behavior Bar
```

dfremont (changes requested):
Looks good -- one last minor nitpick.

dfremont (approved):

Issue #229, opened 2024-04-06
Problem with carla simulation (only 2d allowed)


davidscientistcom:
### System Details

 1. python version: Python 3.10.0
 2. Scenic version: Scenic 3.0.0b2
 3. OS Version: Microsoft Windows 11
 4. Simulator version: Carla 0.9.15

### Detailed Description

It is impossible to load carla in 3d simulator. Only the 2d version is possible to load even with the simplest possible script.

![error](https://github.com/BerkeleyLearnVerify/Scenic/assets/118633949/eda9df51-7902-4ba0-b31c-6e200ef1cfe9)
(3_10) PS D:\carla_dev\Scenic\examples\carla> scenic -b  .\car.scenic 
Beginning scenario construction...
  Loading cached version of road network...
  Loaded cached network in 0.11 seconds.
D:\carla_dev\Scenic\examples\carla\3_10\lib\site-packages\shapely\set_operations.py:419: RuntimeWarning: divide by zero encountered in unary_union
  return lib.unary_union(collections, **kwargs)
Traceback (most recent call last):
  File "C:\Users\battl\AppData\Local\Programs\Python\Python310\lib\runpy.py", line 196, in _run_module_as_main
    return _run_code(code, main_globals, None,
  File "C:\Users\battl\AppData\Local\Programs\Python\Python310\lib\runpy.py", line 86, in _run_code
    exec(code, run_globals)
  File "D:\carla_dev\Scenic\examples\carla\3_10\Scripts\scenic.exe\__main__.py", line 4, in <module>
    from scenic.__main__ import dummy
  File "D:\carla_dev\Scenic\examples\carla\3_10\lib\site-packages\scenic\__main__.py", line 194, in <module>
    scenario = errors.callBeginningScenicTrace(
  File "D:\carla_dev\Scenic\examples\carla\3_10\lib\site-packages\scenic\core\errors.py", line 282, in callBeginningScenicTrace
    return func()
  File "D:\carla_dev\Scenic\examples\carla\3_10\lib\site-packages\scenic\__main__.py", line 195, in <lambda>
    lambda: translator.scenarioFromFile(
  File "D:\carla_dev\Scenic\examples\carla\3_10\lib\site-packages\scenic\syntax\translator.py", line 142, in scenarioFromFile
    return _scenarioFromStream(
[...]

PR #254, opened 2024-05-02
Driving Domain 3D Compatibility

Eric-Vin:
### Description
Allows for limited 3D mode functionality in the driving domain and associated simulators.

### Issue Link
https://github.com/BerkeleyLearnVerify/Scenic/issues/229

### Checklist

### Additional Notes
N/A


src/scenic/core/object_types.py
```diff
@@ -1769,11 +1769,11 @@ def __init_subclass__(cls):
...
 
             props = cls._scenic_properties
-            # Raise error if parentOrientation already defined
-            if "parentOrientation" in props:
+            # Raise error if parentOrientation and heading already defined
+            if "parentOrientation" in props and "heading" in props:
                 raise RuntimeError(
                     "this scenario cannot be run with the --2d flag (the "
-                    f'{cls.__name__} class defines "parentOrientation")'
+                    f'{cls.__name__} class defines "parentOrientation" and "heading")'
```
dfremont, line 1776: Hmm, I think still saying `this scenario cannot be run with the --2d flag` is misleading, since the scenario won't work without that flag either (you'd get a `"heading" property cannot be overridden` error). How about `<name> defines both "parentOrientation" and "heading"`?

src/scenic/core/scenarios.py
```diff
@@ -533,6 +533,8 @@ def generateDefaultRequirements(self):
         for obj in filter(
             lambda x: x.requireVisible and x is not self.egoObject, self.objects
         ):
+            if not self.egoObject:
+                raise InvalidScenarioError("requireVisible set to true with no ego.")
```
dfremont, line 537: Better add a test covering this. Also I think `requireVisible set to true but no ego is defined` would be clearer.

src/scenic/domains/driving/model.scenic
```diff
@@ -17,6 +17,12 @@ If you are writing a generic scenario that supports multiple maps, you may leave
 ``map`` parameter undefined; then running the scenario will produce an error unless the
 user uses the :option:`--param` command-line option to specify the map.
 
+The ``use2DMap`` global parameter determines whether or not maps are generated in 2D. Currently
+3D maps are not supported, but are under development. If 2D mode is enabled, this parameter is
+set to `True`, and otherwise it is set to `False` by default. If 2D mode is disabled, this 
+flag can still be manually set to `True`, allowing full Scenic 3D functionality, albeit with
+2D maps.
```
dfremont, line 24: Something like this might be clearer:

```
By default, this parameter is `False` (so that future versions of Scenic will automatically use
3D maps), unless :ref:`2D compatibility mode` is enabled, in which case the default is `True`.
The parameter can be manually set to `True` to ensure 2D maps are used even if the
scenario is not compiled in 2D compatibility mode.
```

src/scenic/domains/driving/model.scenic
```diff
@@ -38,6 +44,23 @@ from scenic.domains.driving.behaviors import *
 from scenic.core.distributions import RejectionException
 from scenic.simulators.utils.colors import Color
 
+## 2D mode flag & checks
+
+def is2DMode():
+    from scenic.syntax.veneer import mode2D
+    return mode2D
+
+if 'use2DMap' not in globalParameters:
```
dfremont, line 53: You should double-check, but I don't think this conditional is necessary: `param` statements in world models loaded via the `model` statement do not override `param` statements elsewhere. (This is why `scenic.simulators.carla.model` for example can provide a default value for the `weather` parameter without having to worry about whether the user provided a different value before declaring they want to use the CARLA model.)

src/scenic/domains/driving/model.scenic
```diff
@@ -38,6 +44,23 @@ from scenic.domains.driving.behaviors import *
...
+    param use2DMap = True if is2DMode() else False
+
+if is2DMode() and not globalParameters.use2DMap:
+    raise RuntimeError('if using 2D mode global parameter "use2DMap" must be False')
```
dfremont, line 57:
```suggestion
    raise RuntimeError('in 2D mode, global parameter "use2DMap" must be True')
```

src/scenic/simulators/carla/model.scenic
```diff
@@ -18,6 +18,8 @@ Global Parameters:
     timestep (float): Timestep to use for simulations (i.e., how frequently Scenic
         interrupts CARLA to run behaviors, check requirements, etc.), in seconds. Default
         is 0.1 seconds.
+    snapToGroundDefault (bool): Default value for snapToGround on `CarlaActor` objects.
```
dfremont, line 21:
```suggestion
    snapToGroundDefault (bool): Default value for :prop:`snapToGround` on `CarlaActor` objects.
```

src/scenic/simulators/carla/model.scenic
```diff
@@ -18,6 +18,8 @@ Global Parameters:
...
+        Default is True if 2D mode is enabled and false otherwise.
```
dfremont, line 22:
```suggestion
        Default is True if :ref:`2D compatibility mode` is enabled and False otherwise. 
```

src/scenic/simulators/carla/model.scenic
```diff
@@ -95,6 +97,8 @@ param weather = Uniform(
     'MidRainSunset',
     'HardRainSunset'
 )
+if "snapToGroundDefault" not in globalParameters:
```
dfremont, line 100: As above, can get rid of this conditional.

src/scenic/simulators/carla/utils/utils.py
```diff
@@ -25,11 +26,13 @@ def scenicToCarlaVector3D(x, y, z=0.0):
...
 
 
-def scenicToCarlaLocation(pos, z=None, world=None, blueprint=None):
-    if z is None:
+def scenicToCarlaLocation(pos, world=None, blueprint=None, snapToGround=False):
+    if snapToGround:
         assert world is not None
-        return snapToGround(world, carla.Location(pos.x, -pos.y, 0.0), blueprint)
-    return carla.Location(pos.x, -pos.y, z)
+        return scenic.simulators.carla.utils.utils.snapToGround(
```
dfremont, line 32: It's strange to use a fully-qualified name for this purely internal function. Why not rename the function to `_snapToGround`? Alternatively you could make it a pure function and call it something like `_groundSpawnPointFor`.

tests/syntax/test_basic.py
```diff
@@ -292,3 +292,35 @@ def test_mode2D_interference():
...
+
+            class Bar(Foo):
+                parentOrientation: 1.2
+
+            ego = new Bar
+        """
+
+    scenario = compileScenic(program, mode2D=True)
+    scene, _ = scenario.generate()
+    obj = scene.objects[0]
```
dfremont, line 310:
```suggestion
    obj = sampleEgo(program, mode2D=True)
```
And likewise below.

dfremont (changes requested):
Looks good -- just a few minor comments.

dfremont (approved):

PR #258, opened 2024-05-07
fix: adding on-call reminders to every Wednesday

abanuelo:
### Description

0 is Sunday when I assumed 0 as Monday. Updating the on-call reminder to fire on Wednesday mornings when on-call rotations are meant to be handed off!

### Checklist


dfremont (approved):
Thanks, I'd missed this when I reviewed the original PR!

PR #259, opened 2024-05-08
Fix cross-entropy tests and leakage issue

dfremont:
### Description

The tests for the interface with VerifAI's cross-entropy sampler keep failing occasionally since I chose the parameters poorly. This PR improves the tests in various ways and should ensure the false positive rate is extremely low. In the process I noticed an issue whereby the distribution of one sampling run would leak into the next (even after calling `resetExternalSampler`), so I've fixed that and added an assertion catching it.

### Issue Link

n/a

### Checklist


Eric-Vin (approved):
A little over my head but nothing jumped out as wrong!

PR #267, opened 2024-05-16
chore: Adding flag for `pyglet` downgrade for Apple Silicon Devices

abanuelo:
### Description

As a temporarily workaround in place, we are working to add some steps in the documentation FAQs to get Apple Silicon users of Scenic to downgrade to `pyglet` 1.5.26 to be able to generate multiple scenes.

### Issue Link

#230 

### Checklist

### DIFF
--- a/codecov.yml
+++ b/codecov.yml
@@ -17,6 +17,7 @@ codecov:
   require_ci_to_pass: true
   ignore:
     - "src/scenic/simulators/**"
+    - "tests/**"
     - "!src/scenic/simulators/newtonian/**"
     - "!src/scenic/simulators/utils/**"
 cli:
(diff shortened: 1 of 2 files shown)


dfremont (changes requested):
It's not totally clear, but I think you're suggesting people edit `pyproject.toml`? I'm not sure that's a great idea since it's somewhat error-prone and changes their local version relative to what's in the repo. What about just telling them to go through the usual installation (including activating a virtual environment) and then run `pip install pyglet==1.5.26`? Then checking the version of pyglet shouldn't be necessary, but if you want to add that I'd suggest `pip show pyglet` since it will print a lot less unrelated information.

lola831:
With PR [#330](https://github.com/BerkeleyLearnVerify/Scenic/pull/267#330) now updating our pyglet dependency to allow version 2.0.18—which fixes the Apple Silicon window termination issue—we no longer need these workarounds.

Issue #261, opened 2024-05-09
city_intersection worlds `city_intersection.wbt` file is broken when loaded into Webots


abanuelo:
### System Details

1. Python version: 3.10.12
2. Scenic Version: 3.0.0b2
3. Operating System: Linux
4. Simulator Version: Webots 2023a

### Detailed Description

1. When you load the `city_intersection.wbt` file into Webots, it generates the stack trace below
2. `scenic` runs when you load the `city_intersection.wbt` file
3. Error log:
```
INFO: scenic_supervisor: Starting controller: /home/ubuntu/Scenic/venv/bin/python -u scenic_supervisor.py
WARNING: autonomous_vehicle: Could not find controller file:
WARNING: autonomous_vehicle: Expected either: autonomous_vehicle, autonomous_vehicle.jar, autonomous_vehicle.class, autonomous_vehicle.py, autonomous_vehicle.m  or Dockerfile
INFO: autonomous_vehicle: Try to compile the C/C++ source code, to get a new executable file.
WARNING: autonomous_vehicle: Starts the <generic> controller instead.
INFO: <generic>: Starting controller: /software/webotsR2023a/resources/projects/controllers/generic/generic
WARNING: autonomous_vehicle: Could not find controller file:
WARNING: autonomous_vehicle: Expected either: autonomous_vehicle, autonomous_vehicle.jar, autonomous_vehicle.class, autonomous_vehicle.py, autonomous_vehicle.m  or Dockerfile
INFO: autonomous_vehicle: Try to compile the C/C++ source code, to get a new executable file.
WARNING: autonomous_vehicle: Starts the <generic> controller instead.
INFO: <generic>: Starting controller: /software/webotsR2023a/resources/projects/controllers/generic/generic
Traceback (most recent call last; use -b to show Scenic internals):
  File "/home/ubuntu/actions/Scenic/examples/webots/city_intersection/controllers/scenic_supervisor/scenic_supervisor.py", line 26, in <module>
    sim_results = simulator.simulate(scene, timestep=0.5, verbosity=2)
  File "/home/ubuntu/Scenic/src/scenic/simulators/webots/simulator.py", line 56, in createSimulation
    return WebotsSimulation(
  File "/home/ubuntu/Scenic/src/scenic/simulators/webots/simulator.py", line 82, in __init__
    super().__init__(scene, timestep=timestep, **kwargs)
[...]

PR #279, opened 2024-06-11
Minor Fixes for 3.0

Eric-Vin:
### Description
Adjusts a pruning test to be less difficult and fixes a Webots example to account for a change to the `localPath` function.

### Issue Link
https://github.com/BerkeleyLearnVerify/Scenic/issues/261

### Checklist

### Additional Notes
N/A

### DIFF
--- a/examples/webots/city_intersection/city_intersection.scenic
+++ b/examples/webots/city_intersection/city_intersection.scenic
@@ -52,7 +52,7 @@ class LogImageAction(Action):
     def applyTo(self, obj, sim):
         print("Other Car Visible:", self.visible)
         
-        target_path = self.path + "/"
+        target_path = str(self.path) + "/"
         target_path += "visible" if self.visible else "invisible"
 
         if not os.path.exists(target_path):
--- a/tests/syntax/test_pruning.py
+++ b/tests/syntax/test_pruning.py
@@ -55,14 +55,14 @@ def test_containment_2d_region():
 
     # Test both combined, in a slightly more complicated case.
     # Specifically, there is a non vertical component to baseOffset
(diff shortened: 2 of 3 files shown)


examples/webots/city_intersection/city_intersection.scenic
```diff
@@ -52,7 +52,7 @@ class LogImageAction(Action):
     def applyTo(self, obj, sim):
         print("Other Car Visible:", self.visible)
         
-        target_path = self.path + "/"
+        target_path = str(self.path) + "/"
```
dfremont, line 55: While we're changing this, we should probably make it work on Windows too by switching to `pathlib`. (I think the `os.path` and `shutil` functions accept path-like objects, so you might not even need to change the code below; using the `pathlib` equivalents would be fine too.)

tests/syntax/test_pruning.py
```diff
@@ -55,14 +55,14 @@ def test_containment_2d_region():
 
     # Test both combined, in a slightly more complicated case.
     # Specifically, there is a non vertical component to baseOffset
-    # that should be accounted for and the height is random.
+    # that should be accounted.
```
dfremont, line 58:
```suggestion
    # that should be accounted for.
```

dfremont (changes requested):
Looks fine, although I think we should fix the path manipulation (see above).

examples/webots/city_intersection/city_intersection.scenic
Eric-Vin: Good point, I'll make the change!

dfremont (approved):
Looks good!

Issue #276, opened 2024-06-05
VerifaiRange sampling values far beyond range (Fixed with solution mentioned on Slack, posting for the record)


Kai-X-Org:
### System Details

1.Python 3.10.14
2.Scenic 3.0.0rc1, with latest pull of VerifAI
3.Ubuntu 20.04 on x86
4. No simulator. Just Scenic

### Detailed Description

1. When sampling Scenic values using VerifaiRange, certain programs shows sampling far beyond range or just weird sampling behaviors.
2. For instance, consider the following program
```
param x1 = VerifaiRange(5, 6.5)
param y1 = VerifaiRange(0, 1.5)
param yaw1 = VerifaiRange(0, 180)

param x2 = VerifaiRange(5, 6.5)
param y2 = VerifaiRange(0, 3)
param yaw2 = VerifaiRange(0, 180)

param x3 = VerifaiRange(5, 6.5)
param y3 = VerifaiRange(0, 3)
param yaw3 = VerifaiRange(0, 180)

param x4 = VerifaiRange(-0.5, 1)
param y4 = VerifaiRange(0, 0.5)


obj1 = new Object at (globalParameters.x1, globalParameters.y1, 0), with yaw globalParameters.yaw1,
                    with shape SpheroidShape(dimensions=(0.5, 0.5, 0.5)), with color[0, 0, 0]
obj1 = new Object at (globalParameters.x2, globalParameters.y2, 0), with yaw globalParameters.yaw2,
                    with shape SpheroidShape(dimensions=(0.5, 0.5, 0.5)), with color[0.25, 0, 0]
obj1 = new Object at (globalParameters.x3, globalParameters.y3, 0), with yaw globalParameters.yaw3,
                    with shape SpheroidShape(dimensions=(0.5, 0.5, 0.5)), with color[0.5, 0, 0]
obj1 = new Object at (globalParameters.x4, globalParameters.y4, 0), with shape SpheroidShape(dimensions=(0.5, 0.5, 0.5)), with color[0.75, 0, 0]
```
The sampled Scene using the Scenic GUI is shown in the first picture, where some objects are far out of range in position. Each object has radius 0.5m. The solution mentioned on Slack worked with changing the body of scenic.core.external_params.VerifaiSampler.valueFor to return getattr(self.cachedSample, f'param{param.index}')

[...]

PR #280, opened 2024-06-11
Fix Verifai Sampler with more than 10 objects

tcdanielh:
### Description
Fixes problem with `VerifaiSampler` where there may be strange behavior when there are more than 10 samples in the scene. The problem is similar to the problem that was fixed recently in the VerifAI repo. The enumeration of objects goes as follows if there are more than 10 objects: 0, 1, 10, 2, 3, …, 9 when it expects 0, 1, 2, 3, …, 10 which causes a mismatch. This PR should fix this issue by checking that the parameter names match.

This issue might pop up for other external samplers as well, but `VerifaiSampler` is fixed at least.

### Issue Link
#276 

### Checklist

### Additional Notes
N/A

### DIFF
--- a/src/scenic/core/external_params.py
+++ b/src/scenic/core/external_params.py
@@ -195,7 +195,7 @@ class VerifaiSampler(ExternalSampler):
                 usingProbs = True
         space = verifai.features.FeatureSpace(
             {
-                f"param{index}": verifai.features.Feature(param.domain)
+                self.nameForParam(index): verifai.features.Feature(param.domain)
                 for index, param in enumerate(self.params)
             }
         )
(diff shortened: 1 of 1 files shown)


src/scenic/core/external_params.py
```diff
@@ -262,7 +262,12 @@ def getSample(self):
         return self.sampler.getSample()
 
     def valueFor(self, param):
-        return self.cachedSample[param.index]
+        return getattr(self.cachedSample, self.nameForParam(param.index))
+
+    @staticmethod
+    def nameForParam(i):
+        """Paramter name for a given index in the Feature Space."""
```
dfremont, line 269:
```suggestion
        """Parameter name for a given index in the Feature Space."""
```

dfremont (approved):
Thanks for fixing this, Daniel! I'll add a test for this in a new PR (I can't easily add it to this one).

Issue #256, opened 2024-05-06
oppositeLaneGroup is broken for scenic.domains.driving.model


abanuelo:
### System Details

1. Python Version: 3.10.12
2. Scenic Version: 3.0.0b2
3. Operating System: Linux
4. Simulator Version: CARLA 0.9.14

### Detailed Description

1. Scenic's `oppositeLaneGroup` is not working in the scenic.domains.driving.model
2. Scenic command: `scenic examples/driving/pedestrianAcrossRoad.scenic --2d --model scenic.simulators.carla.model --simulate -b`
3. Error:
```
Beginning scenario construction...
pygame 2.5.2 (SDL 2.28.2, Python 3.10.12)
Hello from the pygame community. https://www.pygame.org/contribute.html
  Loading cached version of road network...
  Loaded cached network in 0.11 seconds.
Traceback (most recent call last):
  File "/home/abanuelo/Scenic/venv/bin/scenic", line 5, in <module>
    from scenic.__main__ import dummy
  File "/home/abanuelo/Scenic/src/scenic/__main__.py", line 195, in <module>
    scenario = errors.callBeginningScenicTrace(
  File "/home/abanuelo/Scenic/src/scenic/core/errors.py", line 282, in callBeginningScenicTrace
    return func()
  File "/home/abanuelo/Scenic/src/scenic/__main__.py", line 196, in <lambda>
    lambda: translator.scenarioFromFile(
  File "/home/abanuelo/Scenic/src/scenic/syntax/translator.py", line 142, in scenarioFromFile
    return _scenarioFromStream(
  File "/home/abanuelo/Scenic/src/scenic/syntax/translator.py", line 172, in _scenarioFromStream
    compileStream(stream, namespace, compileOptions, filename)
  File "/home/abanuelo/Scenic/src/scenic/syntax/translator.py", line 321, in compileStream
    executeCodeIn(code, namespace)
  File "/home/abanuelo/Scenic/src/scenic/syntax/translator.py", line 560, in executeCodeIn
    exec(code, namespace)
  File "/home/abanuelo/Scenic/examples/driving/pedestrianAcrossRoad.scenic", line 12, in <module>
    new Pedestrian on visible ego.oppositeLaneGroup.sidewalk
  File "/home/abanuelo/Scenic/src/scenic/syntax/veneer.py", line 1465, in On
    if isA(target, Region) and alwaysProvidesOrientation(target):
[...]

PR #287, opened 2024-07-07
Heuristic Sampling Improvements

Eric-Vin:
### Description
Improves heuristic sampling functionality in the veneer. Now everything should raise RejectionException, which is converted to an InvalidScenarioError at a higher level if appropriate.

### Issue Link
https://github.com/BerkeleyLearnVerify/Scenic/issues/256

### Checklist

### DIFF
--- a/src/scenic/domains/driving/roads.py
+++ b/src/scenic/domains/driving/roads.py
@@ -1211,7 +1211,7 @@ class Network:
                 message = reject
             else:
                 message = "requested element does not exist"
-            rejectSample(message)
+            raise RejectionException(message)
         return None
 
     def _findPointInAll(self, point, things, key=lambda e: e):
--- a/src/scenic/domains/driving/roads.py
+++ b/src/scenic/domains/driving/roads.py
@@ -34,7 +34,6 @@ from scenic.core.distributions import (
     distributionFunction,
     distributionMethod,
 )
-from scenic.core.errors import InvalidScenarioError
 import scenic.core.geometry as geometry
 from scenic.core.object_types import Point
 from scenic.core.regions import PolygonalRegion, PolylineRegion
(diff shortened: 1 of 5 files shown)


abanuelo (approved):
Looks great!

src/scenic/syntax/veneer.py
```diff
@@ -331,6 +332,25 @@ def isActive():
     return activity > 0
 
 
+def allowSampleRejection():
+    """Should a RejectionException() be allowed
+
+    Returning True means a RejectionException is valid. Returning False indicates
+    a RejectionException() means the scenario is invalid.
+    """
+    return isActive() or heuristicSampling
```
dfremont, line 341:
```suggestion
    return not isActive() or heuristicSampling
```

src/scenic/domains/driving/roads.py
```diff
@@ -57,7 +57,7 @@ def _toVector(thing: Vectorlike) -> Vector:
 
 
 def _rejectSample(message):
-    if veneer.isActive():
+    if not veneer.allowSampleRejection():
```
dfremont, line 60: Since `allowSampleRejection` should only ever be used in this way, I'd suggest replacing it with a function `veneer.rejectSample` which copies the logic of this function (raising either `InvalidScenarioError` or `RejectionException` as appopriate). Then the roads library and other libraries can call `veneer.rejectSample` when they detect a requirement violation and not worry about how to handle it.

dfremont (changes requested):
I think the sense of the main test is backwards (rejections should ordinarily *not* be allowed at compile time), so we'd better add a test making sure it works correctly. I have a refactoring suggestion above too.

src/scenic/domains/driving/roads.py
Eric-Vin: Good idea, I'll make the swap!

dfremont (approved):
Looks good! If you want to fix the issue of having to set a fixed color in the test by changing `DrivingObject.lane` etc. to use `self.position` instead of `self` in this PR, go ahead; otherwise we can make a new one.

Eric-Vin:
Went ahead and made the change. Please go ahead and take a final look and merge when you get the chance!

PR #290, opened 2024-07-11
Minor Tweaks to PolygonalRegion

Eric-Vin:
### Description
Minor tweaks to PolygonalRegion's `__eq__` and `hash` functions.

### Issue Link
N/A

### Checklist

### Additional Notes
N/A

### DIFF
--- a/src/scenic/core/regions.py
+++ b/src/scenic/core/regions.py
@@ -2774,7 +2774,6 @@ class PolygonalRegion(Region):
             points = tuple(pt[:2] for pt in points)
             if len(points) == 0:
                 raise ValueError("tried to create PolygonalRegion from empty point list!")
-            self.points = points
             polygon = shapely.geometry.Polygon(points)
 
         if isinstance(polygon, shapely.geometry.Polygon):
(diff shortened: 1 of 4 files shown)


src/scenic/core/regions.py
```diff
@@ -3049,7 +3049,14 @@ def __eq__(self, other):
 
     @cached
     def __hash__(self):
-        return hash((self.polygons, self.orientation, self.z))
+        return hash(
+            (
+                tuple(self._points) if self._points else None,
```
dfremont, line 3054: This seems like an odd place to normalize the type. We don't want someone to pass a list of random points to `PolygonalRegion`, mutate that list, and have the mutation end up affecting the resulting sampled region. Probably better to convert to `tuple` in the initializer and use plain `self._points` here.

dfremont (changes requested):
Looks good, just a minor comment above.

src/scenic/core/regions.py
Eric-Vin: Good call, I'll make the change.

dfremont (approved):
Looks great!

PR #293, opened 2024-07-16
fix: shut down instance after workflow finishes

abanuelo:
### Description

As part of enhancing the simulator CI, we are introducing a condition for the `stop_ec2_instance` step that will get triggered regardless of success or failures of previous steps in the run to ensure that the instance is always stopped after a run. 

### Checklist
- [ ] I have tested the changes locally via `pytest` and/or other means


Eric-Vin (approved):

dfremont:
Superseded by #310.

PR #297, opened 2024-08-05
feat: MetaDrive Interface

abanuelo:
### Description

These changes introduce the metadrive x scenic simulator interface.  At the moment, this interface only supports `Vehicle` objects but will be extended to introduce `Pedestrians` and other obstacle objects. Below is an example of the MetaDrive simulator interface:

https://github.com/user-attachments/assets/d442b4ce-8d56-410d-97d6-dd58290f813f

### Checklist


dfremont:
Finished in #326

Issue #294, opened 2024-07-29
External sampler does not work with object positioning


amolenaar:
### System Details

1.  Python Version: Python 3.12.4
1.  Scenic Version: Scenic 3.0.0
1.  Operating System / Platform: Apple M2 Air macOS Sonoma 14.5

### Detailed Description

For our use case (railway) we want to make sure we create samples in different regions: plains, tunnel, forest area.

To make this work we created an external sampler. Although we can use the external sampler to define values for arbitrary parameters (defined with the `with` keyword), we cannot assign a region (`on`/`in` keyword) from our distribution.

It would be really nice if we can use external parameters to define regions.

For comparison, this is a working scenario:

```scenic
from scenic_external_sampler.regions import NorthEast, SouthWest

ego = new Object in Uniform(NorthEast, SouthWest)
```

While this scenario fails:

```scenic
from scenic_external_sampler.regions import NorthEast, SouthWest
from scenic_external_sampler.sampler import SimpleSampler, Values
param externalSampler = SimpleSampler

ego = new Object in Values(NorthEast, SouthWest)
```

### Steps To Reproduce

Have a look at a sample repository I created: https://github.com/amolenaar/scenic-external-sampler

You can easily set up an environment with Poetry:

```bash
poetry install
poetry run pytest
```

All tests succeed, except [test_regions_with_externally_sampled_distribution](https://github.com/amolenaar/scenic-external-sampler/blob/main/tests/test_scenario.py#L66). This test tries to set a region from an external parameter. It fails, since the external sampler has not been initialized yet.

It shows a stack trace:

```
.venv/lib/python3.12/site-packages/scenic/syntax/translator.py:105: in scenarioFromString
    return _scenarioFromStream(stream, options, filename, scenario=scenario, **kwargs)
.venv/lib/python3.12/site-packages/scenic/syntax/translator.py:172: in _scenarioFromStream
    compileStream(stream, namespace, compileOptions, filename)
.venv/lib/python3.12/site-packages/scenic/syntax/translator.py:321: in compileStream
    executeCodeIn(code, namespace)
.venv/lib/python3.12/site-packages/scenic/syntax/translator.py:560: in executeCodeIn
    exec(code, namespace)
<string>:1: in <module>
    ???
.venv/lib/python3.12/site-packages/scenic/syntax/veneer.py:1411: in In
    if alwaysProvidesOrientation(region):
[...]

PR #300, opened 2024-08-15
alwaysProvidesOrientation Patch and Test

Eric-Vin:
### Description
Fix for alwaysProvidesOrientation

### Issue Link
https://github.com/BerkeleyLearnVerify/Scenic/issues/294

### Checklist

### Additional Notes
N/A

### DIFF
--- a/src/scenic/syntax/veneer.py
+++ b/src/scenic/syntax/veneer.py
@@ -249,6 +249,7 @@ from pathlib import Path
 import sys
 import traceback
 import typing
+import warnings
 
 from scenic.core.distributions import (
     Distribution,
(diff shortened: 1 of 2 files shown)


dfremont (approved):
Looks good for now -- let's still plan to remove the sampling heuristic entirely.

amolenaar:
@dfremont @Eric-Vin Thanks for the workaround. I can achieve the same by throwing a `RejectionException` from the external parameter if no sampler is present.

Does this have effects down the line, if it cannot determine if the sample has an orientation (that's why it does the sampling IIRC)?

Issue #304, opened 2024-09-19
Cars in NewtonianSimulator do not always get rendered on correct coordinates


JACY24:
### System Details

1. Pytthon 3.10.8
2. Scenic 3.0.0
3. Microsoft Windows Version 23H2
4. Newtonian simulator (the built in version)

### Detailed Description

1. When running a simulation with the newtonian simulator, the cars seem to not always get rendered on the location that they say they are. In the simulator, there is an if-else statement that draws a car if an object is a car and a polygon otherwise. I tried removing this statement and just drawing both, and the two do not overlap in the rendered simulation, see the screenshots.
2. I ran `scenic examples/driving/badlyParkedCarPullingIn.scenic --2d --simulate --model scenic.simulators.newtonian.driving_model --time 200`
3. ![scenario](https://github.com/user-attachments/assets/7a713090-028b-43e9-8e08-7af10055762e)
![scenario_1](https://github.com/user-attachments/assets/2d805721-447e-43b2-a35a-289163f0ae68)

### Steps To Reproduce

1. Open `Scenic\src\scenic\simulators\newtonian\simulator.py`
2. Go to line 217
3. Replace the if-statement with:
```
            # if hasattr(obj, "isCar") and obj.isCar:
            self.rotated_car = pygame.transform.rotate(
                self.car, math.degrees(obj.heading)
            )
            self.screen.blit(self.rotated_car, (rect_x, rect_y))
            # else:
            corners = [self.scenicToScreenVal(corner) for corner in obj._corners2D]
            pygame.draw.polygon(self.screen, color, corners)
```
The only thing this should do is draw both a Polygon and a car, the logic of the location should not be changed.
4. In the terminal, run ```scenic examples/driving/badlyParkedCarPullingIn.scenic --2d --simulate --model scenic.simulators.newtonian.driving_model --time 200```
6. See the resulting error

### Issue Submission Checklist

- [X] I am reporting an issue, not asking a question
- [X] I checked the open and closed issues, forum, etc. and have not found any solution
- [X] I have provided all necessary code, etc. to reproduce the issue

Eric-Vin:
Apologies for the late response! This definitely looks incorrect: I'll figure out what's wrong and make a patch.

PR #309, opened 2024-10-25
Newtonian Simulator Fixes and Cleanup

Eric-Vin:
### Description
This PR fixes several issues with the Newtonian simulator:
- Previously scaling was done per dimension, warping the objects.
- Objects were being drawn to a position that did not account for their rotation, leading to incorrect images and wobbling.
- Car dimensions are no longer hard-coded and are now drawn from the object's width and height on a per-object basis.
- Added a debug global parameter `debugRender` that always draws polygons to help verify that cars (or any future objects) are being drawn correctly.
- General code cleanup, removing several unused lines.

### Issue Link
This will resolve Issue #304.

### Checklist

### Additional Notes
To verify that  this resolves Issue #304, run an example Scenic program with the `debugRenderer` parameter set to `True`. An example is included below:
`scenic examples/driving/badlyParkedCarPullingIn.scenic --2d --simulate --model scenic.simulators.newtonian.driving_model --time 200 --count 1 -p debugRender True`

I've also included several screenshots showing the correct behavior which is now present:
Screenshot from 2024-10-25 11-15-48
Screenshot from 2024-10-25 11-15-45


src/scenic/simulators/newtonian/simulator.py
```diff
@@ -102,10 +107,27 @@ def setup(self):
...
             self.size_x = self.max_x - self.min_x
             self.size_y = self.max_y - self.min_y
+
+            # Generate a uniform screen scaling (applied to width and height)
+            # that includes all of both dimensions.
+            self.screenScaling = min(WIDTH / self.size_x, HEIGHT / self.size_y)
+
+            # Calculate a screen translation that brings the mean vehicle
+            # position to the center of the screen.
+            self.screenTranslation = (0, 0)
```
dfremont, line 121: This line is a bit confusing since the attribute is overwritten below.

src/scenic/simulators/newtonian/simulator.py
```diff
@@ -149,9 +169,14 @@ def addRegion(region, color, width=1):
...
         x, y = pos[:2]
-        x_prop = (x - self.min_x) / self.size_x
-        y_prop = (y - self.min_y) / self.size_y
-        return int(x_prop * WIDTH), HEIGHT - 1 - int(y_prop * HEIGHT)
+
+        screen_x = int((x - self.min_x) * self.screenScaling)
+        screen_y = HEIGHT - 1 - int((y - self.min_y) * self.screenScaling)
+
+        screen_x = screen_x + self.screenTranslation[0]
+        screen_y = screen_y + self.screenTranslation[1]
```
dfremont, line 177: These will be floats in general, not ints, but I guess that doesn't break pygame?

dfremont (changes requested):
Looks great, just one tiny fix and a question.

src/scenic/simulators/newtonian/simulator.py
Eric-Vin: Hmm, do you have a suggestion to clarify? I wrote it this way since the best way to compute this seemed to be to just see what the Scenic to screen mapping computed with no offset and then center everything.

src/scenic/simulators/newtonian/simulator.py
Eric-Vin: Ah good catch, I'll cast them.

src/scenic/simulators/newtonian/simulator.py
Eric-Vin: Added a comment to clarify.

dfremont (approved):

PR #318, opened 2024-12-09
Adding support for CARLA 0.10

glopezdiest:
### Description

This PR changes the CARLA example scenarios and the simulator scripts to add compatibility with 0.10. The changes are as follows:

- Modified the scenarios to use Town10HD_Opt, the only town in 0.10
- Modified the available blueprints
- Removed a camera transform when rendering as it is no longer needed
- Changed the controllers to match the change from PhysX to Chaos

### Checklist

### Additional Notes

This breaks compatibility with previous CARLA releases (specifically because of the blueprint changes), so I'm not sure what would be the best approach to follow, and some feedback would be appreciated


glopezdiest:
Actually, closing this due to the compatibility issues

PR #326, opened 2025-02-05
Feat: Initial MetaDrive Integration

lola831:
### Description
These changes introduce the **MetaDrive x Scenic simulator interface**, allowing **MetaDrive to be used within Scenic.**

**Key Features:**

- Supports Vehicle and Pedestrian objects within MetaDrive.

- Adds MetaDrive as a required dependency.

**Known Issues (Pending Fixes in Next MetaDrive Release):**

- Rendering in 3D does not work on Mac M1 devices.

- Vehicles do not fully stop when applying full brakes.

These issues are **already identified** and will be fixed in the next MetaDrive release.

**Example:**
Below is an example of the MetaDrive simulator interface rendering in both **2D and 3D:**

https://github.com/user-attachments/assets/c3ff5c30-65d2-4df6-9101-2bbfe67e5392

https://github.com/user-attachments/assets/33e86d67-8439-4066-895b-27d57a42826b

### Checklist


docs/simulators.rst
```diff
@@ -16,6 +16,36 @@ See the individual entries for details on each interface's capabilities and how
...
+MetaDrive (Optional)
+----------------------------
+
+Scenic supports integration with the `MetaDrive <https://metadriverse.github.io/metadrive/>`_ simulator as an optional dependency,
+enabling users to describe dynamic simulations of vehicles, pedestrians, and traffic scenarios.
+If your system supports it, you can install it with:
+
+.. code-block:: console
+
+    pip install scenic[metadrive]
```
dfremont, line 28:
```suggestion
    python -m pip install scenic[metadrive]
```
This is what we do in the Scenic installation instructions -- it's a little safer since sometimes virtual environments don't capture the `pip` command but only `python` (for reasons I don't understand).

docs/simulators.rst
```diff
@@ -16,6 +16,36 @@ See the individual entries for details on each interface's capabilities and how
 Currently Supported
 ===================
 
+MetaDrive (Optional)
```
dfremont, line 19: The simulators are almost all optional, so I don't think we need to state that explicitly here.
```suggestion
MetaDrive
```

docs/simulators.rst
```diff
@@ -16,6 +16,36 @@ See the individual entries for details on each interface's capabilities and how
...
+These issues are expected to be addressed in the next version of MetaDrive.
+
+Scenic uses OpenDRIVE maps, while MetaDrive relies on SUMO maps. Scenic provides corresponding SUMO maps for OpenDRIVE maps under the :file:`assets/maps/CARLA` directory.
+Additionally, you can convert your own OpenDRIVE maps to SUMO maps using the `netconvert <https://sumo.dlr.de/docs/Networks/Import/OpenDRIVE.html/>`_ tool.
+To avoid setting the SUMO map manually, name it the same as your OpenDRIVE file and place it in the same directory.
+Otherwise, specify it explicitly with:
+
+.. code-block:: console
+
+    --param sumo_map your_map.net.xml
```
dfremont, line 43: People don't necessarily need to do this on the command line: they could also set the parameter in their Scenic script. I'd just say something like "Otherwise, you can specify it explicitly using the `sumo_map` global parameter."

docs/tutorials/dynamics.rst
```diff
@@ -396,10 +396,19 @@ You can see all of the above syntax in action by running some of our examples of
 scenarios. We have examples written for the CARLA and LGSVL driving simulators, and those
 in :file:`examples/driving` in particular are designed to use Scenic's abstract
 :ref:`driving domain <driving_domain>` and so work in either of these simulators, as well
-as Scenic's built-in Newtonian physics simulator. The Newtonian simulator is convenient
-for testing and simple experiments; you can find details on how to install the more
-realistic simulators in our :ref:`simulators` page (they should work on both Linux and
-Windows, but not macOS, at the moment).
+as Scenic's built-in Newtonian physics simulator and MetaDrive simulator. While the Newtonian simulator is convenient
```
dfremont, line 399:
```suggestion
as Scenic's built-in Newtonian physics simulator and the MetaDrive simulator. While the Newtonian simulator is convenient
```

docs/tutorials/dynamics.rst
```diff
@@ -396,10 +396,19 @@ You can see all of the above syntax in action by running some of our examples of
...
+for testing simple experiments, we recommend using MetaDrive for more realistic driving scenarios.
+
+MetaDrive support is **optional**. If your system supports MetaDrive, you can install it separately using:
+
+.. code-block:: console
+
+    pip install scenic[metadrive]
```
dfremont, line 406:
```suggestion
    python -m pip install scenic[metadrive]
```

docs/tutorials/dynamics.rst
```diff
@@ -414,16 +423,16 @@ usual schematic diagram of the generated scenes:
 To run dynamic simulations, add the :option:`--simulate` option (:option:`-S` for short).
 Since this scenario is not written for a particular simulator, you'll need to specify
 which one you want by using the :option:`--model` option (:option:`-m` for short) to
-select the corresponding Scenic :term:`world model`: for example, to use the Newtonian simulator we could add
-``--model scenic.simulators.newtonian.driving_model``. It's also a good idea to put a time bound on
-the simulations, which we can do using the :option:`--time` option.
+select the corresponding Scenic :term:`world model`: for example, to use the Metadrive simulator we could add
```
dfremont, line 426:
```suggestion
select the corresponding Scenic :term:`world model`: for example, to use the MetaDrive simulator we could add
```

examples/driving/OAS_Scenarios/oas_scenario_03.scenic
```diff
@@ -2,6 +2,9 @@
 Voyage OAS Scenario Unique ID: 2-2-XX-CF-STR-CAR
 The ego vehicle follows the lead car
 
+To run this file using the MetaDrive simulator:
+    scenic examples/driving/Carla_Challenge/carlaChallenge3.scenic --2d --model scenic.simulators.metadrive.model --simulate
```
dfremont, line 6:
```suggestion
    scenic examples/driving/OAS_Scenarios/oas_scenario_03.scenic --2d --model scenic.simulators.metadrive.model --simulate
```
and similar for the other files.

src/scenic/simulators/metadrive/model.scenic
```diff
@@ -0,0 +1,152 @@
...
+Vehicles and pedestrians support the basic actions and behaviors from the driving domain.
+
+The model defines several global parameters, whose default values can be overridden
+in scenarios using the ``param`` statement or on the command line using the
+:option:`--param` option:
+
+Global Parameters:
+    sumo_map (str or Path): Path to the SUMO map (``.net.xml`` file) to use in the simulation.
+        This map should correspond to the **map** file used in the scenario, which must be in the OpenDrive
+        format (``.xodr`` file).
```
dfremont, line 14: We might add other formats later (we have a prototype branch for Lanelet2 maps somewhere), so I'd just point to the driving domain documentation for `map` here (in `scenic.domains.driving.model`).

src/scenic/simulators/metadrive/model.scenic
```diff
@@ -0,0 +1,152 @@
...
+    timestep (float): The interval (in seconds) between each simulation step. This determines how often Scenic
+        interrupts MetaDrive to run behaviors, check requirements, and update the simulation state.
+        The default value is 0.1 seconds.
+    render (bool): Whether to render the simulation screen. If True (default), it will open a screen and render
+        the simulation. If False, rendering is disabled.
+    render3D (bool): Whether to render the simulation in 3D. If True, it will render the simulation in 3D.
+        If False (default), it will render in 2D.
+    real_time (bool): If True (default), the simulation will run in real time, ensuring each step takes at least
+        as long as the specified timestep. If False, the simulation may run faster, based on the time it takes
+        to process each step.
```
dfremont, line 24:
```suggestion
        as long as the specified timestep. If False, the simulation runs as fast as possible.
```

src/scenic/simulators/metadrive/model.scenic
```diff
@@ -0,0 +1,152 @@
...
+from scenic.core.errors import InvalidScenarioError
+
+try:
+    from scenic.simulators.metadrive.simulator import MetaDriveSimulator
+    from scenic.simulators.metadrive.utils import scenicToMetaDriveHeading
+except ModuleNotFoundError:
+    # for convenience when testing without the metadrive package
+    from scenic.core.simulators import SimulatorInterfaceWarning
+    import warnings
+    warnings.warn('The "metadrive" package is not installed; '
```
dfremont, line 41: The package is called `metadrive-simulator`, right?

src/scenic/simulators/metadrive/model.scenic
```diff
@@ -0,0 +1,152 @@
...
+
+    def MetaDriveSimulator(*args, **kwargs):
+        """Dummy simulator to allow compilation without the 'metadrive' package.
+
+        :meta private:
+        """
+        raise RuntimeError('the "metadrive" package is required to run simulations '
+                           'from this scenario')
+
+if globalParameters.get("sumo_map") is None:
```
dfremont, line 53:
```suggestion
if "sumo_map" not in globalParameters:
```

docs/simulators.rst
lola831: Applied suggestion

docs/tutorials/dynamics.rst
lola831: Applied suggestion

docs/tutorials/dynamics.rst
lola831: Applied suggestion

examples/driving/OAS_Scenarios/oas_scenario_03.scenic
lola831: Applied suggestion

src/scenic/simulators/metadrive/model.scenic
lola831: Applied suggestion

src/scenic/simulators/metadrive/model.scenic
lola831: Applied suggestion

docs/tutorials/dynamics.rst
lola831: Applied suggestion

src/scenic/simulators/metadrive/model.scenic
lola831: Applied suggestion

src/scenic/simulators/metadrive/model.scenic
```diff
@@ -0,0 +1,151 @@
...
+class Vehicle(Vehicle, Steers, MetaDriveActor):
+    """Abstract class for steerable vehicles."""
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self._control = {"steering": 0, "throttle": 0, "brake": 0}
+
+    @property
+    def control(self):
+        """Returns the current accumulated control inputs (throttle, brake, and steering)."""
```
dfremont, line 105: Is there a reason to expose this publicly? I would have thought this was an implementation detail. (Likewise for `resetControl` below.)

src/scenic/simulators/metadrive/model.scenic
```diff
@@ -0,0 +1,151 @@
...
+class Pedestrian(Pedestrian, MetaDriveActor, Walks):
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+
+    @property
+    def isPedestrian(self):
+        return True
+
+    def setWalkingDirection(self, heading):
+        converted_heading = scenicToMetaDriveHeading(self.heading)
```
dfremont, line 144:
```suggestion
        converted_heading = scenicToMetaDriveHeading(heading)
```
Better double-check that `SetWalkingDirectionAction` works properly!

src/scenic/simulators/metadrive/simulator.py
```diff
@@ -0,0 +1,321 @@
...
+        self.real_time = real_time
+        (
+            self.center_x,
+            self.center_y,
+            self.offset_x,
+            self.offset_y,
+            self.sumo_map_boundary,
+        ) = self._get_map_parameters()
+
+    def _get_map_parameters(self):
```
dfremont, line 51: Might this method be used somewhere else? If not, I'd simplify the code by inlining it above (i.e. have `__init__` call `utils.getMapParameters` directly).

src/scenic/simulators/metadrive/simulator.py
```diff
@@ -0,0 +1,321 @@
...
+        """Helper method to extract map parameters."""
+        # Use utility function to get the map parameters
+        return utils.getMapParameters(self.sumo_map)
+
+    def createSimulation(self, scene, *, timestep, **kwargs):
+        if timestep is not None and timestep != self.timestep:
+            raise RuntimeError(
+                "cannot customize timestep for individual MetaDrive simulations; "
+                "set timestep when creating the MetaDriveSimulator instead"
+            )
```
dfremont, line 61: I don't see the need for this error: it looks like `timestep` is only used within `MetaDriveSimulation`, so you might as well delete this code and allow `timestep` to be customized per-simulation.

src/scenic/simulators/metadrive/simulator.py
```diff
@@ -0,0 +1,321 @@
...
+            real_time=self.real_time,
+            center_x=self.center_x,
+            center_y=self.center_y,
+            offset_x=self.offset_x,
+            offset_y=self.offset_y,
+            sumo_map_boundary=self.sumo_map_boundary,
+            **kwargs,
+        )
+
+    def destroy(self):
```
dfremont, line 79: No need to override this method if it just calls the superclass implementation.

src/scenic/simulators/metadrive/simulator.py
```diff
@@ -0,0 +1,321 @@
...
+        render,
+        render3D,
+        scenario_number,
+        timestep,
+        sumo_map,
+        real_time,
+        center_x,
+        center_y,
+        offset_x,
+        offset_y,
```
dfremont, line 96: As far as I can tell these 4 parameters are always used together, so I think it would be cleaner to pass them all together as a single tuple (e.g. `sumo_center_offset` or something).

src/scenic/simulators/metadrive/simulator.py
```diff
@@ -0,0 +1,321 @@
...
+                raise SimulationCreationError(
+                    "Metadrive requires you to define at least one Scenic object."
+                )
+            if not scene.objects[0].isCar:
+                raise SimulationCreationError(
+                    "The first object must be a car to serve as the ego vehicle in Metadrive."
+                )
+        except SimulationCreationError as e:
+            print(e)
+            sys.exit(1)
```
dfremont, line 113: Was this debugging code left in by accident? If it's intentional, I think raising and then immediately catching an exception is confusing: I'd just do what you want directly. But we shouldn't just call `sys.exit` in case the user wants to catch this and handle it at a higher level. If `SimulationCreationError` isn't a good exception here since the bug in the user's scenario (no ego) is guaranteed to cause a failure every time, you could raise `InvalidScenarioError`.

src/scenic/simulators/metadrive/simulator.py
```diff
@@ -0,0 +1,321 @@
...
+        self.real_time = real_time
+        self.center_x = center_x
+        self.center_y = center_y
+        self.offset_x = offset_x
+        self.offset_y = offset_y
+        self.sumo_map_boundary = sumo_map_boundary
+        super().__init__(scene, timestep=timestep, **kwargs)
+
+    def setup(self):
+        super().setup()
```
dfremont, line 131: Unnecessary.

src/scenic/simulators/metadrive/simulator.py
```diff
@@ -0,0 +1,321 @@
...
+        if not self.defined_ego:
+            decision_repeat = math.ceil(self.timestep / 0.02)
+            physics_world_step_size = self.timestep / decision_repeat
+
+            # Initialize the simulator with ego vehicle
+            self.client = utils.DriveEnv(
+                dict(
+                    decision_repeat=decision_repeat,
+                    physics_world_step_size=physics_world_step_size,
+                    use_render=self.render if self.render3D else False,
```
dfremont, line 155: This logic confuses me. Didn't you set `render3D` to be false when `render` is false above? Then I think this is equivalent to just `self.render3D` (although I would have thought it should be `self.render`).

src/scenic/simulators/metadrive/simulator.py
```diff
@@ -0,0 +1,321 @@
...
+            self.client.config["sumo_map"] = self.sumo_map
+            self.client.reset()
+
+            # Assign the MetaDrive actor to the ego
+            metadrive_objects = self.client.engine.get_objects()
+            if metadrive_objects:
+                ego_metaDriveActor = list(metadrive_objects.values())[0]
+                obj.metaDriveActor = ego_metaDriveActor
+                self.defined_ego = True
+                return obj.metaDriveActor
```
dfremont, line 175: This method doesn't need to return anything. (Looks like a bug in the CARLA interface too!)

src/scenic/simulators/metadrive/simulator.py
```diff
@@ -0,0 +1,321 @@
...
+            else:
+                raise SimulationCreationError(
+                    f"Unable to initialize MetaDrive ego vehicle for object {obj}"
+                )
```
dfremont, line 179: I'm assuming based on this code that it's possible for `engine.get_objects()` to fail? What about `engine.agent_manager.spawn_object()` below?

src/scenic/simulators/metadrive/simulator.py
```diff
@@ -0,0 +1,321 @@
...
+            f"Unsupported object type: {type(obj)} for object {obj}."
+        )
+
+    def executeActions(self, allActions):
+        """Execute actions for all vehicles in the simulation."""
+        super().executeActions(allActions)
+
+        # Apply control updates to vehicles
+        for idx, obj in enumerate(self.scene.objects):
+            if obj.isCar and idx != 0:  # Skip ego vehicle (it is handled separately)
```
dfremont, line 213:
```suggestion
        for obj in self.scene.objects[1:]:  # Skip ego vehicle (it is handled separately)
            if obj.isCar:
```

src/scenic/simulators/metadrive/simulator.py
```diff
@@ -0,0 +1,321 @@
...
+                action = obj.collectAction()
+                obj.metaDriveActor.before_step(action)
+                obj.resetControl()
+
+    def step(self):
+        start_time = time.monotonic()
+
+        # Special handling for the ego vehicle
+        ego_obj = self.scene.objects[0]
+        if ego_obj.isCar:
```
dfremont, line 223: I think you checked that the ego was a car during initialization above.

src/scenic/simulators/metadrive/simulator.py
```diff
@@ -0,0 +1,321 @@
...
+            action = ego_obj.collectAction()
+            o, r, tm, tc, info = self.client.step(action)  # Apply action in the simulator
```
dfremont, line 225: If none of the return values are used, I would not bother assigning them to variables.

src/scenic/simulators/metadrive/simulator.py
```diff
@@ -0,0 +1,321 @@
...
+            ego_obj.resetControl()
+
+        # Render the scene in 2D if needed
+        if self.render and not self.render3D:
+            film_size = utils.calculateFilmSize(self.sumo_map_boundary, scaling=5)
```
dfremont, line 230: Let's just compute this once at initialization and reuse it.

src/scenic/simulators/metadrive/simulator.py
```diff
@@ -0,0 +1,321 @@
...
+
+        return values
+
+    def getLaneFollowingControllers(self, agent):
+        dt = self.timestep
+        if agent.isCar:
+            lon_controller = PIDLongitudinalController(K_P=0.5, K_D=0.1, K_I=0.7, dt=dt)
+            lat_controller = PIDLateralController(K_P=0.13, K_D=0.3, K_I=0.05, dt=dt)
+            # Good for Town 07
+            # lat_controller = PIDLateralController(K_P=0.15, K_D=0.69, K_I=0.03, dt=dt)
```
dfremont, line 291: Left in by accident?

src/scenic/simulators/metadrive/utils.py
```diff
@@ -0,0 +1,113 @@
+# NOTE: MetaDrive uses a coordinate system where (0,0) is centered
+# around the middle of the SUMO map. To ensure alignment, we shift
+# positions using both the computed SUMO map center (center_x, center_y)
+# and adjust for SUMO’s netOffset (offset_x, offset_y).
+
+try:
+    from metadrive.envs import BaseEnv
+    from metadrive.manager.sumo_map_manager import SumoMapManager
+    from metadrive.obs.observation_base import DummyObservation
+except ImportError as e:
```
dfremont, line 10: I think this check is redundant since you already do it in `simulator.py`. I'd rather just have a single place for the error to be raised.

src/scenic/simulators/metadrive/utils.py
```diff
@@ -0,0 +1,113 @@
...
+    location_tag = root.find("location")
+    net_offset = tuple(map(float, location_tag.attrib["netOffset"].split(",")))
+    sumo_map_boundary = tuple(map(float, location_tag.attrib["convBoundary"].split(",")))
+    return net_offset, sumo_map_boundary
+
+
+def getMapParameters(sumo_map_path):
+    """Retrieve the map parameters."""
+    net_offset, sumo_map_boundary = extractNetOffsetAndBoundary(sumo_map_path)
+    if net_offset and sumo_map_boundary:
```
dfremont, line 47: It's not clear to me how this condition can be false. If the `netOffset` or `convBoundary` attributes are missing, the `extractNetOffsetAndBoundary` function will fail, and otherwise the `net_offset` and `sumo_map_boundary` tuples will be nonempty, right?

src/scenic/simulators/metadrive/utils.py
```diff
@@ -0,0 +1,113 @@
...
+        offset_y = net_offset[1]
+        return center_x, center_y, offset_x, offset_y, sumo_map_boundary
+    else:
+        raise RuntimeError("Failed to extract netOffset or convBoundary from SUMO map.")
+
+
+def metadriveToScenicPosition(loc, center_x, center_y, offset_x, offset_y):
+    """Converts MetaDrive position to Scenic position using map parameters."""
+    x_scenic = loc[0] + center_x - offset_x
+    y_scenic = loc[1] + center_y - offset_y
```
dfremont, line 61: If the `center` and `offset` numbers are always used together like this, maybe we should just compute combined `offset_x` and `offset_y` values so we can just write `loc[0] + offset_x` here etc.?

tests/simulators/metadrive/basic.scenic
```diff
@@ -0,0 +1,12 @@
+param map = localPath('../../../assets/maps/CARLA/Town01.xodr')
+param sumo_map = localPath('../../../assets/maps/CARLA/Town01.net.xml')
+from scenic.simulators.metadrive.model import *
```
dfremont, line 3:
```suggestion
model scenic.simulators.metadrive.model
```

tests/simulators/metadrive/test_metadrive.py
```diff
@@ -0,0 +1,131 @@
...
+        terminate after 5 steps
+    """
+    scenario = compileScenic(code, mode2D=True)
+    scene = sampleScene(scenario)
+    simulation = simulator.simulate(scene)
+    records = simulation.result.records["CarSpeed"]
+    assert records[len(records) // 2][1] < records[-1][1]
+
+
+@pytest.mark.skip(
```
dfremont, line 63: Let's do `xfail` instead of `skip` so that when the MetaDrive problem is fixed the test will pass, `pytest` will highlight it, and we can remove the `xfail`.

tests/simulators/metadrive/test_metadrive.py
```diff
@@ -0,0 +1,131 @@
...
+    code = f"""
+        param map = r'{openDrivePath}'
+        param sumo_map = r'{sumoPath}'
+
+        model scenic.simulators.metadrive.model
+
+        behavior WalkForward():
+            while True:
+                take SetWalkingDirectionAction(self.heading)
+                take SetWalkingSpeedAction(0.5)
```
dfremont, line 109:
```suggestion
                take SetWalkingDirectionAction(self.heading), SetWalkingSpeedAction(0.5)
```
This may break, if the actions interfere with each other as we discussed, but it's supposed to be possible to use these actions simultaneously (see the `WalkForwardBehavior` in CARLA, for example).

dfremont (changes requested):
Looks good overall! Various comments above, and I'm going to do a bit more testing on my end while you're working on those.

src/scenic/simulators/metadrive/simulator.py
lola831: I'll update the code to use `use_render=self.render3D`. In MetaDrive, `use_render` controls 3D rendering.

src/scenic/simulators/metadrive/model.scenic
lola831: Fixed as suggested in commit 89878f32b7a4dcb66c54d287524193c0d030c1be

src/scenic/simulators/metadrive/simulator.py
lola831: Fixed as suggested in commit 89878f32b7a4dcb66c54d287524193c0d030c1be

src/scenic/simulators/metadrive/simulator.py
lola831: Fixed as suggested in commit 89878f32b7a4dcb66c54d287524193c0d030c1be

src/scenic/simulators/metadrive/simulator.py
lola831: Fixed as suggested in commit 89878f32b7a4dcb66c54d287524193c0d030c1be

src/scenic/simulators/metadrive/simulator.py
lola831: Fixed as suggested in commit 89878f32b7a4dcb66c54d287524193c0d030c1be

src/scenic/simulators/metadrive/simulator.py
lola831: Fixed as suggested in commit 89878f32b7a4dcb66c54d287524193c0d030c1be

src/scenic/simulators/metadrive/simulator.py
lola831: Fixed as suggested in commit 89878f32b7a4dcb66c54d287524193c0d030c1be

src/scenic/simulators/metadrive/simulator.py
lola831: Fixed in commit 89878f32b7a4dcb66c54d287524193c0d030c1be

src/scenic/simulators/metadrive/simulator.py
lola831: Fixed as suggested in commit 89878f32b7a4dcb66c54d287524193c0d030c1be

src/scenic/simulators/metadrive/simulator.py
lola831: Fixed as suggested in commit 89878f32b7a4dcb66c54d287524193c0d030c1be

src/scenic/simulators/metadrive/simulator.py
lola831: Fixed as suggested in commit 89878f32b7a4dcb66c54d287524193c0d030c1be

src/scenic/simulators/metadrive/simulator.py
lola831: Fixed as suggested in commit 89878f32b7a4dcb66c54d287524193c0d030c1be

src/scenic/simulators/metadrive/simulator.py
lola831: Fixed as suggested in commit 89878f32b7a4dcb66c54d287524193c0d030c1be

src/scenic/simulators/metadrive/simulator.py
```diff
@@ -0,0 +1,321 @@
...
+            self.center_y,
+            self.offset_x,
+            self.offset_y,
+        )
+        velocity = Vector(*metaDriveActor.velocity, 0)
+        speed = metaDriveActor.speed
+        angularSpeed = 0
+        angularVelocity = utils.metadriveToScenicPosition(
+            (0, 0), self.center_x, self.center_y, self.offset_x, self.offset_y
+        )
```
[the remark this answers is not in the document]
lola831: Fixed as suggested in commit 89878f32b7a4dcb66c54d287524193c0d030c1be

src/scenic/simulators/metadrive/utils.py
lola831: Check removed in commit 89878f32b7a4dcb66c54d287524193c0d030c1be

src/scenic/simulators/metadrive/utils.py
lola831: You're correct. Removed in commit 89878f32b7a4dcb66c54d287524193c0d030c1be

src/scenic/simulators/metadrive/utils.py
lola831: Fixed in commit 89878f32b7a4dcb66c54d287524193c0d030c1be

tests/simulators/metadrive/basic.scenic
lola831: Fixed as suggested in commit 89878f32b7a4dcb66c54d287524193c0d030c1be

tests/simulators/metadrive/test_metadrive.py
lola831: Fixed in commit 89878f32b7a4dcb66c54d287524193c0d030c1be

tests/simulators/metadrive/test_metadrive.py
lola831: Updated in commit 89878f32b7a4dcb66c54d287524193c0d030c1be

tests/simulators/metadrive/test_metadrive.py
```diff
@@ -0,0 +1,131 @@
...
+                take SetThrottleAction(1)
+
+        ego = new Car at (369, -326), with behavior DriveWithThrottle
+        record ego.speed as CarSpeed
+        terminate after 5 steps
+    """
+    scenario = compileScenic(code, mode2D=True)
+    scene = sampleScene(scenario)
+    simulation = simulator.simulate(scene)
+    records = simulation.result.records["CarSpeed"]
```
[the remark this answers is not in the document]
lola831: Updated in commit 89878f32b7a4dcb66c54d287524193c0d030c1be

src/scenic/simulators/metadrive/simulator.py
```diff
@@ -0,0 +1,289 @@
...
+
+    def getProperties(self, obj, properties):
+        metaDriveActor = obj.metaDriveActor
+        position = utils.metadriveToScenicPosition(
+            metaDriveActor.position, self.scenic_offset
+        )
+        md_ang_vel = metaDriveActor.body.getAngularVelocity()
+        velocity = Vector(*metaDriveActor.velocity, 0)
+        speed = metaDriveActor.speed
+        angularSpeed = 0
```
dfremont, line 233: This should also come from MetaDrive, or you can compute it with `math.hypot(*angularVelocity)`

src/scenic/simulators/metadrive/simulator.py
```diff
@@ -0,0 +1,289 @@
...
+        angularVelocity = Vector(md_ang_vel[0], md_ang_vel[1], md_ang_vel[2])
```
dfremont, line 234:
```suggestion
        angularVelocity = Vector(*metaDriveActor.body.getAngularVelocity())
```
For consistency with the velocity (also getting rid of `md_ang_vel` above). If you prefer to keep `md_ang_vel` that's fine, but then I'd write `*md_ang_vel` here rather than pulling out all 3 coordinates explicitly, as the intent will be clearer.

src/scenic/simulators/metadrive/simulator.py
```diff
@@ -168,26 +137,21 @@ def createObjectInSimulator(self, obj):
...
-                ego_metaDriveActor = list(metadrive_objects.values())[0]
-                obj.metaDriveActor = ego_metaDriveActor
-                self.defined_ego = True
-                return obj.metaDriveActor
-            else:
-                raise SimulationCreationError(
-                    f"Unable to initialize MetaDrive ego vehicle for object {obj}"
-                )
+            ego_metaDriveActor = list(metadrive_objects.values())[0]
+            obj.metaDriveActor = ego_metaDriveActor
```
dfremont, line 141:
```suggestion
            obj.metaDriveActor = list(metadrive_objects.values())[0]
```

src/scenic/simulators/metadrive/simulator.py
```diff
@@ -208,28 +172,38 @@ def executeActions(self, allActions):
...
+                if obj._walking_direction is None:
+                    obj._walking_direction = utils.scenicToMetaDriveHeading(obj.heading)
+                if obj._walking_speed is None:
+                    obj._walking_speed = obj.speed
+                direction = Vector(
+                    math.cos(obj._walking_direction), math.sin(obj._walking_direction)
+                )
+                obj.metaDriveActor.set_velocity(
+                    [direction.x, direction.y], obj._walking_speed
+                )
```
dfremont, line 192:
```suggestion
                direction = [
                    math.cos(obj._walking_direction), math.sin(obj._walking_direction)
                ]
                obj.metaDriveActor.set_velocity(direction, obj._walking_speed)
```

dfremont (changes requested):
The angular speed still needs to be fixed; otherwise looks good!

src/scenic/simulators/metadrive/simulator.py
lola831: Changed in commit a9d0d0079689358452d86403cf4bfb0aa6021058, where angularSpeed is now computed via math.hypot on the angular velocity from MetaDrive.

src/scenic/simulators/metadrive/simulator.py
lola831: Updated in commit a9d0d007

src/scenic/simulators/metadrive/simulator.py
lola831: Updated in commit a9d0d007

src/scenic/simulators/metadrive/simulator.py
lola831: Updated in commit a9d0d007

docs/simulators.rst
```diff
@@ -16,6 +16,32 @@ See the individual entries for details on each interface's capabilities and how
...
+    python -m pip install scenic[metadrive]
+
+Scenic supports both 2D and 3D rendering modes for MetaDrive simulations.
+2D rendering is available on all systems, providing a top-down view.
+However, 3D rendering may not work properly on macOS devices with M-series chips.
+Additionally, there is an issue where cars do not fully brake in certain scenarios.
+These issues are expected to be addressed in the next version of MetaDrive.
+
+Scenic uses OpenDRIVE maps, while MetaDrive relies on SUMO maps. Scenic provides corresponding SUMO maps for OpenDRIVE maps under the :file:`assets/maps/CARLA` directory.
+Additionally, you can convert your own OpenDRIVE maps to SUMO maps using the `netconvert <https://sumo.dlr.de/docs/Networks/Import/OpenDRIVE.html/>`_ tool.
```
dfremont, line 37:
```suggestion
Additionally, you can convert your own OpenDRIVE maps to SUMO maps using the `netconvert <https://sumo.dlr.de/docs/Networks/Import/OpenDRIVE.html>`_ tool.
```

lola831: Fixed broken link in commit 096baa8

dfremont (approved):
Looks great, thanks again Lola!

lola831:
Thank you!

PR #327, opened 2025-02-12
chore: remove OpsGenie and maintenance bot on-call reminder workflow

lola831:
### Description
Removes the OpsGenie on-call reminder script and GitHub Actions workflow since we are no longer using OpsGenie or the maintenance bot. Maintenance on-call rotation is now handled via Slack Round Robin.

### Issue Link
N/A

### Checklist

### Additional Notes
N/A


dfremont (approved):
Thanks Lola!

PR #329, opened 2025-03-04
Update CI for Python 3.13

lola831:
### Description
This PR adds Python 3.13 to tox and CI, and updates the Sphinx dependency to a version that supports Python 3.13.

### Issue Link
N/A

### Checklist

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,11 +1,12 @@
 repos:
-  - repo: https://github.com/psf/black
-    rev: 23.3.0
+  - repo: https://github.com/psf/black-pre-commit-mirror
+    rev: 25.1.0
     hooks:
       - id: black
-        language_version: python3.11
+        language_version: python3.13
 
   - repo: https://github.com/pycqa/isort
     rev: 5.12.0
     hooks:
       - id: isort
+        language_version: python3.13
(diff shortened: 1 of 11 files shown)


.pre-commit-config.yaml
```diff
@@ -1,11 +1,12 @@
 repos:
-  - repo: https://github.com/psf/black
-    rev: 23.3.0
+  - repo: https://github.com/psf/black-pre-commit-mirror
```
dfremont, line 2: What's the difference between this repo and the old one?

dfremont (changes requested):
Looks good. One small question above, and one minor change: the commit where you ran Black should be added to `.git-blame-ignore-revs` to avoid polluting `git blame`.

.pre-commit-config.yaml
lola831: The black-pre-commit-mirror is essentially the same code but pre-compiled and optimized for faster execution within pre-commit hooks. Black's documentation suggests using it in the .pre-commit-config.yaml [here](https://black.readthedocs.io/en/stable/integrations/source_version_control.html)

dfremont: OK, thanks!

dfremont (approved):

Issue #331, opened 2025-03-20
MetaDrive Sumo Map Not Found


oliverc1623:
### System Details

1. Python 3.8.20
2. Scenic 3.1.0a1
3. Operating System Apple M2 macOS Sonoma 14.6.1
4. metadrive-simulator 0.4.3

### Detailed Description

1. MetaDriveSimulator instantiation bug. When initializing MetaDriveSimulator, not specifying a SumoMap causes the error below. 
2. Calling `MetaDriveSimulator()` causes TypeError.
3. Calling `MetaDriveSimulator(sumo_map="../../assets/maps/CARLA/Town05.net.xml")` works fine. 
```
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
/var/folders/w0/3ttb6cs14rq8l4ggd5r1v9440000gn/T/ipykernel_25460/2790578972.py in <cell line: 1>()
----> 1 simulator = MetaDriveSimulator()

~/Documents/tracr/Scenic/src/scenic/simulators/metadrive/simulator.py in __init__(self, timestep, render, render3D, sumo_map, real_time)
     41         self.sumo_map = sumo_map
     42         self.real_time = real_time
---> 43         self.scenic_offset, self.sumo_map_boundary = utils.getMapParameters(self.sumo_map)
     44         if self.render and not self.render3D:
     45             self.film_size = utils.calculateFilmSize(self.sumo_map_boundary, scaling=5)

~/Documents/tracr/Scenic/src/scenic/simulators/metadrive/utils.py in getMapParameters(sumo_map_path)
     39 def getMapParameters(sumo_map_path):
     40     """Retrieve the map parameters."""
---> 41     net_offset, sumo_map_boundary = extractNetOffsetAndBoundary(sumo_map_path)
     42     xmin, ymin, xmax, ymax = sumo_map_boundary
     43     center_x = (xmin + xmax) / 2

~/Documents/tracr/Scenic/src/scenic/simulators/metadrive/utils.py in extractNetOffsetAndBoundary(sumo_map_path)
     29 def extractNetOffsetAndBoundary(sumo_map_path):
     30     """Extracts the net offset and boundary from the given SUMO map file."""
---> 31     tree = ET.parse(sumo_map_path)
     32     root = tree.getroot()
     33     location_tag = root.find("location")

~/opt/anaconda3/envs/tracr/lib/python3.8/xml/etree/ElementTree.py in parse(source, parser)
   1200     """
[...]

PR #335, opened 2025-04-03
Raise ValueError for missing sumo_map in MetaDriveSimulator initializer

lola831:
This PR updates the MetaDriveSimulator initializer to explicitly check for a missing sumo_map parameter. If it's not provided, a clear ValueError is raised. This change is meant for users who manually instantiate the simulator via the API. When using the Scenic command line, the required sumo_map is automatically provided by the Scenic program.

### Issue Link
[Issue #331](https://github.com/BerkeleyLearnVerify/Scenic/issues/331) 

### Checklist

### DIFF
--- a/src/scenic/simulators/metadrive/simulator.py
+++ b/src/scenic/simulators/metadrive/simulator.py
@@ -38,6 +38,10 @@ class MetaDriveSimulator(DrivingSimulator):
         self.render3D = render3D if render else False
         self.scenario_number = 0
         self.timestep = timestep
+        if sumo_map is None:
+            raise ValueError(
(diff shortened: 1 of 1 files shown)


dfremont (changes requested):
Hmm, if there's no use case for having `sumo_map` be `None`, why give it a default value to begin with? I think it would be clearer to have it be a required parameter, like `carla_map` in `CarlaSimulator`.

lola831:
@dfremont Good point—making it a required parameter does make much more sense. The first time I pushed the change, there were some issues on GitHub’s end and nothing got updated, but I've now re-triggered the PR with an empty commit and the changes are reflected here.

dfremont (approved):

PR #340, opened 2025-04-07
Add support for initial velocity in MetaDrive simulations

lola831:
### Description
This PR adds the ability to set an initial velocity for vehicles in the MetaDrive simulator.

### Issue Link
N/A

### Checklist

### Additional Notes
N/A

### DIFF
--- a/src/scenic/simulators/metadrive/simulator.py
+++ b/src/scenic/simulators/metadrive/simulator.py
@@ -112,6 +112,19 @@ class MetaDriveSimulation(DrivingSimulation):
         )
         converted_heading = utils.scenicToMetaDriveHeading(obj.heading)
 
+        vehicle_config = {}
+        if obj.isVehicle:
+            vehicle_config["spawn_position_heading"] = [
+                converted_position,
+                converted_heading,
+            ]
+            if obj.velocity and (obj.velocity.x != 0 or obj.velocity.y != 0):
+                full_md_velocity = utils.scenicToMetaDriveVelocity(obj.velocity)
+                vehicle_config["spawn_velocity"] = utils.scenicToMetaDriveVelocity(
+                    obj.velocity
+                )
+                vehicle_config["spawn_velocity_car_frame"] = True
+
         if not self.defined_ego:
             decision_repeat = math.ceil(self.timestep / 0.02)
             physics_world_step_size = self.timestep / decision_repeat
@@ -122,12 +135,7 @@ class MetaDriveSimulation(DrivingSimulation):
                     decision_repeat=decision_repeat,
                     physics_world_step_size=physics_world_step_size,
                     use_render=self.render3D,
-                    vehicle_config={
-                        "spawn_position_heading": [
-                            converted_position,
-                            converted_heading,
-                        ],
-                    },
+                    vehicle_config=vehicle_config,
                     use_mesh_terrain=self.render3D,
                     log_level=logging.CRITICAL,
                 )
@@ -145,9 +153,7 @@ class MetaDriveSimulation(DrivingSimulation):
         if obj.isVehicle:
             metaDriveActor = self.client.engine.agent_manager.spawn_object(
                 DefaultVehicle,
-                vehicle_config=dict(),
-                position=converted_position,
-                heading=converted_heading,
+                vehicle_config=vehicle_config,
             )
             obj.metaDriveActor = metaDriveActor
             return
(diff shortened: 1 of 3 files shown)


src/scenic/simulators/metadrive/simulator.py
```diff
@@ -112,6 +112,19 @@ def createObjectInSimulator(self, obj):
...
         converted_heading = utils.scenicToMetaDriveHeading(obj.heading)
 
+        vehicle_config = {}
+        if obj.isVehicle:
+            vehicle_config["spawn_position_heading"] = [
+                converted_position,
+                converted_heading,
+            ]
+            if obj.velocity and (obj.velocity.x != 0 or obj.velocity.y != 0):
+                full_md_velocity = utils.scenicToMetaDriveVelocity(obj.velocity)
```
dfremont, line 122: This value isn't used: did you mean to use it in the next line?

src/scenic/simulators/metadrive/simulator.py
```diff
@@ -112,6 +112,19 @@ def createObjectInSimulator(self, obj):
         )
         converted_heading = utils.scenicToMetaDriveHeading(obj.heading)
 
+        vehicle_config = {}
+        if obj.isVehicle:
+            vehicle_config["spawn_position_heading"] = [
+                converted_position,
+                converted_heading,
+            ]
+            if obj.velocity and (obj.velocity.x != 0 or obj.velocity.y != 0):
```
dfremont, line 121: What's the purpose of this check? Is there a problem with setting the `spawn_velocity` to zero? Even if it's redundant to specify a zero velocity, I would do it just to simplify the code.

src/scenic/simulators/metadrive/utils.py
```diff
@@ -79,6 +79,26 @@ def metaDriveToScenicHeading(metaDriveHeading):
     return (scenicHeading + math.pi) % (2 * math.pi) - math.pi
 
 
+def scenicToMetaDriveVelocity(scenic_velocity):
+    """Converts Scenic velocity to MetaDrive velocity by rotating the vector by π/2 (90 degrees)."""
```
dfremont, line 83: This doesn't look right to me: MetaDrive's coordinate system is only shifted from Scenic's, not rotated (`scenicToMetaDrivePosition` only applies a constant offset). So you should not need to convert the velocity vector at all.

dfremont (changes requested):
A few questions and issues above. Also please add a test to make sure it works correctly: for example, you could create a car with a nonzero velocity but no behavior and check that it actually moves in the expected direction after one time step.

src/scenic/simulators/metadrive/simulator.py
lola831: Thanks for catching that—it's a leftover from my previous code and isn't needed.

src/scenic/simulators/metadrive/simulator.py
lola831: I've updated the code so that it always sets the initial velocity, even when it's zero. This change is included in commit 5907be30d64453b7a630880848253d1922ced824.

lola831:
@dfremont Thanks for your comments. I've addressed the issues you raised and added a test to verify that a car with a nonzero initial velocity moves in the expected direction after one time step. These changes are included in commit 5907be30d64453b7a630880848253d1922ced824.

tests/simulators/metadrive/test_metadrive.py
```diff
@@ -128,3 +128,26 @@ def test_pedestrian_movement(getMetadriveSimulator):
...
+        record final ego.position as FinalPos
+        terminate after 1 steps
+    """
+    scenario = compileScenic(code, mode2D=True)
+    scene = sampleScene(scenario)
+    simulation = simulator.simulate(scene)
+    initialPos = simulation.result.records["InitialPos"]
+    finalPos = simulation.result.records["FinalPos"]
+    dx = finalPos[0] - initialPos[0]
+    assert dx < 0, f"Expected car to move west (negative dx), but got dx = {dx}"
```
dfremont, line 153:
```suggestion
    assert dx < -0.1, f"Expected car to move west (negative dx), but got dx = {dx}"
```
Let's use a significant offset here so that small numerical errors can't cause this test to wrongly pass even if the velocity isn't being set properly.

dfremont (changes requested):
Looks great, just one tiny tweak above.

lola831:
> Looks great, just one tiny tweak above.

Sounds good.

dfremont (approved):

PR #357, opened 2025-07-03
Pure Pursuit Controller Implementation + Lane Error calculation Script

aryavenkatesan:
### Description
Added a new simulation file for testing purposes. Added code to behaviors.scenic, simulators.py, and controllers.py in src/scenic/domains/driving.

### Issue Link
N/A

### Checklist

### Additional Notes
Will refine the algorithm within this branch. Once quality is reached, I will autoformat everything and add the proper documentation before attempting to merge this branch to main. 

Here is some footage of the algorithm in action: https://drive.google.com/file/d/1MYkoEuXbe4OmTyklxB5KJ-kcIxRbGGbJ/view?usp=sharing


examples/driving/arya.scenic
Eric-Vin, line 1: Lets rename this to something else

examples/driving/arya.scenic
```diff
@@ -0,0 +1,46 @@
...
+
+
+TODO:
+    Make some kinda waypoints that the car can follow 
+    make the car follow the waypoints using pure pursuit.
+'''
+
+param map = localPath('../../assets/maps/CARLA/Town05.xodr')
+param carla_map = 'Town05'
+param time_step = 1.0/10
```
Eric-Vin: A more reliable way to get timestep length in Scenic is to use the [simulation](https://docs.scenic-lang.org/en/latest/reference/functions.html#simulation) function, and then accessing the `timestep` property (see [here](https://docs.scenic-lang.org/en/latest/modules/scenic.core.simulators.html#scenic.core.simulators.Simulation)). Note that this can only be done at simulation time, but that's probably the only time you need to get the timestep anyway.

examples/driving/arya.scenic
```diff
@@ -0,0 +1,46 @@
...
+
+model scenic.domains.driving.model
+
+behavior PullIntoRoad():
+    while (distance from self to ego) > 15:
+        wait
+    _lon_controller, _lat_controller = simulation().getPurePursuitControllers()
```
Eric-Vin: Don't we handle this default-style behavior deeper down? If so we can remove this.

pid_err.csv
Eric-Vin, line 1: Instead of CSVs, I think we should go with a Python script that runs Scenic programs live and then generates a plot in [matplotlib](https://matplotlib.org/). That way we don't have these bulky files in the repo, and if we run the benchmarks we know the results are relevant to the current state of the code.

src/scenic/domains/driving/behaviors.scenic
```diff
@@ -16,14 +21,11 @@ behavior ConstantThrottleBehavior(x):
     while True:
         take SetThrottleAction(x), SetReverseAction(False), SetHandBrakeAction(False)
 
-behavior DriveAvoidingCollisions(target_speed=25, avoidance_threshold=10):
-    try:
-        do FollowLaneBehavior(target_speed=target_speed)
+behavior DriveAvoidingCollisions(_lon_controller, _lat_controller, target_speed=25, avoidance_threshold=10):
```
Eric-Vin: I don't think we want an underscore at the start of the longitudinal and lateral controller parameters, since that indicates that it's something internal. `lon_controller` and `lat_controller` are probably the way to go.

src/scenic/domains/driving/behaviors.scenic
```diff
@@ -16,14 +21,11 @@ behavior ConstantThrottleBehavior(x):
...
+    try:    
+        do FollowLaneBehavior(_lon_controller, _lat_controller, target_speed=target_speed)
     interrupt when self.distanceToClosest(_model.Vehicle) <= avoidance_threshold:
-        take SetThrottleAction(0), SetBrakeAction(1)
-
-behavior AccelerateForwardBehavior():
-    take SetReverseAction(False), SetHandBrakeAction(False), SetThrottleAction(0.5)
+        take SetThrottleAction(0), SetBrakeActioakeAction(False), SetThrottleAction(0.5)
```
Eric-Vin: Typo on this line (`SetBrakeActioakeAction`). Let's make sure to get test coverage on this code.

src/scenic/domains/driving/controllers.py
```diff
@@ -107,3 +124,127 @@ def run_step(self, cte):
...
         return np.clip(self.output, -1, 1)
+    
+class PurePursuitLateralController:
+    """
+    Pure Pursuit Controller
+
+    Arguments:
+        wb: wheelbase length
+        ld: lookahead distance
+        dt: time step
```
Eric-Vin, line 133: We should probably pull in timestep directly from the simulator (see above comment)

src/scenic/domains/driving/controllers.py
```diff
@@ -107,3 +124,127 @@ def run_step(self, cte):
...
+        cl: car length
```
Eric-Vin, line 134: We can go ahead and pull this from `self` in `run_step` instead of having it as a parameter.

src/scenic/domains/driving/controllers.py
```diff
@@ -107,3 +124,127 @@ def run_step(self, cte):
...
+
+
+            #takes current position and the path 
+            #add plot 
+        """
+        # Define variables
+        lookahead_distance = self.ld
+        circlular_region = CircularRegion(ego.position, lookahead_distance, resolution = 64)
+        polyline_circle = circlular_region.boundary
+        shapely_boundary = polyline_circle.lineString # extract shapley circle and shapely path
```
Eric-Vin: shapley -> shapely

src/scenic/domains/driving/controllers.py
```diff
@@ -107,3 +124,127 @@ def run_step(self, cte):
...
+
+        # Get intersection points of the circle with the second half of the path
+
+        shapely_intersection = shapely_boundary.intersection(output[1])
+        candidate_points = []
+        if isinstance(shapely_intersection, ShapelyPoint):
+            candidate_points = [shapely_intersection]
+        elif isinstance(shapely_intersection, MultiPoint):
+            candidate_points = list(shapely_intersection.geoms)
+        assert all(isinstance(p, ShapelyPoint) for p in candidate_points)
```
Eric-Vin, line 239: We probably want an assertion under an `else` in case all the conditionals fall through. (The assertion here will pass on an empty list)

src/scenic/domains/driving/controllers.py
```diff
@@ -107,3 +124,127 @@ def run_step(self, cte):
...
+
+
+        # Find lookahead point by traversing the path
+
+        if len(candidate_points) == 0:
+            # print("No candidate point found") 
+            # Sometimes no point is found, this is an error
+            # In that case, dont touch steering wheel
+            # Pure Pursuit is a self correcting algorithm so it should be fine
+            cte = self.past_cte
```
Eric-Vin, line 248: How do we handle reaching the end of a desired trajectory? We might want to add some proper handling for this. Any thoughts @psduggirala?

src/scenic/domains/driving/controllers.py
```diff
@@ -107,3 +124,127 @@ def run_step(self, cte):
...
+        else:
+            lookahead_point = candidate_points[0]
```
Eric-Vin, line 255: I think the best way to do this would be to sort all the `candidate_points` by their distance along the path from the ego, and then selecting the first object. Try defining a function here that takes a point and returns the relevant distance, and then you can sort the list directly using `.sort(key=myfunc)`

src/scenic/domains/driving/simulators.py
```diff
@@ -43,7 +44,15 @@ def getLaneFollowingControllers(self, agent):
             )
             lat_controller = PIDLateralController(K_P=0.2, K_D=0.1, K_I=0.0, dt=dt)
         return lon_controller, lat_controller
+    
+    def getPurePursuitControllers(self, agent, cl = 4.5, ld = 7, clwbr = 0.65):
+        dt = self.timestep
```
Eric-Vin, line 49: As above, let's grab this directly from `simulator().timestep`

src/scenic/domains/driving/simulators.py
```diff
@@ -43,7 +44,15 @@ def getLaneFollowingControllers(self, agent):
             )
             lat_controller = PIDLateralController(K_P=0.2, K_D=0.1, K_I=0.0, dt=dt)
         return lon_controller, lat_controller
+    
+    def getPurePursuitControllers(self, agent, cl = 4.5, ld = 7, clwbr = 0.65):
```
Eric-Vin: I'm a little confused about the naming here, since this returns both a pure pursuit controller and a PID controller.

Eric-Vin (changes requested):
Looking good so far! There are a few changes we should make to the code, and we need to add tests (ensuring we get as close to 100% coverage of new lines as possible) and ideally we should move the CSV-based plots to a Python script that we can put under `utils`.

examples/driving/arya.scenic
aryavenkatesan: This file will be deleted before the big merge, it doesn't add anything to Scenic. It was made to help me test without changing existing files.

examples/driving/arya.scenic
aryavenkatesan: This code was copied from Scenic/examples/driving/badlyParkedCarPullingIn.scenic- I can make changes to this if you want me to.

src/scenic/domains/driving/controllers.py
aryavenkatesan: Timestep is taken from simulator.timestep and then passed to the controller upon initialization. See line 49 in Scenic/src/scenic/domains/driving/simulators.py

src/scenic/domains/driving/controllers.py
aryavenkatesan: I've changed it so the car length gets pulled from the ego. Do you think the lookahead distance should be able to be changed as a parameter in run_step? There may be cases where it would lead to better performance but in the papers i've read usually its fixed for the algorithm.

src/scenic/domains/driving/simulators.py
aryavenkatesan: What would you suggest to name it? We might also be inclined to make a pure pursuit lon_controller because there are some issues when the car goes fast and tries to turn (it turns understeers a significant amount).

src/scenic/domains/driving/simulators.py
aryavenkatesan: I changed this to dt = Simulation.timestep and got AttributeError: type object 'Simulation' has no attribute 'timestep'

aryavenkatesan: Also, simulator is an undefined variable

aryavenkatesan: **make this being passed from scenic to python**

Issue #355, opened 2025-06-18
Pedestrian doesn't move in newtonian simulator, and moves too early in MetaDrive with render3D=True


syoung64:
### System Details

1. Python 3.10.12
2. Scenic 3.1.0a1
3. Linux Mint 21.3
4. MetaDrive 0.4.3

### Running `examples/driving/pedestrianJaywalking.scenic`
- behaves as normal in metadrive with render3D=False
- pedestrian doesn't walk in newtonian simulator
- pedestrian starts walking before the car approaches in metadrive with render3D=True

### Steps To Reproduce

1. `scenic --2d -S --model scenic.simulators.metadrive.model examples/driving/pedestrianJaywalking.scenic`
2. `scenic --2d -S --model scenic.simulators.newtonian.driving_model examples/driving/pedestrianJaywalking.scenic`
3. `scenic --2d -S --model scenic.simulators.metadrive.model --param render3D true examples/driving/pedestrianJaywalking.scenic` 

### Issue Submission Checklist

- [x] I am reporting an issue, not asking a question
- [x] I checked the open and closed issues, forum, etc. and have not found any solution
- [x] I have provided all necessary code, etc. to reproduce the issue

dfremont:
@syoung64 Hello Stella, these issues should now be fixed in the `main` branch -- could you please double-check?

PR #358, opened 2025-07-08
MetaDrive: initialize pedestrian velocity on spawn to Scenic starting speed

lola831:
### Description
In MetaDrive’s 3D mode, pedestrians would auto-start at MetaDrive’s default non-zero speed before Scenic applied its initial (usually zero) speed, causing them to move too early. This patch fixes that by calling `set_velocity(direction, obj.speed)` immediately after spawning each Pedestrian, enforcing Scenic’s starting speed.

### Issue Link
[#355 ](https://github.com/BerkeleyLearnVerify/Scenic/issues/355)

### Checklist

### Additional Notes
N/A

### DIFF
--- a/src/scenic/simulators/metadrive/simulator.py
+++ b/src/scenic/simulators/metadrive/simulator.py
@@ -162,6 +162,10 @@ class MetaDriveSimulation(DrivingSimulation):
                 heading_theta=converted_heading,
             )
             obj.metaDriveActor = metaDriveActor
+
+            # Manually initialize pedestrian velocity to Scenic’s starting speed
+            direction = [math.cos(converted_heading), math.sin(converted_heading)]
+            metaDriveActor.set_velocity(direction, obj.speed)
             return
 
         # If the object type is unsupported, raise an error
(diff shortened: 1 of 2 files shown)


lola831:
Added test_static_pedestrian, which launches MetaDrive in 3D (render=True, render3D=True) and asserts that new Pedestrian instances spawn with Scenic’s default speed of zero. On platforms where render3D isn’t supported (e.g. Mac M1), the test now detects the “Could not open window” error and is automatically skipped.

dfremont (approved):
Works great, thanks Lola!

dfremont:
@lola831 I was too hasty: the new test should be decorated with `@pytest.mark.slow` and `@pytest.mark.graphical` since it's slow (at least on my machine, it takes >10 seconds to open the MetaDrive window) and won't run without graphics support. Please make a follow-up PR.

lola831:
@dfremont Thanks for catching that. I’ve added both decorators to `test_static_pedestrian` in PR [#362](https://github.com/BerkeleyLearnVerify/Scenic/pull/362).

Issue #339, opened 2025-04-07
Documentation/Implementation Discrepancy: Default Velocity Not Derived from Speed and Orientation


lola831:
The documentation for the [Object](https://docs.scenic-lang.org/en/latest/_modules/scenic/core/object_types.html#Object) class states that the default velocity is determined by the object's speed and orientation, implying that if you initialize an object (e.g., a car) with a nonzero speed, its velocity should be computed accordingly. However, the code currently defines velocity with a default lambda that always returns Vector(0, 0, 0), regardless of the speed value.

For example, creating a car with an initial speed of 5:
`ego = new Car with speed 5`
results in velocity remaining as Vector(0, 0, 0). As a workaround, one would have to explicitly set the velocity:
`ego = new Car with velocity (5, 0)`

We should update the implementation so that the initial velocity is automatically set based on the initial speed (and orientation). If this behavior is intentional, then the documentation should be revised to accurately describe the current behavior.

PR #359, opened 2025-07-08
Support pedestrian movement in Newtonian simulator

lola831:
### Description
Enable pedestrian movement in Newtonian simulator by storing heading & speed in control and applying via setVelocity in step().

### Issue Link
[#355 ](https://github.com/BerkeleyLearnVerify/Scenic/issues/355)

### Checklist

### Additional Notes
N/A

### DIFF
--- a/src/scenic/simulators/newtonian/simulator.py
+++ b/src/scenic/simulators/newtonian/simulator.py
@@ -35,7 +35,7 @@ current_dir = pathlib.Path(__file__).parent.absolute()
 WIDTH = 1280
 HEIGHT = 800
 MAX_ACCELERATION = 5.6  # in m/s2, seems to be a pretty reasonable value
-MAX_BRAKING = 4.6
+MAX_BRAKING = 6.8
 
 ROAD_COLOR = (0, 0, 0)
 ROAD_WIDTH = 2
@@ -196,7 +196,22 @@ class NewtonianSimulation(DrivingSimulation):
     def step(self):
         for obj in self.objects:
             current_speed = obj.velocity.norm()
-            if hasattr(obj, "hand_brake"):
+            # 1) Pedestrian: we stash .control['heading']/['speed']
+            if hasattr(obj, "control") and "speed" in obj.control:
+                h = (
+                    obj.control["heading"]
+                    if obj.control["heading"] is not None
+                    else obj.heading
+                )
+                s = (
+                    obj.control["speed"]
+                    if obj.control["speed"] is not None
+                    else obj.speed
+                )
+                vel = Vector(0, s).rotatedBy(h)
+                obj.setVelocity(vel)
+            # 2) Vehicle: throttle/brake/steer physics
+            elif hasattr(obj, "hand_brake"):
                 forward = obj.velocity.dot(Vector(0, 1).rotatedBy(obj.heading)) >= 0
                 signed_speed = current_speed if forward else -current_speed
                 if obj.hand_brake or obj.brake > 0:
@@ -221,8 +236,12 @@ class NewtonianSimulation(DrivingSimulation):
                 else:
                     obj.angularSpeed = 0
                 obj.speed = abs(signed_speed)
+
+            # 3) Everything else:
             else:
                 obj.speed = current_speed
+
+            # 4) Integrate motion for all
             obj.position += obj.velocity * self.timestep
             obj.heading += obj.angularSpeed * self.timestep
 
--- a/src/scenic/simulators/newtonian/driving_model.scenic
+++ b/src/scenic/simulators/newtonian/driving_model.scenic
@@ -59,9 +59,18 @@ class Car(Vehicle, Steers):
         return True
 
 class Pedestrian(Pedestrian, NewtonianActor, Walks):
-    pass
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.control = {'heading': None, 'speed': None}
+
+    def setWalkingDirection(self, heading):
+        self.control['heading'] = heading
+
+    def setWalkingSpeed(self, speed):
+        self.control['speed'] = speed
 
 class Debris:
     """Abstract class for debris scattered randomly in the workspace."""
     position: new Point in workspace
     yaw: Range(0, 360) deg
+


src/scenic/simulators/newtonian/simulator.py
```diff
@@ -196,7 +196,22 @@ def isOnScreen(self, x, y):
...
+                    if obj.control["heading"] is not None
+                    else obj.heading
+                )
+                s = (
+                    obj.control["speed"]
+                    if obj.control["speed"] is not None
+                    else obj.speed
+                )
+                vel = Vector(0, s).rotatedBy(h)
+                obj.setVelocity(vel)
```
dfremont:
```suggestion
                obj.velocity = Vector(0, s).rotatedBy(h)
```
I think this is easier to understand, since you don't need to go look up what `setVelocity` does.

tests/simulators/newtonian/test_newtonian.py
```diff
@@ -59,3 +59,13 @@ def test_pickle(loadLocalScenario):
...
+
+
+def test_pedestrian_movement(loadLocalScenario):
+    scenario = loadLocalScenario("pedestrian.scenic", mode2D=True)
+    scene, _ = scenario.generate(maxIterations=1)
+    simulator = NewtonianSimulator()
+    simulation = simulator.simulate(scene, maxSteps=8)
+    init = simulation.result.records["InitialPos"]
+    fin = simulation.result.records["FinalPos"]
+    assert init != fin, f"Pedestrian did not move."
```
dfremont:
```suggestion
    assert init != fin, "Pedestrian did not move."
```

src/scenic/simulators/newtonian/simulator.py
```diff
@@ -196,7 +196,22 @@ def isOnScreen(self, x, y):
...
+            # 1) Pedestrian: we stash .control['heading']/['speed']
+            if hasattr(obj, "control") and "speed" in obj.control:
+                h = (
+                    obj.control["heading"]
+                    if obj.control["heading"] is not None
+                    else obj.heading
+                )
+                s = (
+                    obj.control["speed"]
+                    if obj.control["speed"] is not None
```
dfremont, line 214: If the `heading` and `speed` controls are both `None`, this will change the pedestrian to be moving in the direction `obj.heading`, even if it wasn't before (e.g. if you manually set the velocity so the pedestrian is moving in some other direction: `new Pedestrian with velocity (1, 1)`). If the pedestrian doesn't take any walking actions, we should leave its velocity unchanged.

dfremont (changes requested):
A couple minor and one more serious comment above. Also in testing this I came across #339 again; we should discuss how to fix that.

lola831:
@dfremont  I’ve applied all your suggestions in commit 369035b, thanks!

lola831:
I’ve updated the step function so pedestrians no longer get misclassified. I also added a test (test_pedestrian_velocity_vector) to check that a pedestrian with velocity (1, 1) moves northeast as expected.

tests/simulators/newtonian/test_newtonian.py
```diff
@@ -59,3 +59,27 @@ def test_pickle(loadLocalScenario):
     simulation = simulator.simulate(scene, maxSteps=100)
     egoPos, otherPos = simulation.result.trajectory[-1]
     assert egoPos.distanceTo(otherPos) < 1
+
+
+def test_pedestrian_movement(loadLocalScenario):
+    scenario = loadLocalScenario("pedestrian.scenic", mode2D=True)
```
dfremont: For new tests, let's inline the Scenic code into the function unless there's a specific reason why the code needs to be in another file. It's easier to understand what the test does, and you can use the `getAssetPath` fixture to get the map path in a more robust way (like in the CARLA tests).

src/scenic/simulators/newtonian/simulator.py
```diff
@@ -35,7 +35,7 @@
 WIDTH = 1280
 HEIGHT = 800
 MAX_ACCELERATION = 5.6  # in m/s2, seems to be a pretty reasonable value
-MAX_BRAKING = 4.6
+MAX_BRAKING = 6.8
```
dfremont, line 38: Why was this changed? I don't particularly care what the value is, but I'd rather not change it without a good reason since it will change the behavior of existing Scenic code.

tests/simulators/newtonian/test_newtonian.py
```diff
@@ -59,3 +59,27 @@ def test_pickle(loadLocalScenario):
...
+    scene, _ = scenario.generate(maxIterations=1)
+    simulator = NewtonianSimulator()
+    simulation = simulator.simulate(scene, maxSteps=8)
+    init = simulation.result.records["InitialPos"]
+    fin = simulation.result.records["FinalPos"]
+    assert init != fin, "Pedestrian did not move."
```
dfremont: Let's robustify this a bit so that the condition wouldn't just be satisfied by roundoff error, etc. How about something like `init.distanceTo(fin) > 0.1`?

dfremont (changes requested):
It's working now -- just a couple minor nitpicks.

src/scenic/simulators/newtonian/simulator.py
lola831: I ran into this when testing pedestrianJaywalking.scenic — with MAX_BRAKING = 4.6, the car doesn’t stop in time and just drives past the pedestrian. Increasing it to 6.8 made the braking more realistic and let the ego stop as expected, similar to how it behaves in the MetaDrive simulator.

lola831:
@dfremont Thanks for the feedback. I’ve applied your suggestions.

tests/simulators/newtonian/test_newtonian.py
```diff
@@ -6,7 +6,8 @@
 
 from scenic.domains.driving.roads import Network
 from scenic.simulators.newtonian import NewtonianSimulator
-from tests.utils import pickle_test, sampleScene, tryPickling
+from tests.conftest import getAssetPath
```
dfremont: You don't need this line: `pytest` makes the fixtures available automatically to tests which list them as a parameter. Having the fixture also available in the global namespace could be confusing, so I'd just delete this.

src/scenic/simulators/newtonian/simulator.py
dfremont: OK, thanks.

tests/simulators/newtonian/test_newtonian.py
lola831: @dfremont Thanks! I've gone ahead and removed the `getAssetPath` import.

dfremont (approved):
Thanks, Lola!

PR #363, opened 2025-07-18
Fix boundingPolygon issue and other recent CI errors/warnings

dfremont:
### Description

Main changes in this PR:

- Improve the quality of `MeshRegion.boundingPolygon`, hopefully fixing the frequent CI failures we've been seeing from it.
- Fix a bunch of warnings that I saw in the CI.
- Restore the `test_reproducibility_3d` test, which I'd accidentally commented out parts of previously. Uncommenting those parts triggered a reproducibility failure which I'd previously discussed with @Eric-Vin but never fixed, caused by extra random numbers being generated on the first sample as part of computing a mesh for a region which is then cached and re-used on subsequent samples (so that resetting the random seed doesn't give the exact same behavior, since on the second run fewer random numbers are generated). I've fixed the specific problem by changing our code to not need to generate random numbers for the mesh computation, but we should probably try to come up with a fix for this whole class of bugs later.
- Reduce the boilerplate needed to define AST nodes: I had to make changes so that `ast.dump` continues to work with Scenic ASTs on Python 3.13.5 and newer, and while I was at it I took the opportunity to simplify this code.

### Checklist

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -60,7 +60,7 @@ metadrive = [
 	"sumolib >= 1.21.0",
 ]
 test = [	# minimum dependencies for running tests (used for tox virtualenvs)
-	"pytest >= 7.0.0, <8",
+	"pytest >= 7.0.0, <9",
 	"pytest-cov >= 3.0.0",
 	"pytest-randomly ~= 3.2",
 ]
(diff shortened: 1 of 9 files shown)


lola831:
Tested locally, looks good.

PR #61, opened 2022-07-11
[Parser] Rewrite Scenic Class Definitions

shumbo:
This PR implements special transformations needed for Scenic classes.

PR #371, opened 2025-08-04
Add Python 3.14 Support

lola831:
### Description
This PR adds support for Python 3.14 in Scenic, including CI , tox, and dependency updates.

#### Dependency updates
- Switched Scenic to use `pygame-ce` for Python 3.10+.
- Kept `pygame` for Python versions below 3.10, since `pygame-ce` only supports Python 3.10+.
- This change helps with Python 3.14 support, since `pygame-ce` provides Python 3.14 support.
- ***Because Scenic and VerifAI depend on each other, both need the same conditional `pygame` --> `pygame-ce` change for Python 3.10+.*** (Related VerifAI PR: [PR #61](https://github.com/BerkeleyLearnVerify/VerifAI/pull/61))
- Updated the CARLA Python version guard from `<= 3.10` to `<= 3.12`.

#### CI Support:
- Added Python 3.14 to the GitHub Actions test matrix (`run-tests.yml`)
- Enabled `allow-prereleases: true` to test against prereleases

#### Tox support:
- Extended `tox.ini` to include Python 3.14 environments
- Worked around an issue where `scenic[metadrive]` was being installed even when `python_version > 3.11` by listing `metadrive-simulator` directly with the correct version guard

#### Pre-commit updates:
- Updated black and isort to run under Python 3.14

#### Sphinx compatibility:
- Increased the upper bound for sphinx to `<8` to allow versions compatible with Python 3.14
- Patched `object_description` in `conf.py` to accept `**kwargs` for compatibility with Sphinx ≥7.2 (which passes a `_seen` kwarg) while remaining backward-compatible with older versions

#### Documentation:
- Added a note in the simulator docs reminding users that not all simulators may support the same range of Python versions, users should verify support per simulator.

### Checklist

### DIFF
--- a/.github/workflows/run-tests.yml
+++ b/.github/workflows/run-tests.yml
@@ -33,11 +33,11 @@ jobs:
     strategy:
       fail-fast: true
       matrix:
-        python-version: ["3.8", "3.9", "3.10", "3.11", "3.12", "3.13"]
+        python-version: ["3.8", "3.9", "3.10", "3.11", "3.12", "3.13", "3.14"]
         os: [ubuntu-latest, windows-latest, macos-13, macos-latest]
         include:
           # Only run slow tests on the latest version of Python
-          - python-version: "3.13"
+          - python-version: "3.14"
             slow: true
     runs-on: ${{ matrix.os }}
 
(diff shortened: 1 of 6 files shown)


dfremont (approved):
Looks great, thanks Lola!

PR #372, opened 2025-08-05
Add screen recording support to MetaDrive interface

lola831:
### Description
This PR adds support for screen recording in the MetaDrive simulator interface. When `screen_record=True` and the simulation is run in 2D mode (`render=True`, `render3D=False`), a .gif of the scenario is saved to a `metadrive_gifs/ ` directory by default(created automatically if it doesn't exist). Users can optionally override the filename or output path using `screen_record_filename` and `screen_record_path`. Filenames default to timestamps to avoid overwriting. MetaDrive's screen recording feature only works in top-down 2D render mode. The `metadrive_gifs/ ` directory has been added to `.gitignore` to avoid committing generated recordings.

### Issue Link
N/A

### Checklist

### Additional Notes
N/A

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -140,4 +140,6 @@ dmypy.json
 # generated parser
 src/scenic/syntax/parser.py
 
+# generated media/output
 simulation.gif
+metadrive_gifs/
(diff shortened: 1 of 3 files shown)


dfremont (approved):
Looks good, we can make some minor changes later.

PR #375, opened 2025-08-14
Add MetaDrive RGB/SS sensors

lola831:
### Description
-Add MetaDrive RGBSensor and SSSensor.
-Add default `visionSensorOffset` on Object (front center of the bounding box).
-Standardize sensor params across CARLA and MetaDrive: `offset`, `rotation`, `width`, `height`. Simulator-specific options go under `attributes` (e.g., CARLA convert).
-Add a generic driving domain sensor example that works in both MetaDrive and CARLA

### Issue Link
N/A

### Checklist

### Additional Notes
N/A

### DIFF
--- a/src/scenic/simulators/metadrive/simulator.py
+++ b/src/scenic/simulators/metadrive/simulator.py
@@ -111,7 +111,7 @@ class MetaDriveSimulation(DrivingSimulation):
                 for name, sensor in obj.sensors.items():
                     if isinstance(sensor, MetaDriveRGBSensor):
                         self.drive_env_config[name] = [RGBCamera, 85, 85]
-                    elif isinstance(sensor, MetaDriveRGBSensor):
+                    elif isinstance(sensor, MetaDriveSSSensor):
                         self.drive_env_config[name] = [SemanticCamera, 85, 85]
                     else:
                         raise RuntimeError(f"Unknown sensor type: {type(sensor)}")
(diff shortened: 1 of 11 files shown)


src/scenic/simulators/metadrive/simulator.py
```diff
@@ -139,6 +170,20 @@ def createObjectInSimulator(self, obj):
             metadrive_objects = self.client.engine.get_objects()
             obj.metaDriveActor = list(metadrive_objects.values())[0]
             self.defined_ego = True
+
+            # Adding sensors if available
+            if obj.sensors:
+                for name, sensor in obj.sensors.items():
+                    metadrive_sensor = self.client.engine.get_sensor(name)
```
dfremont: It looks like this part would go wrong if two cars have a sensor with the same name. You could append the index of the object to the name to make it unique (for example).

src/scenic/simulators/metadrive/simulator.py
```diff
@@ -150,6 +195,20 @@ def createObjectInSimulator(self, obj):
                 heading=converted_heading,
             )
             obj.metaDriveActor = metaDriveActor
+
+            # Adding sensors if available
```
dfremont: Can we avoid duplicating this code?

src/scenic/simulators/metadrive/simulator.py
```diff
@@ -160,6 +219,20 @@ def createObjectInSimulator(self, obj):
                 heading_theta=converted_heading,
             )
             obj.metaDriveActor = metaDriveActor
+
+            # Adding sensors if available
```
dfremont: Likewise, let's not write this code out 3 times.

dfremont (changes requested):
This looks good overall, but I'm not sure it will work with multiple cars with the same sensors (see comment above). If you can fix that, I'll go ahead and merge this so we can start testing it out. (But please make a note to add a test for the case of 2 cars with a sensor of the same name.)

PR #376, opened 2025-08-17
Move CARLA-specific setup from sensors to simulator to unblock docs build

lola831:
### Description
Move the` carla` import and Transform/ColorConverter handling from `sensors.py` into the simulator so CARLA-specific setup lives there and lets docs build without CARLA installed.

### Issue Link
N/A

### Checklist

### Additional Notes
N/A

### DIFF
--- a/src/scenic/simulators/carla/sensors.py
+++ b/src/scenic/simulators/carla/sensors.py
@@ -32,9 +32,7 @@ class CarlaVisionSensor(CallbackSensor):
 
         self.convert = None
         convert = self.attributes.get("convert")
-        if convert is not None and not (
-            isinstance(convert, str) or isinstance(convert, int)
-        ):
+        if convert is not None and not isinstance(convert, (str, int)):
             raise TypeError("'convert' has to be int or string.")
         self.convert = convert
 
--- a/src/scenic/simulators/carla/sensors.py
+++ b/src/scenic/simulators/carla/sensors.py
@@ -1,8 +1,3 @@
-import os
-from time import sleep
-
-import carla
-import cv2
 import numpy as np
 
 from scenic.core.sensors import CallbackSensor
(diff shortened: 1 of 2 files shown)


src/scenic/simulators/carla/sensors.py
```diff
@@ -38,13 +32,11 @@ def __init__(
...
-        if convert is not None:
-            if isinstance(convert, int):
-                self.convert = convert
-            elif isinstance(convert, str):
-                self.convert = carla.ColorConverter.names[convert]
-            else:
-                raise TypeError("'convert' has to be int or string.")
+        if convert is not None and not (
+            isinstance(convert, str) or isinstance(convert, int)
+        ):
```
dfremont:
```suggestion
        if convert is not None and not isinstance(convert, (str, int)):
```

src/scenic/simulators/carla/simulator.py
```diff
@@ -258,8 +258,27 @@ def createObjectInSimulator(self, obj):
...
+                    try:
+                        sensor.convert = carla.ColorConverter.names[sensor.convert]
+                    except KeyError:
+                        raise ValueError(
+                            f"Unknown CARLA ColorConverter '{sensor.convert}'"
+                        )
+
+                transform = carla.Transform(
+                    carla.Location(
+                        x=sensor.offset[0], y=sensor.offset[1], z=sensor.offset[2]
```
dfremont: Better double-check that we don't need to do any coordinate conversion here: the Y axis of CARLA's global coordinate system points in the opposite direction to Scenic's, so we negate the Y coordinate for objects. Presumably we'll have to do the same here, but I would check. (We want a positive Y offset to move the sensor forward with respect to the object, and a negative Y offset to move it backward.)

src/scenic/simulators/carla/simulator.py
```diff
@@ -258,8 +258,27 @@ def createObjectInSimulator(self, obj):
...
+                    ),
+                    carla.Rotation(
+                        pitch=sensor.rotation[0],
+                        yaw=sensor.rotation[1],
```
dfremont: Likewise, better check if we need to do a conversion here. For objects we negate `yaw` (see `utils.py`) because CARLA uses a left-handed rotation for `yaw` for some reason (i.e., a positive `yaw` rotates the object clockwise when viewed from above, whereas Scenic's convention is to rotate it counterclockwise).

PR #381, opened 2025-09-04
Add shoulderAt and make roadDirection include shoulders

lola831:
Adds `Network.shoulderAt` and updates `roadDirection` to return the shoulder’s orientation when the point lies on a shoulder (without treating shoulders as roads). Adds a test covering both the new query and the orientation behavior. This fixes the driving-domain example `badlyParkedCarPullingIn`, where a parked car on the shoulder previously received a default 0 roadDirection

### Issue Link
N/A

### Checklist

### Additional Notes
N/A


tests/domains/driving/test_network.py
```diff
@@ -226,3 +226,19 @@ def checkGroup(group):
                     assert rev is not maneuver
                     assert rev.startLane.road is maneuver.endLane.road
                     assert rev.endLane.road is maneuver.startLane.road
+
+
+def test_shoulder_at_and_direction(network):
+    if not network.shoulders:
+        pytest.skip("map has no shoulders")
```
dfremont: Does one of `Town01` or `Town03` not have shoulders? We wouldn't want this test to become useless later if we accidentally switch to a map with no shoulders -- I'd rather have it fail and inform us that we need to select a different map.

src/scenic/domains/driving/roads.py
```diff
@@ -1335,6 +1348,29 @@ def show(self, labelIncomingLanes=False):
...
+                plt.quiver(
+                    cx,
+                    cy,
+                    cu,
+                    cv,
+                    pivot="middle",
+                    headlength=4.5,
+                    scale=0.06,
+                    units="dots",
+                    color="#FF4444",
```
dfremont: Let's use the same color as for the curb outline (`#606060` I think) so that the arrows aren't confused with the red lane boundaries. Actually I'm not sure if we should show these arrows by default at all: it might be better to not show them to avoid confusion with the lane centerlines (which convey the direction information anyway). You could add a keyword argument to `show` to turn them on so that we can do that for debugging purposes in the future, but have them off by default.

dfremont (changes requested):
This looks good. Just a few comments above, and I would also update `nominalDirectionsAt` to include shoulders so that it's consistent with `roadDirection`.

I think we'll also want to update `elementAt` to include shoulders and sidewalks (currently you get nothing inside those regions, even though there are network elements there), but this gets to the broader question of whether shoulders should be considered part of their parent roads, which we need to discuss in the broader group.

tests/domains/driving/test_network.py
lola831: @dfremont Sounds good. I've updated this to fail if the map has no shoulders instead of skipping.

src/scenic/domains/driving/roads.py
lola831: @dfremont I switched the arrows to use the same color as the curb outline and added `showCurbArrows` argument to `show()` so they are hidden by default.

src/scenic/domains/driving/roads.py
```diff
@@ -1230,15 +1242,18 @@ def _findPointInAll(self, point, things, key=lambda e: e):
...
-        simulation if **reject** is true (default false).
+        The elements are considered in the following order:
+        1. Intersection
+        2. Road
+        3. Shoulder
+        4. Sidewalk
+
+        If the point lies in one of these, that element is returned. Otherwise
+        we return :obj:`None`, or reject the simulation if **reject** is true
+        (default false).
```
dfremont: Let's add a comment that points lying within `self.tolerance` of one of these elements are also OK.

tests/domains/driving/test_network.py
```diff
@@ -48,9 +48,6 @@ def test_element_tolerance(cached_maps, pytestconfig):
         assert network.roadAt(pt) is None
         with pytest.raises(RejectionException):
             network.roadAt(pt, reject=True)
-        pt = toofar_noint.uniformPointInner()
-        assert network.elementAt(pt) is None
-        assert not network.nominalDirectionsAt(pt)
```
dfremont, line 53: Let's put these lines back but change `toofar_noint` to be far away from all elements (so instead of starting with just the drivable region, also union in the sidewalk region and the shoulder region before buffering).

tests/domains/driving/test_network.py
```diff
@@ -226,3 +223,36 @@ def checkGroup(group):
...
+def test_shoulder(network):
+    sh = network.shoulders[0]
+    for _ in range(5):
+        pt = sh.uniformPointInner()
+        assert network.shoulderAt(pt) is sh
+        assert network.elementAt(pt) is sh
+        so = sh.orientation[pt]
+        assert network.roadDirection[pt] == pytest.approx(so)
+        dirs = network.nominalDirectionsAt(pt)
+        assert dirs[0] == pytest.approx(so)
```
dfremont: Let's add `assert len(dirs) == 1` before this.

tests/domains/driving/test_network.py
dfremont: Oops, I see you essentially did this in the new `test_reject_outside` below. Let's fold that test into this one as above.

lola831:
@dfremont Updated `defaultRoadDirection` to use `findPointIn` over `_roadDirElems `(roads + shoulders), and `nominalDirectionsAt` to use `findPointIn` over ` _nominalDirElems` (intersections + roads + shoulders). This fixes the prior failure in `test_orientation_consistency`.

dfremont (approved):

PR #382, opened 2025-09-05
3d driving

ethan10mak:
### Description
This PR adds support for 3D mode.
- Elevation parsed from Open Drive files
- Added mesh generation through the Trimesh library
- Tested driving simulations through CARLA
- Updated older test cases to generate in 2D mode

### Issue Link
N/A

### Checklist

### Additional Notes
N/A


src/scenic/simulators/carla/model.scenic
```diff
@@ -102,7 +102,7 @@ param weather = Uniform(
     'MidRainSunset',
     'HardRainSunset'
 )
-param snapToGroundDefault = is2DMode()
+param snapToGroundDefault = True
```
dfremont: What was this change meant to fix?

tests/domains/driving/test_driving.py
```diff
@@ -55,34 +55,45 @@ def test_driving_2D_map(cached_maps):
...
     try:
         # First, try the original .xodr file
         scenario = compileDrivingScenario(
-            cached_maps, path=path, code=basicScenario, useCache=False
+            cached_maps if use2DMap else cached_maps3D,
+            path=path,
+            code=basicScenario,
+            useCache=False,
+            mode2D=use2DMap,
+            params={"use2DMap": use2DMap, "tolerance": 0.115},
```
dfremont: Let's not manually specify the tolerance here unless we have to: it's better to test with the default tolerance, since that's what people will probably use.

tests/domains/driving/test_driving.py
```diff
@@ -96,6 +107,8 @@ def test_elements_at(cached_maps):
             param crossing = network.crossingAt(spot)
             param intersection = network.intersectionAt(spot)
     """,
+        mode2D=use2DMap,
+        params={"use2DMap": use2DMap},
```
dfremont: Here and in the other tests, you shouldn't need to specify `use2DMap` since it will be turned on automatically in 2D mode and otherwise turned off.

src/scenic/core/regions.py
```diff
@@ -563,6 +563,7 @@ def evaluateInner(self, context):
     def containsPoint(self, point):
         return any(region.containsPoint(point) for region in self.footprint.regions)
 
+    # TODO: Need to implement this
```
dfremont:
```suggestion
```

src/scenic/simulators/carla/model.scenic
ethan10mak: This change was meant to fix the issue of the vehicle clipping into the road when spawning. After testing again to confirm, this change was the primary fix to this issue. Reverting snapToGroundDefault = is2DMode() will cause the vehicle to clip into the road.

Also, changing the height of the vehicle in /driving/model.scenic seems to have no effect on the clipping issue, regardless of if I set snapToGroundDefault to True or False.

lola831:
The test suite was passing except for the two CARLA tests in `tests/simulators/carla/test_basic.py` that use `basic.scenic` (`test_basic[False]` and `test_car_created[False]`).

The new surface-volume sampler, `sampleSurfaceInVolume`, helped, but those 3D cases can still occasionally require more iterations to generate a scene, so I increased the iteration limit for those tests from 1000 to 2000.

src/scenic/core/regions.py
```diff
@@ -2767,11 +2881,22 @@ def __init__(
             numpy.linalg.norm(b - a, axis=1), (len(a), 1)
         )
 
+    def intersects(self, other, triedReversed=False):
```
dfremont, line 2884: The implementation here is not correct: it ignores the Z coordinates. Using Shapely isn't a bad idea (it's probably faster than trying to compute intersections ourselves), but you'd have to check that at least one of the 2D intersections it finds is actually an intersection in 3D. For intersection points, that's easy; for segments, I think it's enough to check each of the endpoints.

src/scenic/domains/driving/model.scenic
```diff
@@ -283,13 +307,14 @@ class Vehicle(DrivingObject):
             distribution derived from car color popularity statistics; see
             :obj:`Color.defaultCarColor`.
     """
-    regionContainedIn: roadOrShoulder
-    position: new Point on road
+    regionContainedIn: roadOrShoulder.boundingPolygon # Change region to bounding polygon region
+    position: centerOnSurface(road, self.baseOffset, self.contactTolerance)
```
dfremont, line 311: I think I see why we need this now, but with a mechanism for default specifiers, this would be equivalent to simply `on road`, I think.

src/scenic/domains/driving/roads.py
```diff
@@ -136,8 +146,18 @@ def guessTypeFromLanes(
...
+        if isinstance(connecting.centerline, PathRegion):
+            startDir = (
+                connecting.centerline.vertices[0] - connecting.centerline.vertices[1]
+            )
+            endDir = (
+                connecting.centerline.vertices[-2] - connecting.centerline.vertices[-1]
+            )
+        else:
+            startDir = connecting.centerline[1] - connecting.centerline[0]
+            endDir = connecting.centerline[-1] - connecting.centerline[-2]
```
dfremont, line 160: I think we can revert this change since `PathRegion.__getitem__` was added above. (Also the indices look backwards for the PathRegion case -- better double-check this!)

src/scenic/domains/driving/roads.py
```diff
@@ -246,10 +269,18 @@ def __attrs_post_init__(self):
         assert self.uid is not None or self.id is not None
         if self.uid is None:
             self.uid = self.id
-
-        super().__init__(
-            polygon=self.polygon, orientation=self.orientation, name=self.name
-        )
+        if isinstance(self.region, MeshSurfaceRegion):
+            self.region.__init__(
```
dfremont, line 273: Shouldn't call `__init__` again: the region is already initialized.

PR #383, opened 2025-09-08
Robosuite-Scenic Integration [Custom only]

sahil-tgs:
# RoboSuite-Scenic Interface

## Setup
```bash
# Install Scenic
pip install -e .
# Install RoboSuite
pip install robosuite
```

## Main Files
- `src/scenic/simulators/robosuite/simulator.py` 
- `src/scenic/simulators/robosuite/model.scenic` 

## Overview
This new version now fully supports custom environment generation. We can now use native Robosuite Objects like Robots, arena elements like Table, Manipulation Objects like Cube, sphere etc, as components for Setting up different Environment configuration inside Robosuite via Scenic.
- Native RoboSuite robots (Panda, UR5e, Sawyer, Jaco, IIWA)
- Arena elements (Tables with configurable dimensions)
- Primitive objects (Box, Ball, Cylinder, Capsule) with customizable sizes
- Complex objects (Milk, Cereal, Bottle, etc.) with fixed dimensions
- Dynamic behaviors for robot control

## Architecture
We are no longer supporting Robosuite's standard set of test environments called "Tasks", which where earlier supported param use_environment:, now every environment is a custom environment. 
The enviroment is custom generated but uses components of Robosuite enviroments like MultiTableArena for Manipulable Table, to build custom workspace/ arena configuration, and EmptyArena for an empty room which is  (default env setup). 
For any Robosuite Simulation instance, the bare minimum required object is a robot object as an ego object without this no simulation env can be generated. [Limitation]
Simmilarly If one Robot generates it lode ups the default RObosuite enviroment setup with a EmptyArena (empty room) [Limitation]
other than that, we have full controll over how many robots we can define or generate, or no ot table we generate and where we postion them or how many manipulation objects like cubes and spheres we generate. The current code also have extended support for some of the preexisting complex objects in Robosuite, so we can place them too.

## Examples

| File | Description |
|------|-------------|
| `dual_table_workspace.scenic` | Two-table setup with random object placement |
| `four_table_workspace.scenic` | Four tables with diverse object types |
| `stacked_cubes.scenic` | Stacked objects on single table |
| `stack_lift.scenic` | **Benchmark** - demonstrates behaviors, multi-object scenes |


examples/robosuite/stack_lift.scenic
```diff
@@ -0,0 +1,28 @@
...
+work_table = new Table at (0.6, 0, 0.8),
+    with width 0.6,
+    with length 1.2,
+    with height 0.05
+
+bottom_cube = new Box at (0.6, 0, 0.83),
+    with color (0.2, 0.3, 0.8, 1),
+    with width 0.06, with length 0.06, with height 0.06
+
+top_cube = new Box at (0.6, 0, 0.89)
```
dfremont, line 12:
```suggestion
top_cube = new Box at (0.6, 0, 0.89),
```

src/scenic/simulators/robosuite/simulator.py
```diff
@@ -0,0 +1,566 @@
+# src/scenic/simulators/robosuite/simulator.py
+
+"""RoboSuite Simulator interface for Scenic - Custom Environment Only."""
+
+from typing import Dict, List, Any, Optional, Union
+import numpy as np
+import mujoco
+
+try:
+    import robosuite as suite
```
dfremont, line 10: This import causes robosuite to print various warnings for me even if we won't actually use robosuite (e.g. if you run `scenic` without the `-S` option). Unless it would cause other problems, it might be better to not import robosuite at all unless it's going to be used. That's what we do for MetaDrive and CARLA: we set up their `model.scenic` files to work without needing to import the simulator packages, but their `simulator.py` files raise an exception if those packages are missing.

src/scenic/simulators/robosuite/simulator.py
```diff
@@ -0,0 +1,566 @@
...
+
+
+def _extract_cylinder_size(config: Dict) -> List[float]:
+    """Extract [radius, height] for cylinder-like objects."""
+    if 'radius' in config and 'height' in config:
+        return [config['radius'], config['height']]
+    
+    size = config.get('size', [0.02, 0.02, 0.04])
+    if len(size) == 3:  # Convert from [width, length, height]
+        radius = (size[0] + size[1]) / 4
```
dfremont, line 90: This doesn't look right, although I'm not sure what you mean by "radius" here.

src/scenic/simulators/robosuite/simulator.py
```diff
@@ -0,0 +1,566 @@
...
+        self.model = ManipulationTask(
+            mujoco_arena=mujoco_arena,
+            mujoco_robots=[robot.robot_model for robot in self.robots],
+            mujoco_objects=self.mujoco_objects
+        )
+    
+    def _position_robots(self):
+        """Position robots based on scenic configuration."""
+        for i, robot_config in enumerate(self.scenic_config.get('robots', [])):
+            if i < len(self.robots):
```
dfremont, line 125: I don't see where `self.robots` is initialized (in the superclass `__init__`?), so I'm not sure how this condition can fail. If it does fail, is skipping positioning one of the robots listed in `scenic_config` really the right behavior? I would have thought this would mean there is a problem somewhere.

src/scenic/simulators/robosuite/simulator.py
```diff
@@ -0,0 +1,566 @@
...
+        """Check task success."""
+        return False
+
+
+class RobosuiteSimulator(Simulator):
+    """Simulator for RoboSuite custom environments."""
+    
+    def __init__(self, render: bool = True, real_time: bool = True, speed: float = 1.0,
+                 env_config: Optional[Dict] = None, controller_config: Optional[Dict] = None,
+                 camera_view: Optional[str] = None, lite_physics: Optional[bool] = None):
```
dfremont, line 208: Docstring should explain what these optional arguments do. (See `CarlaSimulator` or `MetaDriveSimulator` for examples.)

src/scenic/simulators/robosuite/simulator.py
```diff
@@ -0,0 +1,566 @@
...
+        """Execute actions by calling their applyTo methods."""
+        super().executeActions(allActions)
+        
+        self.pending_robot_action = None
+        
+        for agent in self.agents:
+            if agent in allActions and allActions[agent]:
+                for action in allActions[agent]:
+                    if action and hasattr(action, 'applyTo'):
+                        action.applyTo(agent, self)
```
dfremont, line 432: This seems to call `applyTo` twice (first in `executeActions` above, then here). Do you need this loop at all?

src/scenic/simulators/robosuite/simulator.py
```diff
@@ -0,0 +1,566 @@
...
+    def step(self):
+        """Step the simulation forward one timestep."""
+        if hasattr(self, '_done') and self._done:
+            return
+        
+        # Store previous positions
+        for name, body_id in self.body_id_map.items():
+            self.prev_positions[name] = self.data.xpos[body_id].copy()
+        
+        action = (self.pending_robot_action if self.pending_robot_action is not None
```
dfremont, line 444: This seems to assume there's only a single robot. We should support multiple robots taking actions at the same time.

src/scenic/simulators/robosuite/simulator.py
```diff
@@ -0,0 +1,566 @@
...
+        }
+    
+    def _setup_observables(self) -> Dict:
+        """Setup observables for objects."""
+        observables = super()._setup_observables()
+        
+        for obj in self.mujoco_objects:
+            @sensor(modality="object")
+            def obj_pos(obs_cache, name=obj.name):
+                return np.array(self.sim.data.body_xpos[self.obj_body_ids[name]])
```
dfremont, line 169: Do we need to make these observables, or can we just call code like this to directly query the position in `getProperties`?

src/scenic/simulators/robosuite/simulator.py
```diff
@@ -0,0 +1,566 @@
...
+            if prop == 'position':
+                values[prop] = self._get_object_position(obj)
+            elif prop == 'joint_positions' and robot_idx is not None:
+                values[prop] = self._get_robot_joints(robot_idx)
+            elif prop == 'eef_pos' and robot_idx is not None:
+                values[prop] = self._get_eef_position(robot_idx)
+            elif prop == 'gripper_state' and robot_idx is not None:
+                values[prop] = self._get_gripper_state(robot_idx)
+            else:
+                values[prop] = getattr(obj, prop, None)
```
dfremont, line 474: This doesn't properly update the other dynamic properties like `yaw`, `pitch`, `roll`, `velocity`, etc. -- those need to have their current values looked up in robosuite.

examples/robosuite/stack_lift.scenic
```diff
@@ -0,0 +1,28 @@
...
+
+bottle = new Bottle at (0.6, 0.3, 0.83)
+
+pickup_object = top_cube
+
+behavior StackLift():
+    """Pick up the top cube from the stack and lift it."""
+    do PickObject(pickup_object)
+    do LiftToHeight(1.05)
+    for _ in range(10):
```
dfremont, line 23: If this limit of 10 steps is exceeded, the behavior will return and so the simulation will never end (unless you imposed a time limit when running Scenic). Did you mean to write `terminate simulation` at the end of the behavior?

examples/robosuite/stacked_cubes.scenic
```diff
@@ -0,0 +1,19 @@
...
+work_table = new Table at (0.6, 0, 0.8),
+    with width 0.6,
+    with length 1.2,
+    with height 0.05
+
+bottom_cube = new Box at (0.6, 0, 0.83),
+    with color (0.2, 0.3, 0.8, 1),
+    with width 0.06, with length 0.06, with height 0.06
+
+top_cube = new Box at (0.6, 0, 0.89),
```
dfremont, line 14:
```suggestion
top_cube = new Box on bottom_cube, 
```
This should work if the shape of the `Box` is set properly. If you want the top cube centered on the bottom cube, you can do `on top of bottom_cube`.

src/scenic/simulators/robosuite/model.scenic
```diff
@@ -0,0 +1,253 @@
+# src/scenic/simulators/robosuite/model.scenic
+
+"""Scenic world model for RoboSuite - Custom Environments Only."""
+
+from .simulator import RobosuiteSimulator, SetJointPositions, OSCPositionAction
+
+# Global parameters with defaults matching Robosuite's defaults
+param env_config = {}
```
dfremont, line 8: Please add documentation for these global parameters to the module docstring (see `scenic.simulators.carla.model` for an example).

src/scenic/simulators/robosuite/model.scenic
```diff
@@ -0,0 +1,253 @@
...
+    real_time=globalParameters.real_time,
+    speed=globalParameters.speed,
+    env_config=globalParameters.env_config,
+    controller_config=globalParameters.controller_config,
+    camera_view=globalParameters.camera_view,
+    lite_physics=globalParameters.lite_physics
+)
+
+# Default values dictionary
+DEFAULTS = {
```
dfremont, line 28: I don't think gathering all the defaults into a dict like this makes the code any clearer. I'd just put the values directly into the classes where they are used, so that they're immediately present if you go to look up the default value of a property in the class definition.

src/scenic/simulators/robosuite/model.scenic
```diff
@@ -0,0 +1,253 @@
...
+}
+
+# Base classes
+class RoboSuiteObject(Object):
+    """Base class for all RoboSuite objects."""
+    density: DEFAULTS['density']
+    friction: DEFAULTS['friction']
+    solref: DEFAULTS['solref']
+    solimp: DEFAULTS['solimp']
+    shape: BoxShape()
```
dfremont, line 66: This is the default value, so you don't need to override `shape`.

src/scenic/simulators/robosuite/model.scenic
```diff
@@ -0,0 +1,253 @@
...
+    allowCollisions: True
```
dfremont, line 67: Why allow collisions by default? We don't want to be able to place cubes inside the table, for example.

src/scenic/simulators/robosuite/model.scenic
```diff
@@ -0,0 +1,253 @@
...
+
+# Table for arena setup
+class Table(RoboSuiteObject):
+    """Table in environment."""
+    isTable: True
+    width: DEFAULTS['table_width']
+    length: DEFAULTS['table_length']
+    height: DEFAULTS['table_thickness']
+    position: (0, 0, DEFAULTS['table_height'])
```
dfremont, line 76: This should not be necessary if you create the table with `new Table on floor` or similar: Scenic will take into account the height of the table.

src/scenic/simulators/robosuite/model.scenic
```diff
@@ -0,0 +1,253 @@
...
+    length: DEFAULTS['object_size']
+    height: DEFAULTS['object_size']
+
+class Ball(ManipulationObject):
+    """Ball/sphere object."""
+    objectType: "Ball"
+    radius: DEFAULTS['object_size']
+    width: DEFAULTS['object_size'] * 2
+    length: DEFAULTS['object_size'] * 2
+    height: DEFAULTS['object_size'] * 2
```
dfremont, line 97:
```suggestion
    width: self.radius * 2
    length: self.radius * 2
    height: self.radius * 2
```
And likewise for the other shapes below.

src/scenic/simulators/robosuite/model.scenic
```diff
@@ -0,0 +1,253 @@
...
+    objectType: "SquareNut"
+
+class RoundNut(ManipulationObject):
+    """Round nut object."""
+    objectType: "RoundNut"
+
+# Robots (matching RoboSuite's naming)
+class Robot(RoboSuiteObject):
+    """Base robot class."""
+    robot_type: "Panda"
```
dfremont, line 149: For consistency with Scenic, please use camel case for these properties, e.g. `robotType`

src/scenic/simulators/robosuite/simulator.py
```diff
@@ -0,0 +1,566 @@
...
+            return list(self._current_obs.get(f'robot{robot_idx}_gripper_qpos', [0, 0]))
+        return [0, 0]
+    
+    def getCurrentObservation(self) -> Optional[Dict]:
+        """Get current observation dictionary."""
+        return self._current_obs
+    
+    def checkSuccess(self) -> bool:
+        """Check if task is successfully completed."""
+        if hasattr(self.robosuite_env, '_check_success'):
```
dfremont, line 516: How can this condition ever be false? More broadly, what's the point of this method?

src/scenic/simulators/robosuite/simulator.py
```diff
@@ -0,0 +1,566 @@
...
+    def __init__(self, position_delta: Optional[List[float]] = None,
+                 orientation_delta: Optional[List[float]] = None,
+                 gripper: Optional[float] = None):
+        self.position_delta = position_delta or [0, 0, 0]
+        self.orientation_delta = orientation_delta or [0, 0, 0]
+        self.gripper = gripper if gripper is not None else 0
+    
+    def applyTo(self, agent, sim: RobosuiteSimulation):
+        """Apply action to agent."""
+        if hasattr(sim, 'robots') and agent in sim.robots:
```
dfremont, line 554: The `hasattr` check will always be true and so is redundant. The other check looks wrong: if `agent` is not a robot and so does not support this action, the interface should raise an exception rather than silently do nothing. You can use the `Action.canBeTakenBy` method to restrict which objects can take which actions.

src/scenic/simulators/robosuite/simulator.py
```diff
@@ -0,0 +1,566 @@
...
+            robot_idx = sim.robots.index(agent)
+            if robot_idx < len(sim.robosuite_env.robots):
```
dfremont, line 556: I'm not sure what this check is meant to do. Can it be false?

src/scenic/simulators/robosuite/simulator.py
```diff
@@ -0,0 +1,566 @@
...
+# Actions
+class SetJointPositions(Action):
+    """Set robot joint positions."""
+    
+    def __init__(self, positions: List[float]):
+        self.positions = positions
+    
+    def applyTo(self, agent, sim: RobosuiteSimulation):
+        """Apply action to agent."""
+        if hasattr(sim, 'robots') and agent in sim.robots:
```
dfremont, line 536: See similar comment for `OSCPositionAction`.

src/scenic/simulators/robosuite/simulator.py
```diff
@@ -0,0 +1,566 @@
...
+        self.agents = []
+        
+        super().__init__(scene, **kwargs)
+    
+    def setup(self):
+        """Initialize the RoboSuite environment."""
+        super().setup()
+        
+        # Extract configuration from Scenic scene
+        scenic_config = self._extract_scenic_config()
```
dfremont, line 274: Packing all the Scenic data into a dict here and then unpacking it later seems redundant to me. Can you just pass the `RobosuiteSimulation` object into `ScenicManipulationEnv` and access its attributes?

src/scenic/simulators/robosuite/simulator.py
```diff
@@ -0,0 +1,566 @@
...
+    "frontview": 0,
+    "birdview": 1,
+    "agentview": 2,
+    "sideview": 3,
+    "robot0_robotview": 4,
+    "robot0_eye_in_hand": 5
+}
+
+# Object type mapping
+OBJECT_FACTORIES = {
```
dfremont, line 49: If all the `objectType` property is used for is to index into this dictionary, it would be simpler to change these lambdas into methods on the original classes. For example the `Ball` class's implementation of `makeRobosuiteObject` (or whatever you want to call it) would create the `BallObject` as below referring to the appropriate attributes of `self`.

examples/robosuite/stack_lift.scenic
sahil-tgs: This was actually leftover from an earlier debugging session where I needed the simulation to continue running after the behavior completed, it allowed me to inspect the final robot state and manipulate the environment directly through Robosuite's viewer for testing purposes. And yes for the release version, this should properly terminate, I'll add the  `terminate simulation` at the end.

examples/robosuite/stacked_cubes.scenic
sahil-tgs: yes in the new PR and most recent commit, I have done exactly that.

PR #384, opened 2025-09-08
Add wait for and wait until statements

lola831:
### Description
This PR adds two new dynamic statements:
`wait for <scalar> (seconds | steps)`
`wait until <boolean>`

They work like `do ... for/until`, but instead of running a behavior, they advance the scenario without performing any actions.

Here's an example  that uses the new `wait` statements: 

```
param map = localPath('../../assets/maps/CARLA/Town01.xodr')
param time_step = 1.0/10

model scenic.simulators.metadrive.model

behavior WaitForThenDrive():
    wait for 20 steps
    do FollowLaneBehavior()

behavior WaitUntilThenDrive():
    wait until distance from self to ego < 40
    do FollowLaneBehavior()

ego = new Car with behavior WaitForThenDrive()

frontCar = new Car ahead of ego by 50,
                    with behavior WaitUntilThenDrive()

terminate after 100 steps
```

### Issue Link
N/A

### Checklist

### Additional Notes
N/A

### DIFF
--- a/src/scenic/syntax/scenic.gram
+++ b/src/scenic/syntax/scenic.gram
@@ -2155,9 +2155,9 @@ scenic_abort_stmt: "abort" { s.Abort(LOCATIONS) }
 
 scenic_take_stmt: "take" elts=(','.expression+) { s.Take(elts=elts, LOCATIONS) }
 
-scenic_wait_for_stmt: "wait" "for" u=scenic_dynamic_duration { s.WaitFor(duration=u, LOCATIONS) }
+scenic_wait_for_stmt: "wait" 'for' u=scenic_dynamic_duration { s.WaitFor(duration=u, LOCATIONS) }
 
-scenic_wait_until_stmt: "wait" "until" cond=expression { s.WaitUntil(cond=cond, LOCATIONS) }
+scenic_wait_until_stmt: "wait" 'until' cond=expression { s.WaitUntil(cond=cond, LOCATIONS) }
 
 scenic_wait_stmt: "wait" { s.Wait(LOCATIONS) }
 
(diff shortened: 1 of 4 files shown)


src/scenic/syntax/scenic.gram
```diff
@@ -2153,6 +2155,10 @@ scenic_abort_stmt: "abort" { s.Abort(LOCATIONS) }
 
 scenic_take_stmt: "take" elts=(','.expression+) { s.Take(elts=elts, LOCATIONS) }
 
+scenic_wait_for_stmt: "wait" "for" u=scenic_dynamic_duration { s.WaitFor(duration=u, LOCATIONS) }
+
+scenic_wait_until_stmt: "wait" "until" cond=expression { s.WaitUntil(cond=cond, LOCATIONS) }
```
dfremont:
```suggestion
scenic_wait_for_stmt: "wait" 'for' u=scenic_dynamic_duration { s.WaitFor(duration=u, LOCATIONS) }

scenic_wait_until_stmt: "wait" 'until' cond=expression { s.WaitUntil(cond=cond, LOCATIONS) }
```
Since `for` and `until` are hard keywords, they should be in single quotes. I guess Pegen is being permissive in also allowing them to appear in double quotes, but we should be consistent.

docs/reference/statements.rst
```diff
@@ -328,6 +328,18 @@ wait
 ----
 Take no actions this time step.
 
+.. _wait for {scalar} (seconds | steps):
```
dfremont, line 331: We need to add these to the Syntax Guide too.

dfremont (changes requested):
This all looks great, but we should also have tests that the new statements actually work properly in a behavior (in `tests/syntax/test_dynamics.py`, similar to `test_subbehavior_until`) and in the `compose` block of a modular scenario (in `tests/syntax/test_modular.py`).

dfremont (approved):
Looks great, thanks Lola!

Issue #386, opened 2025-09-13
Add driving actions for the MetaDrive simulator


kevinchang73:
### Describe the feature and motivation

Several driving actions (e.g., `SetReverseAction`, `SetSpeedAction`, `SetVelocityAction`, etc.) are not implemented for the MetaDrive simulator.

### Additional context

Here’s a sample code snippet: using the MetaDrive model results in a NotImplementedError, whereas the Newtonian model works as expected.

```
param map = localPath('../maps/Town05.xodr')
model scenic.simulators.metadrive.model
#model scenic.simulators.newtonian.driving_model

EGO_MODEL = "vehicle.toyota.prius"

behavior EgoBehavior():
    take SetSpeedAction(1)

ego = new Car at (0, 0),
    with blueprint EGO_MODEL,
    with behavior EgoBehavior()
```

### Issue Submission Checklist

- [x] I checked the open and closed issues, forum, etc. and have not found any solution

PR #387, opened 2025-09-18
Implement missing driving domain actions in MetaDrive interface

lola831:
### Description
Implements the missing driving domain actions in the MetaDrive interface (`SetPositionAction`, `SetVelocityAction`, `SetHandBrakeAction`, `SetReverseAction`) and makes control inputs persist until changed (as Scenic intends).

### Issue Link
[Issue 386](https://github.com/BerkeleyLearnVerify/Scenic/issues/386)

### Checklist

### Additional Notes
N/A

### DIFF
--- a/tests/simulators/metadrive/test_metadrive.py
+++ b/tests/simulators/metadrive/test_metadrive.py
@@ -1,11 +1,23 @@
 import math
 import os
+from urllib.error import URLError
 
 import numpy as np
 import pytest
 
 try:
     import metadrive
+    from metadrive.pull_asset import pull_asset
+
+    # Make sure MetaDrive assets are available.
+    # On CI, if the asset server is down or unreachable, skip this whole file.
+    try:
+        pull_asset(update=False)
+    except URLError as e:
+        pytest.skip(
+            f"MetaDrive assets could not be pulled ({e}); skipping MetaDrive tests",
+            allow_module_level=True,
+        )
 
     from scenic.simulators.metadrive import MetaDriveSimulator
 except ModuleNotFoundError:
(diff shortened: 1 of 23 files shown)


tests/simulators/metadrive/test_metadrive.py
```diff
@@ -237,3 +337,56 @@ def test_duplicate_sensor_names(getMetadriveSimulator):
...
+    outdir = tmp_path / "md_gifs"
+    outdir.mkdir()
+
+    simulator, openDrivePath, sumoPath = getMetadriveSimulator(
+        "Town01", render=True, render3D=False
+    )
+
+    simulator.screen_record = True
+    simulator.screen_record_filename = "render2d.gif"
+    simulator.screen_record_path = str(outdir)
```
dfremont: I'd rather use the supported API of `MetaDriveSimulator` than modify its internal attributes like this. Maybe you can have `getMetadriveSimulator` support arbitrary keyword arguments which get passed along to `MetaDriveSimulator` (rather than only `render` and `render3D`)?

tests/simulators/metadrive/test_metadrive.py
```diff
@@ -237,3 +337,56 @@ def test_duplicate_sensor_names(getMetadriveSimulator):
...
+        terminate after 2 steps
+    """
+    scenario = compileScenic(code, mode2D=True)
+    scene = sampleScene(scenario)
+    simulator.simulate(scene)
+
+    assert (outdir / "render2d.gif").exists()
+
+
+def test_follow_lane(getMetadriveSimulator):
```
dfremont, line 441: Was there a specific problem with `FollowLaneBehavior` in MetaDrive which prompted adding this test? Just want to make sure we're testing for whatever the bad behavior was.

dfremont (changes requested):
Looks good, just a couple minor comments.

tests/simulators/metadrive/test_metadrive.py
lola831: @dfremont Good point. I updated `getMetadriveSimulator` to forward `**kwargs` to `MetaDriveSimulator`, and the GIF test now passes the `screen_record` options via the constructor instead of mutating attributes.

dfremont (approved):
Looks good, thanks Lola!

dfremont (approved):

PR #390, opened 2025-09-25
add test to ensure obj == self inside behavior over simulation().objects (dynamic object proxy)

lola831:
### Description
Adds `test_obj_equals_self_inside_behavior` to verify `obj == self` for the object with the behavior when iterating `simulation().objects`. Covers prior cases where dynamic object proxies caused `self` misidentification.

### Issue Link
N/A

### Checklist

### Additional Notes
N/A


lola831:
@dfremont Now the test also checks `obj is self`

PR #401, opened 2025-10-21
Generate CARLA blueprints dynamically

lola831:
### Description
This PR generates CARLA blueprints dynamically (per CARLA version).

-`snapshot_blueprints.py`: connects to CARLA, collects blueprint IDs and bounding-box dimensions for the running server version, categorizes them for Scenic, and writes `tools/carla/snapshots/blueprints_<VERSION>.json`.

-`make_blueprints.py`: reads all snapshot JSONs, sorts/normalizes them, and regenerates `src/scenic/simulators/carla/blueprints.py`.

-Generated `blueprints.py`: selects the best-matching snapshot for the installed CARLA version and exposes ids, dims, and helper functions.

-Handles versions that lack certain categories (e.g., CARLA 0.10.0 has no bicycles).

-Uses measured dimensions when available; otherwise falls back to defaults.

-Adds `vanModels` and `busModels` categories based on CARLA’s `base_type`.

### Issue Link
N/A

### Checklist

### Additional Notes
N/A


src/scenic/simulators/carla/model.scenic
```diff
@@ -122,6 +122,12 @@ class CarlaActor(DrivingObject):
         carlaActor (dynamic): Set during simulations to the ``carla.Actor`` representing this
             object.
         blueprint (str): Identifier of the CARLA blueprint specifying the type of object.
+        defaultWidth (float): Default width to use if Scenic has no recorded dimensions for this blueprint. 
+        defaultLength (float): Default length to use if Scenic has no recorded dimensions for this blueprint. 
+        defaultHeight (float): Default height to use if Scenic has no recorded dimensions for this blueprint. 
+        width (float): Width for this blueprint; uses Scenic's recorded dimensions when available, otherwise ``defaultWidth``.
```
dfremont:
```suggestion
        width (float): Width for this object; uses the value from ``blueprint`` when available, otherwise ``defaultWidth``. 
```
Ditto for `length` and `height`.

tools/carla/make_blueprints.py
```diff
@@ -0,0 +1,178 @@
...
+}
+'''
+
+
+FOOTER = '''\
+ids = _pick(_IDS, _CARLA_VER)
+dims = _pick(_DIMS, _CARLA_VER)
+
+
+def any_in(category):
```
dfremont: I think this name is confusing. How about something like `blueprintsInCategory`?

src/scenic/simulators/carla/model.scenic
```diff
@@ -44,7 +44,7 @@ Global Parameters:
 import pathlib
 from scenic.domains.driving.model import *
 
-import scenic.simulators.carla.blueprints as blueprints
+import scenic.simulators.carla.blueprints as bp
```
dfremont: I think the old `blueprints` is more understandable.

src/scenic/simulators/carla/model.scenic
```diff
@@ -192,32 +206,45 @@ class NPCCar(Car):  # no distinction between these in CARLA
...
+    blueprint: Uniform(*bp.any_in("bus"))
+    defaultWidth: 4 
+    defaultLength: 10
+    defaultHeight: 4
+
+class Pedestrian(CarlaActor, Pedestrian, Walks, _CarlaPedestrian):
     """A pedestrian.
 
     The default ``blueprint`` (see `CarlaActor`) is a uniform distribution over the
     blueprints listed in :obj:`scenic.simulators.carla.blueprints.walkerModels`.
```
dfremont: Need to fix the docstring here, since the location of the blueprints has changed.

src/scenic/simulators/carla/model.scenic
```diff
@@ -192,32 +206,45 @@ class NPCCar(Car):  # no distinction between these in CARLA
     pass
 
 class Bicycle(Vehicle):
-    width: 1
-    length: 2
-    blueprint: Uniform(*blueprints.bicycleModels)
```
dfremont, line 197: For backwards compatibility, can we continue to define `bicycleModels`, etc. in the `blueprints` module? People might be using them, and they are documented. I think keeping those and also providing the new function to look up by the name of the category would make sense.

tools/carla/make_blueprints.py
```diff
@@ -0,0 +1,178 @@
...
+def _verkey(s: str):
+    parts = [int(p) for p in s.split(".")]
+    parts += [0] * (3 - len(parts))
+    return tuple(parts[:3])
+
+
+def load_versions():
+    files = list(SNAPSHOT_DIR.glob(SNAPSHOT_PATTERN))
+    if not files:
+        raise SystemExit("No input JSONs found.")
```
dfremont: This looks odd to me. I'd just use `sys.exit()`, which does the same thing.

tools/carla/make_blueprints.py
```diff
@@ -0,0 +1,178 @@
...
+
+    entries = []
+    for jf in files:
+        obj = json.loads(jf.read_text(encoding="utf-8"))
+        ver = str(obj["server_version"])
+        entries.append((ver, {"ids": obj.get("ids", {}), "dims": obj.get("dims", {})}))
```
dfremont: Why the `get()`s here? Aren't the files malformed if either key is missing?

tools/carla/snapshot_blueprints.py
```diff
@@ -0,0 +1,276 @@
...
+        add_group(walker_bps, lambda bp: WALKER_CATEGORY)
+        add_group(prop_bps, categorize_prop)
+
+        out_obj = {
+            "server_version": server_version,
+            "ids": ids,
+            "dims": dims,
+        }
+        out_path = SNAPSHOT_DIR / f"blueprints_{server_version}.json"
+        with open(out_path, "w", encoding="utf-8") as f:
```
dfremont: Since these will sit around every time someone clones the repo, let's compress them (if we need to inspect the files for debugging, it's easy enough to double-click on them). You can just use `gzip.open` here and in `make_blueprints.py`.

dfremont (changes requested):
Looks good in general. Besides some minor comments above, my only suggestion is that since `blueprints.py` is now very long and difficult to read, let's try splitting it into two files. The non-autogenerated part can stay in `blueprints.py`, but let's move the giant `_IDS` and `_DIMS` dictionaries into a separate `_blueprintData.py` module and then do `from _blueprintData import _IDS, _DIMS`. The `make_blueprints.py` script can then just generate `_blueprintData.py`.

src/scenic/simulators/carla/model.scenic
lola831: Updated docstrings for width/length/height as suggested.

tools/carla/make_blueprints.py
lola831: Renamed to `blueprintsInCategory`

src/scenic/simulators/carla/model.scenic
lola831: Updated the Car and Pedestrian docstrings to reflect the new `blueprintsInCategory`

src/scenic/simulators/carla/model.scenic
lola831: Re-added the category lists (`bicycleModels`, etc.) to the blueprints module for backwards compatibility, and kept the new lookup helper.

tools/carla/make_blueprints.py
lola831: Changed this to use `sys.exit()`.

tools/carla/snapshot_blueprints.py
lola831: Snapshots are now written and read via `gzip.open` (`blueprints_*.json.gz`).

tools/carla/make_blueprints.py
lola831: Good point. These snapshot files should always include both `"ids"` and `"dims"` so I switched to direct indexing instead of `get()`.

PR #402, opened 2025-10-21
Add tests for multiple assignments and unpacking

lola831:
### Description
Adds tests for multiple assignments and unpacking.

### Issue Link
N/A

### Checklist

### Additional Notes
N/A

### DIFF
--- a/tests/syntax/test_assignments.py
+++ b/tests/syntax/test_assignments.py
@@ -2,7 +2,7 @@
 
 import pytest
 
-from tests.utils import sampleParamPFrom
+from tests.utils import compileScenic, sampleParamP, sampleParamPFrom
 
 
 def test_multiple_assignment_swap():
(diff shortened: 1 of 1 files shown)


dfremont (changes requested):
The essence of the tests looks good, but let's simplify them by not using dynamic simulations (just top-level code rather than behaviors). You can use `sampleParamPFrom` to get simple values out if needed.

lola831:
@dfremont  I’ve simplified the tests to use top-level code rather than behaviors.

tests/syntax/test_assignments.py
```diff
@@ -0,0 +1,54 @@
...
+    assert p == (3, 4)
+
+
+def test_unpack_from_distribution_pair():
+    code = """
+        pair = (Uniform(0, 1), Uniform(0, 1))
+        x, y = pair
+        param p = (x, y)
+    """
+    outs = [sampleParamPFrom(code) for _ in range(20)]
```
dfremont: Please use `sampleParamP` rather than `sampleParamPFrom` to avoid recompiling the scenario every time.

Also please increase the number of repetitions to 30 to decrease the chance of a false negative.

tests/syntax/test_assignments.py
```diff
@@ -0,0 +1,54 @@
...
+    assert len(set(outs)) > 1
+    for x, y in outs:
+        assert 0 <= x <= 1
+        assert 0 <= y <= 1
```
dfremont:
```suggestion
        assert x == 0 or x == 1
        assert y == 0 or y == 1
```

PR #405, opened 2025-10-23
Add tests for subclassing Point or OrientedPoint instead of Object

lola831:
### Description
Adds two tests that assert TypeError when ego is a subclass of Point or OrientedPoint (must subclass Object).

### Issue Link
N/A

### Checklist

### Additional Notes
N/A

### DIFF
--- a/tests/syntax/test_basic.py
+++ b/tests/syntax/test_basic.py
@@ -63,7 +63,7 @@ def test_ego_subclass_point():
             class Foo(Point):
                 pass
             ego = new Foo
-        """
+            """
         )
 
 
@@ -74,7 +74,7 @@ def test_ego_subclass_orientedpoint():
             class Bar(OrientedPoint):
                 pass
             ego = new Bar
-        """
+            """
         )
 
 


tests/syntax/test_basic.py
```diff
@@ -56,6 +56,28 @@ def test_ego_nonobject():
...
 
 
+def test_ego_subclass_point():
+    with pytest.raises(TypeError):
+        compileScenic(
+            """
+            class Foo(Point):
+                pass
+            ego = new Foo
+        """
```
dfremont:
```suggestion
            """
```

tests/syntax/test_basic.py
```diff
@@ -56,6 +56,28 @@ def test_ego_nonobject():
...
+
+
+def test_ego_subclass_orientedpoint():
+    with pytest.raises(TypeError):
+        compileScenic(
+            """
+            class Bar(OrientedPoint):
+                pass
+            ego = new Bar
+        """
```
dfremont:
```suggestion
            """
```

PR #415, opened 2025-11-03
Add test for Scenic 'new' in list subexpression

lola831:
### Description
Add test for `new` used inside a list expression.

### Issue Link
N/A

### Checklist

### Additional Notes
N/A


Eric-Vin (changes requested):
Thanks for adding this test Lola! The only thing I would add is that I think we're trying to move towards triple-quoted strings in new tests. Once you swap that I'll go ahead and merge it into main.

PR #417, opened 2025-11-13
VerifAI Dynamic Sampling Compatibility

Eric-Vin:
### Description
Adds compatibility with VerifAI dynamic sampling.

### Issue Link
N/A

### Checklist

### Additional Notes
Merge with https://github.com/BerkeleyLearnVerify/VerifAI/pull/58

### DIFF
--- a/src/scenic/core/external_params.py
+++ b/src/scenic/core/external_params.py
@@ -270,6 +270,7 @@ class VerifaiSampler(ExternalSampler):
 
         self._lastSample = None
         self._lastDynamicSample = None
+        self._lastSimulation = None
         self._lastTime = -1
 
     def nextSample(self, feedback):
@@ -284,6 +285,10 @@ class VerifaiSampler(ExternalSampler):
         import scenic.syntax.veneer as veneer
 
         assert veneer.currentSimulation is not None
+
+        if self._lastSimulation is not veneer.currentSimulation:
+            self._lastSimulation = veneer.currentSimulation
+            self._lastTime = -1
 
         if veneer.currentSimulation.currentTime > self._lastTime:
             feedback = veneer.currentSimulation
--- a/src/scenic/core/external_params.py
+++ b/src/scenic/core/external_params.py
@@ -200,7 +200,7 @@ class VerifaiSampler(ExternalSampler):
 
         if timeBound == 0 and any(param.timeSeries for param in self.params):
             warnings.warn(
-                "TimeSeries external parameter used by no time bound specified "
+                "TimeSeries external parameter used but no time bound specified "
                 "(Did you provide `maxSteps` when creating ScenicSampler?)."
             )
 
@@ -358,7 +358,7 @@ class TimeSeriesParameter:
 
 def TimeSeries(param):
     if not isinstance(param, ExternalParameter):
-        raise ValueError("Cannot turn a non `ExternalParameter` into a time series")
+        raise TypeError("Cannot turn a non `ExternalParameter` into a time series")
 
     param.timeSeries = True
     return param
--- a/src/scenic/core/external_params.py
+++ b/src/scenic/core/external_params.py
@@ -252,14 +252,15 @@ class VerifaiSampler(ExternalSampler):
             self.rejectionFeedback = 1
         self.cachedSample = None
 
+        self._lastSample = None
+
     def nextSample(self, feedback):
-        return self.sampler.nextSample(feedback)
+        if feedback is not None:
+            assert self._lastSample is not None
+            self._lastSample.update(feedback)
 
-    def update(self, sample, info, rho):
-        self.sampler.update(sample, info, rho)
-
-    def getSample(self):
-        return self.sampler.getSample()
+        self._lastSample = self.sampler.getSample()
+        return self._lastSample.staticSample
 
     def valueFor(self, param):
         return getattr(self.cachedSample, self.nameForParam(param.index))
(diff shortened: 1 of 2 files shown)


src/scenic/core/external_params.py
```diff
@@ -193,11 +197,23 @@ def __init__(self, params, globalParams):
             param.index = index
             if param.probs is not None:
                 usingProbs = True
+
+        if timeBound == 0 and any(param.timeSeries for param in self.params):
+            warnings.warn(
+                "TimeSeries external parameter used by no time bound specified "
```
dfremont:
```suggestion
                "TimeSeries external parameter used but no time bound specified "
```

src/scenic/core/external_params.py
```diff
@@ -193,11 +197,23 @@ def __init__(self, params, globalParams):
...
+                "(Did you provide `maxSteps` when creating ScenicSampler?)."
```
dfremont: Couldn't somebody get this error without using `ScenicSampler`?

src/scenic/core/external_params.py
```diff
@@ -252,30 +268,57 @@ def __init__(self, params, globalParams):
...
 
 
-class ExternalParameter(Distribution):
+class ExternalParameter(Distribution, ABC):
     """A value determined by external code rather than Scenic's internal sampler."""
 
     def __init__(self):
         super().__init__()
         self.sampler = None
+        self.timeSeries = False
```
dfremont:
```suggestion
        self.isTimeSeries = False
```

src/scenic/core/external_params.py
```diff
@@ -289,6 +332,37 @@ def sampleGiven(self, value):
...
+                "Attempted `getSample` for a timeSeries property twice in one timestep."
+            )
+
+        self._lastTime = veneer.currentSimulation.currentTime
+        return self._callback()
+
+
+def TimeSeries(param):
+    if not isinstance(param, ExternalParameter):
+        raise ValueError("Cannot turn a non `ExternalParameter` into a time series")
```
dfremont:
```suggestion
        raise TypeError("Cannot turn a non `ExternalParameter` into a time series")
```

src/scenic/core/external_params.py
```diff
@@ -289,6 +332,37 @@ def sampleGiven(self, value):
         assert self.sampler is not None
         return self.sampler.valueFor(self)
 
+    @abstractmethod
+    def extractOutput(self, value):
```
dfremont, line 372: Need docstring, since authors of other kinds of external parameters will need to implement this.

dfremont (changes requested):
Looks fine overall, although I'm not sure what prompted the introduction of `extractOutput` -- maybe we can talk offline.

src/scenic/core/external_params.py
```diff
@@ -86,6 +86,15 @@
 in general may be approximated so that VerifAI can handle them -- see
 `VerifaiParameter.withPrior` for details.
 
+To set a time bound when using VerifAI's dynamic sampling, set the `timeBound`
```
dfremont:
```suggestion
To set a time bound when using VerifAI's dynamic sampling, set the ``timeBound``
```

src/scenic/core/external_params.py
```diff
@@ -86,6 +86,15 @@
...
+global parameter to value representing the upper bound on the number of timesteps
+the sampler should account for. For example::
+
+    param timeBound = 250
+
+This value can also be set directly in VerifAI via the `maxSteps` parameter to the
```
dfremont:
```suggestion
This value can also be set directly in VerifAI via the ``maxSteps`` parameter to the
```

src/scenic/core/external_params.py
```diff
@@ -86,6 +86,15 @@
...
+`ScenicSampler`.
```
dfremont: I think this reference will work if you add VerifAI to Intersphinx (see `intersphinx_mapping` in `docs/conf.py`).

src/scenic/core/external_params.py
```diff
@@ -193,11 +206,23 @@ def __init__(self, params, globalParams):
...
+                "(If using VerifAI’s ScenicSampler, set its maxSteps option)."
+            )
+
         space = verifai.features.FeatureSpace(
             {
-                self.nameForParam(index): verifai.features.Feature(param.domain)
+                self.nameForParam(index): (
+                    verifai.features.Feature(param.domain)
+                    if not param.isTimeSeries
+                    else verifai.features.TimeSeriesFeature(param.domain)
```
dfremont, line 231: This will break in an unhelpful way if someone tries using a time series with an old version of VerifAI installed. How about adding a loop before this which builds up a list of features, raising an error saying you need a newer version of VerifAI if `TimeSeriesFeature` doesn't exist, otherwise going on to build the feature space.

dfremont (changes requested):
Looking good -- just a last few minor suggestions.

src/scenic/core/external_params.py
dfremont: I'm just going to make this not be a link for the moment and we can improve it later.

dfremont (approved):
Let's merge this as-is and we can come back to the last two minor comments later.

dfremont:
Thanks for the fixes, Eric! This is looking good -- we can revisit the coverage issue once the new version of VerifAI is out.

PR #419, opened 2025-11-14
Improve test_docs pre-check with Intersphinx URLs

lola831:
### Description
This PR changes our pre-check for running `tests/test_docs.py` so that instead of just trying to connect to `docs.python.org` we now iterate through all configured Intersphinx URLs. Previously, CI would fail fairly often when one of these external inventories timed out during the Sphinx build, even though the pre-check had passed.

### Issue Link
N/A

### Checklist

### Additional Notes
N/A

### DIFF
--- a/tests/test_docs.py
+++ b/tests/test_docs.py
@@ -35,13 +35,17 @@ def _check_intersphinx_connectivity(timeout=10.0):
 
 @pytest.mark.slow
 def test_build_docs():
-    """Test that the documentation builds and doctests pass."""
+    """Test that the documentation builds, and run doctests.
+
+    We do this in a subprocess since the Sphinx configuration file activates the veneer
+    and has other side-effects that aren't reset afterward.
+    """
     _check_intersphinx_connectivity()
 
-    old_directory = os.getcwd()
+    oldDirectory = os.getcwd()
     try:
         os.chdir("docs")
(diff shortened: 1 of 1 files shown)


tests/test_docs.py
```diff
@@ -1,28 +1,47 @@
...
+            + "\n".join(problems)
+        )
+
+
 @pytest.mark.slow
 def test_build_docs():
-    """Test that the documentation builds, and run doctests.
-
-    We do this in a subprocess since the Sphinx configuration file activates the veneer
-    and has other side-effects that aren't reset afterward.
```
dfremont, line 15: Why delete this explanatory note? I don't think it's obvious.

tests/test_docs.py
```diff
@@ -1,28 +1,47 @@
...
-    """
-    try:
-        socket.getaddrinfo("docs.python.org", 80)
-    except OSError:
-        pytest.skip("cannot connect to python.org for Intersphinx")
+    """Test that the documentation builds and doctests pass."""
+    _check_intersphinx_connectivity()
 
-    oldDirectory = os.getcwd()
+    old_directory = os.getcwd()
```
dfremont: We have plenty of camel case in Scenic (it's the default for public APIs, following the style I used in the first implementation), so I wouldn't bother with changes like this.

Issue #32, opened 2021-05-23
Can Scenic  replace OpenSCENARIO?


AI-liu:
What is the different between Scenic and  OpenSCENARIO ?

PR #421, opened 2025-11-18
OpenScenarioXML Export

### LINKED ISSUE #334: Conversion between scenic and openscenario
Hello, 

Is there an existing implementation that converts scenic files to openscenario and vice-versa?

Thank you.

Eric-Vin:
### Description
Added experimental functionality to export the result of a Scenic simulation to OpenScenarioXML.

### Issue Link
#334 #32 

### Checklist

### Additional Notes
N/A


Eric-Vin:
An example scenario of a pedestrian crossing the road, occluded by a car, simulated in Metadrive and then exported to XOSC and replayed in [ESMINI](https://github.com/esmini/esmini).
Metadrive GIF:
pedestrianJaywalking_12

ESMINI Replay MP4:
https://github.com/user-attachments/assets/e7032e35-11f1-49e3-8634-6703ead6e9ba

docs/api.rst
```diff
@@ -173,6 +173,15 @@ it to the replay, but this greatly increases the size of the encoded simulation.
 	can return to one later for further analysis), but it is not guaranteed to be
 	compatible across major versions of Scenic.
 
+.. _xosc_export:
+
+OpenScenarioXML Export
+----------------------
+
+Scenic provides experimental support for exporting completed simulations via `toOpenScenario`.
```
dfremont, line 181: This function lives in an internal module whose documentation says "The functions in this module usually do not need to be used directly.", so I think it's confusing to link to it here. How about making a method `Scenario.toOpenScenario` parallel to `Scenario.simulationToBytes`?

src/scenic/core/serialization.py
```diff
@@ -392,3 +395,224 @@ def readStr(stream):
 
 
 Serializer.addCodec(str, writeStr, readStr)
+
+
+def toOpenScenario(
+    simulationResult,
+    scenario,
+    scene,
```
dfremont: Unless there's a particular reason to separate these, I think it would be simpler to just take a `Simulation` as input (which has a reference to the underlying `Scene`) and have this be a method on `Scenario`.

src/scenic/core/serialization.py
```diff
@@ -392,3 +395,224 @@ def readStr(stream):
...
+    xosc_paramdec = xosc.ParameterDeclarations()
+
+    # Extract map
+    if mapPath is None:
+        if "map" not in scenario.params:
+            raise ValueError(
+                "No `mapPath` provided and scenario does not have a `map` parameter defined."
+            )
+        mapPath = (
+            mapPath if mapPath is not None else os.path.abspath(scenario.params["map"])
```
dfremont: How can `mapPath` *not* be `None` here?

src/scenic/core/serialization.py
```diff
@@ -392,3 +395,224 @@ def readStr(stream):
...
+        )
+    xosc_road = xosc.RoadNetwork(roadfile=mapPath)
+
+    # Create entitities
+    entities = xosc.Entities()
+    xosc_objects = {}
+    for obj_i, obj in enumerate(scene.objects):
```
dfremont: Presumably this will break for dynamically-created objects. If XOSC doesn't support them we could raise an exception if they are present in the simulation, but at some point we should try to handle the case where all the objects are created dynamically in the first timestep.

src/scenic/core/serialization.py
```diff
@@ -392,3 +395,224 @@ def readStr(stream):
...
+        if hasattr(obj, "isVehicle") and obj.isVehicle:
```
dfremont:
```suggestion
        if getattr(obj, "isVehicle", False):
```

src/scenic/core/serialization.py
```diff
@@ -392,3 +395,224 @@ def readStr(stream):
...
+            obj_name = obj.name if hasattr(obj, "name") else f"Vehicle{obj_i}"
```
dfremont:
```suggestion
            obj_name = getattr(obj, "name", f"Vehicle{obj_i}")
```

src/scenic/core/serialization.py
```diff
@@ -392,3 +395,224 @@ def readStr(stream):
...
+                max_acceleration=obj.maxAcceleration,
+                max_deceleration=obj.maxDeceleration,
+                mass=None,
+                model3d=None,
+                max_acceleration_rate=None,
+                max_deceleration_rate=None,
+                role=None,
+            )
+        elif hasattr(obj, "isPedestrian") and obj.isPedestrian:
+            obj_name = obj.name if hasattr(obj, "name") else f"Pedestrian{obj_i}"
```
dfremont: Can simplify as above.

src/scenic/core/serialization.py
```diff
@@ -392,3 +395,224 @@ def readStr(stream):
...
+                obj.width,
+                obj.length,
+                obj.height,
+                0,
+                0,
+                0,
+            )
+            xosc_obj = xosc.Pedestrian(
+                name=obj_name,
+                mass=obj.mass,
```
dfremont: Is `mass` required for pedestrians but not for cars?

src/scenic/core/simulators.py
```diff
@@ -811,7 +811,19 @@ def currentState(self):
 
         The default implementation returns a tuple of the positions of all objects.
         """
-        return tuple(obj.position for obj in self.objects)
+
+        class SimulationState(tuple):
```
dfremont, line 820: Should update the docstring now that the `positions` and `orientations` attributes are available. While we're at it we should probably discourage overriding this method, except possibly by adding extra attributes to `SimulationState`.

src/scenic/core/serialization.py
```diff
@@ -392,3 +395,224 @@ def readStr(stream):
...
+    for obj_i, (obj, xosc_obj) in enumerate(xosc_objects.items()):
+        action_times = []
+        action_positions = []
+        for t, states in enumerate(simulationResult.trajectory):
+            action_positions.append(
+                pos_to_WorldPosition(
+                    obj, states.positions[obj_i], states.orientations[obj_i].yaw
+                )
+            )
+            action_times.append(simulationResult.timestep * t)
```
dfremont: `SimulationResult` doesn't have a `timestep` attribute, so I don't understand how your test can be passing.

dfremont: Oh, I see, it's actually a `Simulation` object. Better change the name of the parameter!

src/scenic/domains/driving/model.scenic
```diff
@@ -282,6 +286,24 @@ class Vehicle(DrivingObject):
         color (:obj:`Color` or RGB tuple): Color of the vehicle. The default value is a
             distribution derived from car color popularity statistics; see
             :obj:`Color.defaultCarColor`.
+        wheelbase: The distance between the front and rear axles of the vehicle. Default value is 0.6
+            times the length of the vehicle.
+        maxSteeringAngle: The maximum steering angle of the vehicle. The full steering range would be
+            two times this value, going from (-maxSteeringAngle, maxSteeringAngle). Default value
+            30 degrees.
```
dfremont: Did you mean 35 degrees, as in the code?

src/scenic/domains/driving/model.scenic
```diff
@@ -282,6 +286,24 @@ class Vehicle(DrivingObject):
...
+        wheelDiameter: The diameter of the *entire* wheel (including the tire). Default value is 0.7 meters.
+        trackWidth: Distance between the vehicle's wheels when pointed straight ahead. Default value
+            is 0.85 times the width of the vehicle.
+        groundClearance: Default value is half the wheel diameter.
```
dfremont: Doesn't actually say what this means.

tests/core/test_serialization.py
```diff
@@ -13,8 +13,14 @@
...
 
-from scenic.core.serialization import SerializationError, Serializer, deterministicHash
+from scenic.core.serialization import (
+    SerializationError,
+    Serializer,
+    deterministicHash,
+    toOpenScenario,
+)
 from scenic.core.simulators import DivergenceError, DummySimulator
+from tests.simulators.metadrive.test_metadrive import getMetadriveSimulator
```
dfremont: Can we use the Newtonian simulator instead? I think this would be the first `core` test which uses a non-built-in simulator, and we should be able to do this test without the optional MetaDrive dependency.

tests/core/test_serialization.py
```diff
@@ -507,3 +513,45 @@ class Foo:
...
+
+        parkedCar = new Car right of spot by 1, with regionContainedIn None
+
+        require distance from ego to parkedCar > 30
+
+        new Pedestrian ahead of parkedCar by 3,
+            facing 90 deg relative to parkedCar,
+            with behavior CrossRoad()
+
+        terminate after 30 seconds
```
dfremont: I'd make the simulation much shorter, just a few steps, so that the test runs faster. I also don't think there's any particular need to use a complex scenario here, and the test would be easier to read if the scenario just had the minimum stuff necessary (say a couple cars, one with `FollowLaneBehavior`, plus a pedestrian with an initial velocity).

dfremont (changes requested):
Looks good overall, just one issue (how to handle/ban dynamically spawned objects) and a bunch of minor comments above.

src/scenic/core/serialization.py
```diff
@@ -392,3 +395,224 @@ def readStr(stream):
...
+            z=state_position.z,
+            h=state_orientation,
+        )
+
+    # Initial states
+    init = xosc.Init()
+
+    for obj, xosc_obj in xosc_objects.items():
+        obj_init_action = xosc.TeleportAction(
+            pos_to_WorldPosition(obj, obj.position, obj.yaw)
```
Eric-Vin: Should reference `obj.heading` instead of `obj.yaw`.

src/scenic/core/serialization.py
```diff
@@ -392,3 +395,224 @@ def readStr(stream):
...
+        ),
+    )
+
+    for obj_i, (obj, xosc_obj) in enumerate(xosc_objects.items()):
+        action_times = []
+        action_positions = []
+        for t, states in enumerate(simulationResult.trajectory):
+            action_positions.append(
+                pos_to_WorldPosition(
+                    obj, states.positions[obj_i], states.orientations[obj_i].yaw
```
Eric-Vin: Should reference `states.orientations[obj_i].heading` instead of `states.orientations[obj_i].yaw`

src/scenic/core/serialization.py
Eric-Vin: Surprisingly, yes:
- Pedestrian: https://github.com/pyoscx/scenariogeneration/blob/debb300ec25c14cf16aae390f399a7311bd209f1/scenariogeneration/xosc/entities.py#L435
- Vehicles: https://github.com/pyoscx/scenariogeneration/blob/debb300ec25c14cf16aae390f399a7311bd209f1/scenariogeneration/xosc/entities.py#L863

And this appears supported by the spec. No idea why though.

src/scenic/core/serialization.py
Eric-Vin: I thought about this, but this is really a function specific to driving domain scenarios. Maybe we should move it there?

tests/domains/driving/test_driving.py
```diff
@@ -229,3 +231,69 @@ def test_invalid_road_scenario(cached_maps):
             param foo = ego.lane
             """,
         )
+
+
+def test_xosc_export(getAssetPath):
+    simulator = NewtonianSimulator("Town01")
+    code = f"""
```
dfremont: No need for this to be an f-string (ditto below).

tests/domains/driving/test_driving.py
```diff
@@ -229,3 +231,69 @@ def test_invalid_road_scenario(cached_maps):
...
+                ego = new Car with behavior FollowLaneBehavior()
+
+                immobileCar = new Car visible
+
+                behavior Walk():
+                    take SetWalkingSpeedAction(1)
+                
+                new Pedestrian behind ego by 5,
+                    with regionContainedIn None,
+                    with behavior Walk()
```
dfremont: We don't need any of this for the test, so I'd drop it to make the test easier to understand.

tests/utils.py
```diff
@@ -19,6 +19,7 @@
 
 from scenic import scenarioFromString
 from scenic.core.distributions import RejectionException
+from scenic.core.sample_checking import SampleChecker
```
dfremont: Unnecessary.

tests/core/test_serialization.py
```diff
@@ -46,7 +44,7 @@ def skeleton(scene):
     return {"objects": objs, "params": scene.params}
 
 
-def assertSceneEquivalence(scene1, scene2, ignoreDynamics=False, ignoreConstProps=False):
+def assertSceneEquivalence(scene1, scene2, ignoreDynamics=False, ignoreScenario=False):
```
dfremont: Let's just merge `ignoreDynamics` into `ignoreScenario`: I don't think there's any case where we would want one but not the other (and it makes sense if we're ignoring the scenario to also ignore the `DynamicScenario`).

dfremont (approved):
Looks good, thanks Eric!

Issue #393, opened 2025-10-03
Save information of parameters in .snet files


ethan10mak:
### Describe the feature and motivation

As of now, the parameters for scenic are not saved in the .snet files. So parameters like use2DMap, tolerance, etc. are not saved when caching. So if you change the parameters and you are using a cached map, it will use the previous parameters and not the new ones.

### Additional context

Changes should include the ability to save the parameters in the .snet files, and create new .snet files if there are no cached maps that have the same parameters. Scenic should also be able to pick the correct cached map if there is a map that exists with the same parameters.

### Issue Submission Checklist

- [x] I checked the open and closed issues, forum, etc. and have not found any solution

PR #425, opened 2025-12-04
Include map options in road network cache

lola831:
### Description
This PR makes the `.snet` cache also depend on map options.
I added a new `deterministicHash` helper and use it for both map options and scene serialization (`CompileOptions.hash`).
Since the scene options hash changed, I bumped the scene format version to 3.

### Issue Link
#393 

### Checklist

### Additional Notes
N/A

### DIFF
--- a/src/scenic/domains/driving/roads.py
+++ b/src/scenic/domains/driving/roads.py
@@ -957,6 +957,11 @@ class Network:
         # Build R-tree for faster lookup of roads, etc. at given points
         self._uidForIndex = tuple(self.elements)
         self._rtree = shapely.STRtree([elem.polygons for elem in self.elements.values()])
+        self._nominalDirElems = self.intersections + self.roads + self.shoulders
+        self._topLevelElements = (
+            self.intersections + self.roads + self.shoulders + self.sidewalks
+        )
+        # NOTE: Order matters here; elementAt resolves in this sequence.
 
     def _defaultRoadDirection(self, point):
         """Default value for the `roadDirection` vector field.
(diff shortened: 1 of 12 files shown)


src/scenic/domains/driving/roads.py
```diff
@@ -1049,21 +1050,34 @@ def fromFile(cls, path, useCache: bool = True, writeCache: bool = True, **kwargs
             data = f.read()
         digest = hashlib.blake2b(data).digest()
 
+        # Hash the map options as well so changing them invalidates the cache.
+        optionsDigest = deterministicHash(kwargs)
```
dfremont: We should specify an explicit length here so that if we change the default for this function later it won't break the code below, which depends on a specific length.

tests/domains/driving/test_network.py
```diff
@@ -226,3 +226,100 @@ def checkGroup(group):
...
+
+    # No digests supplied -> allowed (standalone .snet use case).
+    Network.fromPickle(path)
+
+
+def test_cache_regenerated_when_options_change(cached_maps):
+    """Changing map options should invalidate and rewrite the cached .snet file."""
+    # Get the temp copy of Town01 from cached_maps.
+    xodr_loc = cached_maps[str(mapFolder / "CARLA" / "Town01.xodr")]
+    xodr_path = Path(str(xodr_loc))
```
dfremont, line 332: For a future PR: we should switch from `tmpdir_factory` to `tmp_path_factory` so that we already have `Path` objects and don't need to do this conversion.

Issue #356, opened 2025-06-27
Syntax error wrongly attributed to valid `in` syntax


dfremont:
### System Details

1. Python 3.11.1
2. Scenic main branch
3. Linux and macOS

### Detailed Description

In the code below, which has invalid syntax on line 3, Scenic reports a syntax error on line 2, claiming the expression there is invalid (when in fact it is legal Scenic syntax).
```
Traceback (most recent call last; use -b to show Scenic internals):
  File "/Users/dfremont/Documents/Berkeley/repositories/BerkeleyLearnVerify/Scenic/test.scenic", line 3
    x in [0]
    ^^^^^^^^
ScenicSyntaxError: invalid syntax. Perhaps you forgot 'new'?
```
I think the `invalid_scenic_instance_creation` rule in the grammar is matching since `in` is also a specifier. We'll probably need to exclude `in` (and any other specifiers which are also binary operators) in that rule.

### Steps To Reproduce

Run `scenic` on the following code (reduced from original report directly to me by @Eric-Vin):
```scenic
x = 0
x in [0]
=
```

### Issue Submission Checklist

- [x] I am reporting an issue, not asking a question
- [x] I checked the open and closed issues, forum, etc. and have not found any solution
- [x] I have provided all necessary code, etc. to reproduce the issue

Issue #345, opened 2025-05-07
Syntax parser working incorrectly in some cases


dx6nxr:
### System Details

1. Python 3.10.10
2. Scenic 3.0.0
3. Win 11
4. Carla 0.9.15

### Detailed Description

1. The bug is as follows: If position specifiers are used incorrectly (in my case `offset left` ), then syntax checker shows a syntax error, but locates it in completely different part of the code, which is absolutely correct.
2. scenic -b --2d .\scenic_scenarios\file.scenic
3. ```Beginning scenario construction...
    Traceback (most recent call last):
      File "C:\Users\ivbu01\AppData\Local\Programs\Python\Python310\lib\runpy.py", line 196, in _run_module_as_main
        return _run_code(code, main_globals, None,
      File "C:\Users\ivbu01\AppData\Local\Programs\Python\Python310\lib\runpy.py", line 86, in _run_code
        exec(code, run_globals)
      File "C:\scenic310venv\Scripts\scenic.exe\__main__.py", line 4, in <module>
        from scenic.__main__ import dummy
      File "C:\Scenic\src\scenic\__main__.py", line 195, in <module>
        scenario = errors.callBeginningScenicTrace(
      File "C:\Scenic\src\scenic\core\errors.py", line 282, in callBeginningScenicTrace
        return func()
      File "C:\Scenic\src\scenic\__main__.py", line 196, in <lambda>
        lambda: translator.scenarioFromFile(
      File "C:\Scenic\src\scenic\syntax\translator.py", line 142, in scenarioFromFile
        return _scenarioFromStream(
      File "C:\Scenic\src\scenic\syntax\translator.py", line 172, in _scenarioFromStream
        compileStream(stream, namespace, compileOptions, filename)
      File "C:\Scenic\src\scenic\syntax\translator.py", line 289, in compileStream
        scenic_tree = parse_string(source, "exec", filename=filename)
      File "C:\Scenic\src\scenic\syntax\parser.py", line 107, in parse_string
        return parser.parse(mode if mode == "eval" else "file")
      File "C:\Scenic\src\scenic\syntax\parser.py", line 149, in parse
        res = getattr(self, rule)()
      File "C:\scenic310venv\lib\site-packages\pegen\parser.py", line 66, in memoize_wrapper
        tree = method(self, *args)
      File "C:\Scenic\src\scenic\syntax\parser.py", line 586, in file
        (a := self.statements(),)
[...]

PR #426, opened 2025-12-15
Fix incorrect “forgot new” errors caused by operator-like specifiers

lola831:
### Description
This PR fixes the parser issue where valid expressions like` x in [0]` or complex behavior conditions were incorrectly matched by `invalid_scenic_instance_creation`, causing Scenic to raise a “Perhaps you forgot new?” error at the wrong line. The problem was that operator-like specifiers (`in`, `at`, `offset by`, `offset along`) were treated the same as normal Scenic instance specifiers. I split these into two groups and restricted the “forgot new” rule to non-operator specifiers only. This resolves both #345 and #356: valid expressions are no longer misinterpreted, and incorrect instance specifiers are now reported at the correct location.

### Issue Link
[# 345](https://github.com/BerkeleyLearnVerify/Scenic/issues/345)
[# 356](https://github.com/BerkeleyLearnVerify/Scenic/issues/356)

### Checklist

### DIFF
--- a/src/scenic/syntax/scenic.gram
+++ b/src/scenic/syntax/scenic.gram
@@ -1721,18 +1721,23 @@ scenic_specifiers: ss=','.scenic_specifier+ { ss }
 scenic_specifier:
     | scenic_valid_specifier
     | invalid_scenic_specifier
+# Split into non-operator-like vs operator-like instance specifiers.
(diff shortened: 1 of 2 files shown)


dfremont (approved):
Nice fix, thanks Lola!

PR #431, opened 2025-12-18
Fix MetaDrive dependency for 3.1.0 release

lola831:
### Description
Reverts the MetaDrive extra to a version-based PyPI dependency, gates MetaDrive under `test-full` on Python 3.12+, and updates the docs with installation notes (Python 3.12+, macOS 3D, braking), and adjusts MetaDrive tests to xfail the 0.4.3 braking/handbrake issue.

### Checklist

### DIFF
--- a/docs/simulators.rst
+++ b/docs/simulators.rst
@@ -29,8 +29,6 @@ You can install it with:
 
     python -m pip install scenic[metadrive]
 
-Scenic supports both 2D and 3D rendering modes for MetaDrive simulations.
-
 .. note::
 
    MetaDrive **0.4.3** (the current PyPI release) does **not** support Python 3.12/3.13.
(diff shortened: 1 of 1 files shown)


docs/simulators.rst
```diff
@@ -31,6 +31,19 @@ You can install it with:
 
 Scenic supports both 2D and 3D rendering modes for MetaDrive simulations.
 
+.. note::
```
dfremont, line 32: Let's move this to right after the code block (the line about 2D/3D rendering should come afterward).

docs/simulators.rst
```diff
@@ -31,6 +31,19 @@ You can install it with:
...
+   MetaDrive **0.4.3** (the current PyPI release) does **not** support Python 3.12/3.13.
+   It also has known issues on macOS Apple Silicon (M-series) with 3D rendering and a
+   braking issue where vehicles may not come to a complete stop.
+
+   To use Python 3.12+ **and** get the fixes for the macOS/braking issues, install
+   MetaDrive from the GitHub repo:
+
+   .. code-block:: console
+
+      python -m pip install "metadrive-simulator @ git+https://github.com/metadriverse/metadrive.git@main"
```
dfremont, line 43: I think we need to add `sumolib` here too, since if `scenic[metadrive]` failed to install users may not have `sumolib`, which we need.

PR #452, opened 2026-04-14
Fix accessing ego and workspace in modular scenarios

dfremont:
### Description

When the `ego` and `workspace` special variables were assigned at the top level, they were not properly inherited by modular scenarios defined in the same file. Thanks to @Kv139 for the bug report!

### Issue Link

n/a

### Checklist


lola831 (approved):
Looks good.

PR #453, opened 2026-04-14
Improvements to warnings and tests

dfremont:
### Description

Three small improvements:
* Test that `str`, `float`, and `int` can be used properly in `isinstance` even though we handle them specially.
* Fix AST initializer warnings on Python 3.13 (e.g. `Record.__init__ missing 1 required positional argument`).
* Handle gimbal lock warnings better when a version of SciPy with the new option I added to suppress them is available.

(These are unrelated and probably should have been in separate PRs, but the first two turned out to already be in the branch I used for the third, although not yet merged. Sorry about that!)

### Issue Link

n/a

### Checklist


dfremont:
The coverage failure is because of code which only runs on Python versions other than the one used by our coverage workflow. I tested that code with `tox`, and you can also see it's working in the other CI runs.

lola831 (approved):
Tested locally and looks good.

PR #422, opened 2025-11-21
[WIP] Scenic Parallelization and Simulator Stepping

Eric-Vin:
### Description
Adds parallelization to scene sampling and simulation.

### Issue Link
N/A

### Checklist
- [X] I have tested the changes locally via `pytest` and/or other means
- [X] I have added or updated relevant documentation
- [X] I have autoformatted the code with black and isort
- [X] I have added test cases (if applicable)

### Additional Notes
N/A

PR #454, opened 2026-04-15
Add scenic.setSeed() public API

KE7:
## Summary

Exposes the seeding logic that the `-s`/`--seed` CLI option already uses as a module-level function, so Scenic scenarios can be made deterministic from Python code without touching `random` / `numpy.random` directly.

### Motivation

`docs/api.rst` currently teaches users to call `random.seed(n)` directly to get deterministic generation:

```python
import random, scenic
random.seed(12345)
scenario = scenic.scenarioFromString('ego = new Object with foo Range(0, 5)')
scene, numIterations = scenario.generate()
```

This is both:

1. **Incomplete.** It only seeds Python's `random`; it does not seed `numpy.random`, so any distribution backed by numpy's RNG (and anything that goes through numpy downstream) still drifts across runs.
2. **An implementation detail.** It asks users to know *which* RNGs Scenic draws from internally.

Meanwhile, `scenic/__main__.py` already does the right thing for the CLI:

```python
# src/scenic/__main__.py (before this PR)
if args.seed is not None:
    ...
    random.seed(args.seed)
    numpy.random.seed(args.seed)
```

This PR turns that into a named API — `scenic.setSeed(seed)` — following the same pattern as the existing `scenic.setDebuggingOptions`, which was added in commit 9b9c8509 ("unify control of debugging options; more docs"). Same structural precedent: a `scenic.set*(...)` public function wrapping CLI-equivalent behavior, documented with `.. autofunction::` in `api.rst` and cross-referenced from `options.rst`.

I also hit this flakiness myself while using Scenic in a downstream project — having to remember "seed both random and numpy" kept being the wrong answer when constraints got tight enough to make rejection sampling probabilistic.

## Changes

| File | Change |
|---|---|
| `src/scenic/__init__.py` | Add `setSeed(seed)`, documented, that seeds both `random` and `numpy.random`. |
| `src/scenic/__main__.py` | Replace the inline `random.seed(args.seed); numpy.random.seed(args.seed)` block with `scenic.setSeed(args.seed)`. Drop the now-unused `import random` / `import numpy` at the top. Single source of truth for how `--seed` works. |
| `tests/syntax/test_basic.py` | Add `test_setSeed` next to `test_verbose`, mirroring the existing `tests/test_main.py::test_seed` CLI-level structure (same seed → same sample; different seed → different sample). Uses the existing `sampleParamPFrom` helper. |
| `docs/api.rst` | Document the new API with `.. autofunction::` and update the testcode example to use `scenic.setSeed(12345)` instead of `random.seed(12345)`. Expected output is unchanged — verified locally that `Range(0, 5)` produces the same `foo = 2.083099362726706` either way. |
| `docs/options.rst` | Add the standard "can be configured from the Python API using \`scenic.setSeed\`" cross-reference to the `-s`/`--seed` entry, matching the existing pattern for `scenic.setDebuggingOptions`. |

Net: **+45 / -7 across 5 files.**

## Non-goals / deliberately out of scope

- **`seed=` kwarg on `Scenario.generate()`.** That's a bigger design question (per-call scoping vs. global) and would stall review. This PR is strictly CLI-parity.
- **Wrapping with a fresh `random.Random()` instance.** Keeping the global-RNG semantics matches the CLI exactly and means this PR is pure refactor + API exposure, not a behavior change.

## Test plan

Ran locally against this branch (Linux, Python 3.11, `pip install -e ".[test-full]"` in a fresh venv):

- [x] `black --check` / `isort --check-only` on all touched `.py` files — clean (versions match `.github/workflows/check-formatting.yml`: black 25.1.0, isort 5.12.0)
- [x] `pytest --fast --no-graphics` (matches `.github/workflows/run-tests.yml`): **1444 passed, 538 skipped, 3 xfailed, 0 failures**
- [x] `pytest --no-graphics` (slow suite, including `tests/test_docs.py::test_build_docs` which builds the Sphinx docs via `make html` with `-W`): passes after the `docs/api.rst` + `docs/options.rst` updates
- [x] `tests/test_main.py::test_seed` (CLI-level, verifies the `__main__.py` refactor): passes
- [x] `tests/syntax/test_basic.py::test_setSeed` (new): passes — determinism and variation-across-seeds both hold

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -28,10 +28,16 @@ the sampled values for all the global parameters and objects in the scene from t
 	import os
 	os.chdir('..')
 
+To make sampling deterministic, seed Scenic's random number generators
+before calling `Scenario.generate` using `scenic.setSeed` (the programmatic
+equivalent of the :option:`--seed` command-line option):
+
+.. autofunction:: scenic.setSeed
+
 .. testcode::
 
-	import random, scenic
-	random.seed(12345)
+	import scenic
+	scenic.setSeed(12345)
 	scenario = scenic.scenarioFromString('ego = new Object with foo Range(0, 5)')
 	scene, numIterations = scenario.generate()
 	print(f'ego has foo = {scene.egoObject.foo}')
--- a/docs/options.rst
+++ b/docs/options.rst
@@ -48,6 +48,8 @@ General Scenario Control
 	(although :mod:`random` and :mod:`numpy.random` should not be used in place of
 	Scenic's own sampling constructs in Scenic code).
 
+	This option can be configured from the Python API using `scenic.setSeed`.
+
 .. option:: --scenario <scenario>
 
 	If the given Scenic file defines multiple scenarios, select which one to run.
--- a/src/scenic/__init__.py
+++ b/src/scenic/__init__.py
@@ -1,8 +1,31 @@
 """A compiler and scene generator for the Scenic scenario description language."""
 
+import random as _random
+
+import numpy as _numpy
+
 import scenic.core.errors as _errors
 from scenic.core.errors import setDebuggingOptions
 from scenic.syntax.translator import scenarioFromFile, scenarioFromString
 
 _errors.showInternalBacktrace = False  # see comment in errors module
 del _errors
+
+
+def setSeed(seed):
+    """Seed the random number generators Scenic uses for sampling.
+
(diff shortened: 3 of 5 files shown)


Eric-Vin:
Closed as feature is already being added in #422.

PR #456, opened 2026-04-15
Use commit SHAs instead of refs in benchmark workflow

lola831:
### Description
Updated the benchmark workflow to use commit SHAs instead of refs when checking out the base and compare versions. This makes the benchmark run against specific commits rather than moving branch refs.

### Checklist
- [ ] I have tested the changes locally via `pytest` and/or other means


dfremont (approved):
Good idea, thanks Lola!

PR #458, opened 2026-04-18
Add Fix for the Size-based `containsRegion` Heuristic

Eric-Vin:
### Description
This fixes the size-based fast path used in `containsRegion`. It previously triggered when tolerance is non-zero (but ignored tolerance) and was susceptible to small numerical errors. This fix disables it for cases with non-zero tolerance and adds a small fudge-factor to ensure the heuristic is robust to small numerical errors.

### Issue Link
N/A

### Checklist

### Additional Notes
N/A


dfremont (approved):
Thanks Eric!

PR #461, opened 2026-04-29
Harden simulator workflow with OIDC and SSM

lola831:
### Description
This PR updates the `run_simulators` workflow to use GitHub OIDC and AWS SSM instead of long-lived AWS access keys, SSH keys, and a GitHub access token. The workflow now assumes an AWS role with short-lived credentials, sends remote commands through SSM, and checks out the exact workflow commit by SHA. 

### Checklist


dfremont (approved):
Looks good, thanks Lola!

Issue #459, opened 2026-04-20
Scenic crashes on custom map


ipirwani-blip:
### System Details

1. Python version: 3.10.19
2. Scenic version: 3.1.0
3. Operation system: `Linux a990320-lcedt 6.17.0-14-generic #14~24.04.1-Ubuntu SMP PREEMPT_DYNAMIC Thu Jan 15 15:52:10 UTC 2 x86_64 x86_64 x86_64 GNU/Linux`
4. Simulator: CARLA 0.9.16

### Detailed Description

1. Scenic has a bug
2. `scenic --2d --model scenic.simulators.carla.model --simulate --count 1 --seed 12345 /home/user/carla-camera-placement/monolith/client/test_data/rag18.scenic -b`
3. Error log:
```
Using random seed = 12345
Beginning scenario construction...
/home/user/miniconda3/envs/carla-env/lib/python3.10/site-packages/pygame/pkgdata.py:25: UserWarning: pkg_resources is deprecated as an API. See https://setuptools.pypa.io/en/latest/pkg_resources.html. The pkg_resources package is slated for removal as early as 2025-11-30. Refrain from using this package or pin to Setuptools<81.
  from pkg_resources import resource_stream, resource_exists
pygame 2.6.1 (SDL 2.28.4, Python 3.10.19)
Hello from the pygame community. https://www.pygame.org/contribute.html
  Parsing OpenDRIVE file...
Traceback (most recent call last):
  File "/home/user/miniconda3/envs/carla-env/bin/scenic", line 3, in <module>
    from scenic.__main__ import dummy
  File "/home/user/Scenic_repo/src/scenic/__main__.py", line 195, in <module>
    scenario = errors.callBeginningScenicTrace(
  File "/home/user/Scenic_repo/src/scenic/core/errors.py", line 282, in callBeginningScenicTrace
    return func()
  File "/home/user/Scenic_repo/src/scenic/__main__.py", line 196, in <lambda>
    lambda: translator.scenarioFromFile(
  File "/home/user/Scenic_repo/src/scenic/syntax/translator.py", line 138, in scenarioFromFile
    return _scenarioFromStream(
  File "/home/user/Scenic_repo/src/scenic/syntax/translator.py", line 168, in _scenarioFromStream
    compileStream(stream, namespace, compileOptions, filename)
  File "/home/user/Scenic_repo/src/scenic/syntax/translator.py", line 317, in compileStream
    executeCodeIn(code, namespace)
[...]

PR #467, opened 2026-05-12
Handle inconsistent OpenDRIVE planView lengths

lola831:
### Description
This PR updates the OpenDRIVE parser so that when a `planView` geometry’s declared `length` is inconsistent with the next geometry’s `s` coordinate, Scenic warns instead of failing with `ValueError: planView of road ... has inconsistent length`. In that case, Scenic now uses the next geometry’s `s` coordinate as the effective boundary for the previous geometry,  allowing maps to continue loading. Curve construction is delayed until after this check so the curve object is created with the corrected length.

### Issue Link
[issue #459](https://github.com/BerkeleyLearnVerify/Scenic/issues/459) 

### Checklist

### DIFF
--- a/src/scenic/formats/opendrive/xodr_parser.py
+++ b/src/scenic/formats/opendrive/xodr_parser.py
@@ -212,6 +212,47 @@ class Line(Curve):
         return self.rel_to_abs((s, 0, s))
 
 
+def makeCurve(x0, y0, hdg, length, curve_elem):
+    if curve_elem.tag == "line":
+        return Line(x0, y0, hdg, length)
+    elif curve_elem.tag == "arc":
+        # Arc is clothoid of constant curvature.
+        curv = float(curve_elem.get("curvature"))
+        return Clothoid(x0, y0, hdg, length, curv, curv)
+    elif curve_elem.tag == "spiral":
+        curv0 = float(curve_elem.get("curvStart"))
+        curv1 = float(curve_elem.get("curvEnd"))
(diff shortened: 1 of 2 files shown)


src/scenic/formats/opendrive/xodr_parser.py
```diff
@@ -1434,82 +1480,58 @@ def parse(self, path):
...
-                    raise ValueError(
-                        f"planView of road {road.id_} has inconsistent length"
-                    )
+
                 if l < 0:
                     raise ValueError(f"planView of road {road.id_} is not in order")
-                elif l < 1e-6:
+
+                lastS0, x0, y0, hdg, length, curve_elem = lastCurve
+                if abs(length - l) > 1e-4:
```
dfremont:
```suggestion
                if abs(length - l) > 1e-3:
```
I think we can increase the threshold here: an error of 1 mm is unlikely to matter, and people will get fewer warnings this way.

dfremont (approved):
Looks great, thanks Lola!

ipirwani-blip:
Thank you! I presume this PR resolves some of the bugs but that there are other issues that are there, is that right?

dfremont:
@ipirwani-blip Yes, there are (at least) 2 other issues. The first is also a problem with the map (I'll post on the issue) but we can work around it. The second may actually be a bug in Scenic -- I'm looking into it but we probably won't be able to fix it until early next week.

PR #361, opened 2025-07-15
Support CARLA 0.10.0 (UE5)

lola831:
### Description
This PR adds support for CARLA 0.10.0 and keeps support for older 0.9.x versions.
- Update CARLA examples to use the only supported 0.10.0 map (Town10HD_Opt).
- Select blueprints based on the detected CARLA version.
- Revise docs to include 0.10.0 install steps and note known differences/issues (e.g. slower pedestrian physics).
- Update CARLA tests and GitHub Actions CI so simulator tests run under both 0.9.15 and 0.10.0.

### Issue Link
N/A

### Checklist
- [x] I have tested the changes locally via `pytest` and/or other means
- [x] I have added or updated relevant documentation
- [x] I have autoformatted the code with black and isort
- [x] I have added test cases (if applicable)

### Additional Notes
N/A

PR #474, opened 2026-06-01
Add CARLA 0.10.0 support

lola831:
### Description
This PR adds support for CARLA 0.10.0. 

### Checklist


lola831:
When we are ready to include CARLA 0.10.0 in the `run_simulators` CI workflow, the CARLA 0.10.0 folder on the AWS instance needs to be renamed to `carla0.10.0` so the workflow picks it up. 

I tested the CARLA simulator tests manually on the AWS instance, and CARLA 0.9.15, 0.9.16, and 0.10.0 are currently passing. One issue I ran into was that CARLA 0.9.15/0.9.16 would crash partway through `test_blueprints.py` when Town01 used spawn point `(2, -17)`, so I switched Town01 back to the older spawn point from `main`, `(369, -326)`, which appears to avoid the crash.

This PR supersedes the older CARLA 0.10.0 PR #361

PR #478, opened 2026-06-11
crosswalk-parsing fixed PR

aaravshar:
crosswalk parsing changes to xodr_parser and 1 scenario for pedestrian-vehicle interactions

### DIFF
--- a/examples/driving/crosswalkStraightThrough.scenic
+++ b/examples/driving/crosswalkStraightThrough.scenic
@@ -1,0 +1,87 @@
+""" Scenario Description
+A vehicle drives toward an intersection with the intent to go straight through
+it. The intersection has a pedestrian crosswalk on the side the vehicle is
+entering or exiting. A pedestrian waits next to the crosswalk. As the ego
+approaches, the pedestrian steps onto the crosswalk and walks across.
+The ego is expected to brake before reaching the pedestrian.
+
+To run this file using the MetaDrive simulator:
+    scenic examples/driving/crosswalkStraightThrough.scenic --2d --model scenic.simulators.metadrive.model --simulate
+
+To run this file using the Carla simulator:
+    scenic examples/driving/crosswalkStraightThrough.scenic --2d --model scenic.simulators.carla.model --simulate
+"""
+
+param map = localPath('../../assets/maps/CARLA/Town03.xodr')
+param carla_map = 'Town03'
+model scenic.domains.driving.model
+
+param time_step = 1.0/10
+
+# CONSTANTS
+EGO_SPEED = Range(7, 12)
+EGO_OFFSET = -1 * Range(25, 40)
+PED_SPEED = 1.2
+PED_THRESHOLD = Range(35, 55)
+
+# BEHAVIOR
+behavior EgoBehavior(target_speed=10, trajectory=None):
+    assert trajectory is not None
+    brakeIntensity = 0.7
+
+    try:
+        do FollowTrajectoryBehavior(target_speed=target_speed, trajectory=trajectory)
+        terminate
+
+    interrupt when withinDistanceToAnyPedestrians(self, 12):
+        take SetBrakeAction(brakeIntensity)
+
+behavior PedestrianCrossBehavior(threshold=20, walking_speed=1.2, walking_direction=0):
+    while (distance from self to ego) > threshold:
+        wait
+    take SetWalkingDirectionAction(walking_direction), SetWalkingSpeedAction(walking_speed)
+
+# GEOMETRY
+def orderedEnds(cr, m):
+    ends = [cr.centerline[0], cr.centerline[-1]]
+    ends.sort(key=lambda p: distance from p to m.startLane.centerline)
+    return ends
+
+straightManeuverCrossingTriples = []
+for inter in network.intersections:
+    for m in inter.maneuvers:
+        if m.type != ManeuverType.STRAIGHT:
+            continue
+        for cr in m.connectingLane.road.crossings:
+            near, far = orderedEnds(cr, m)
+            straightManeuverCrossingTriples.append((inter, m, cr, near, far))
+chosen = Uniform(*straightManeuverCrossingTriples)
+intersection = chosen[0]
+straight_maneuver = chosen[1]
+crossing = chosen[2]
+pedSpawn = chosen[3]
+pedEnd = chosen[4]
+
+startLane = straight_maneuver.startLane
+connectingLane = straight_maneuver.connectingLane
+endLane = straight_maneuver.endLane
+
+lane_traj = [startLane, connectingLane, endLane]
+intersection_edge = startLane.centerline[-1]
+egoStartPoint = new OrientedPoint at intersection_edge
+
+# --
+
+walkDirection = angle from pedSpawn to pedEnd
+pedStartPoint = new OrientedPoint at pedSpawn,
+                    facing walkDirection
+
+# PLACEMENT
+ego = new Car following roadDirection from egoStartPoint for EGO_OFFSET,
+        with behavior EgoBehavior(target_speed=EGO_SPEED, trajectory=lane_traj)
+
+ped = new Pedestrian at pedStartPoint,
+        with regionContainedIn None,
+        with parentOrientation 0,
+        with behavior PedestrianCrossBehavior(PED_THRESHOLD, PED_SPEED, walkDirection),
+        facing walkDirection
--- a/examples/driving/crosswalkStraightThrough.scenic
+++ b/examples/driving/crosswalkStraightThrough.scenic
@@ -49,10 +49,10 @@ def orderedEnds(cr, m):
 
 straightManeuverCrossingTriples = []
 for inter in network.intersections:
-    for m in inter.maneuvers:
-        if m.type != ManeuverType.STRAIGHT:
-            continue
-        for cr in m.connectingLane.road.crossings:
+    for cr in inter.crossings:
+        for m in cr.conflictingManeuvers:
+            if m.type != ManeuverType.STRAIGHT:
+                continue
             near, far = orderedEnds(cr, m)
             straightManeuverCrossingTriples.append((inter, m, cr, near, far))
 chosen = Uniform(*straightManeuverCrossingTriples)
--- a/examples/driving/crosswalkStraightThrough.scenic
+++ b/examples/driving/crosswalkStraightThrough.scenic
@@ -37,8 +37,7 @@ behavior EgoBehavior(target_speed=10, trajectory=None):
         take SetBrakeAction(brakeIntensity)
 
 behavior PedestrianCrossBehavior(threshold=20, walking_speed=1.2, walking_direction=0):
-    while (distance from self to ego) > threshold:
-        wait
+    wait until distance from self to ego <= threshold
     take SetWalkingDirectionAction(walking_direction), SetWalkingSpeedAction(walking_speed)
 
 # GEOMETRY
@@ -68,7 +67,6 @@ endLane = straight_maneuver.endLane
 
 lane_traj = [startLane, connectingLane, endLane]
 intersection_edge = startLane.centerline[-1]
-egoStartPoint = new OrientedPoint at intersection_edge
 
 # --
 
@@ -77,7 +75,7 @@ pedStartPoint = new OrientedPoint at pedSpawn,
                     facing walkDirection
 
 # PLACEMENT
-ego = new Car following roadDirection from egoStartPoint for EGO_OFFSET,
+ego = new Car following roadDirection from intersection_edge for EGO_OFFSET,
         with behavior EgoBehavior(target_speed=EGO_SPEED, trajectory=lane_traj)
 
 ped = new Pedestrian at pedStartPoint,
--- a/src/scenic/formats/opendrive/xodr_parser.py
+++ b/src/scenic/formats/opendrive/xodr_parser.py
@@ -2136,6 +2136,7 @@ class RoadMap:
             allRoads, seenRoads = [], set()
             allSignals, seenSignals = [], set()
             maneuversForLane = defaultdict(list)
+            junctionCrossings, seenConnectingRoads = [], set()
             for connection in junction.connections:
                 incomingID = connection.incoming_id
                 incomingRoad = mainRoads.get(incomingID)
@@ -2146,6 +2147,10 @@ class RoadMap:
                 connectingRoad = connectingRoads.get(connectingID)
                 if not connectingRoad:
                     continue  # connecting road has no drivable lanes; skip it
+
+                if connectingID not in seenConnectingRoads:
+                    seenConnectingRoads.add(connectingID)
+                    junctionCrossings.extend(connectingRoad.crossings)
 
                 for signal in connectingRoad.signals:
                     if signal.openDriveID not in seenSignals:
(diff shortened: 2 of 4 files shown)


src/scenic/formats/opendrive/xodr_parser.py
```diff
@@ -10,8 +10,10 @@
 import numpy as np
 from scipy.integrate import quad, solve_ivp
 from scipy.optimize import brentq
-from shapely.geometry import GeometryCollection, MultiPoint, MultiPolygon, Point, Polygon
-from shapely.ops import snap, unary_union
+from scipy.spatial.transform import Rotation as R
+from shapely.geometry import GeometryCollection, MultiPoint, MultiPolygon, Point, Polygon, LineString
```
Eric-Vin: May be a good idea to import `Point` as `ShapelyPoint` or something like that, since otherwise we could easily encounter a conflict with Scenic's `Point`.

src/scenic/formats/opendrive/xodr_parser.py
```diff
@@ -146,6 +153,18 @@ def point_at(self, s):
         pt = (s, self.poly.eval_at(u), s)
         return self.rel_to_abs(pt)
 
+    def heading_at(self, s):
+        root_func = lambda x: self.arclength(x) - s
+        u = float(brentq(root_func, 0, self.ubound))
+        dv_du = self.poly.grad_at(u)
+        local_heading = math.atan2(dv_du, 1)
+        global_heading = local_heading + self.hdg
+        while global_heading > math.pi:
```
Eric-Vin: You can probably just use `normalizeAngle` in `scenic.core.gometry` here.

src/scenic/formats/opendrive/xodr_parser.py
```diff
@@ -173,6 +192,20 @@ def point_at(self, s):
...
 
+    def heading_at(self, s):
+        root_func = lambda x: self.arclength(x) - s
+        p = float(brentq(root_func, 0, self.p_range + 1e-9))
+        p = min(p, self.p_range)
+        du_dp = self.u_poly.grad_at(p)
+        dv_dp = self.v_poly.grad_at(p)
+        local_heading = math.atan2(dv_dp, du_dp)
+        global_heading = local_heading + self.hdg
+        while global_heading > math.pi:
```
Eric-Vin: See above.

src/scenic/formats/opendrive/xodr_parser.py
```diff
@@ -215,6 +248,14 @@ def clothoid_ode(s, state):
             x, y, hdg = sol.y[:, -1]
             return (x, y, s)
 
+    def heading_at(self, s):
+        theta = (self.curv0 * s) + (0.5 * self.curve_rate * s * s) + self.hdg
+        while theta > math.pi:
```
Eric-Vin: See above.

src/scenic/formats/opendrive/xodr_parser.py
```diff
@@ -228,6 +269,14 @@ def __init__(self, x0, y0, hdg, length):
     def point_at(self, s):
         return self.rel_to_abs((s, 0, s))
 
+    def heading_at(self, s):
+        heading = self.hdg
+        while heading > math.pi:
```
Eric-Vin: See above.

examples/driving/crosswalkStraightThrough.scenic
```diff
@@ -0,0 +1,87 @@
...
+    try:
+        do FollowTrajectoryBehavior(target_speed=target_speed, trajectory=trajectory)
+        terminate
+
+    interrupt when withinDistanceToAnyPedestrians(self, 12):
+        take SetBrakeAction(brakeIntensity)
+
+behavior PedestrianCrossBehavior(threshold=20, walking_speed=1.2, walking_direction=0):
+    while (distance from self to ego) > threshold:
+        wait
```
dfremont:
```suggestion
    wait until distance from self to ego <= threshold
```

examples/driving/crosswalkStraightThrough.scenic
```diff
@@ -0,0 +1,87 @@
...
+pedSpawn = chosen[3]
+pedEnd = chosen[4]
+
+startLane = straight_maneuver.startLane
+connectingLane = straight_maneuver.connectingLane
+endLane = straight_maneuver.endLane
+
+lane_traj = [startLane, connectingLane, endLane]
+intersection_edge = startLane.centerline[-1]
+egoStartPoint = new OrientedPoint at intersection_edge
```
dfremont: I don't think an orientation is used here (and you aren't setting one), so you might as well just use `intersection_edge` rather than `egoStartPoint`. Also the name `egoStartPoint` is confusing since the ego doesn't actually start there. I think using `following roadDirection from intersection_edge for ...` below would be clearer.

src/scenic/formats/opendrive/xodr_parser.py
```diff
@@ -146,6 +161,14 @@ def point_at(self, s):
         pt = (s, self.poly.eval_at(u), s)
         return self.rel_to_abs(pt)
 
+    def heading_at(self, s):
+        root_func = lambda x: self.arclength(x) - s
+        u = float(brentq(root_func, 0, self.ubound))
+        dv_du = self.poly.grad_at(u)
+        local_heading = math.atan2(dv_du, 1)
```
dfremont:
```suggestion
        local_heading = math.atan(dv_du)
```

src/scenic/formats/opendrive/xodr_parser.py
```diff
@@ -920,6 +962,26 @@ def hasGap(edge1, edge2):
             id_ = f"road{self.id_}_{name}{direction}"
             return id_, union, centerline, leftEdge, rightEdge
 
+        def makeCrosswalk():
```
dfremont: Since this is only called once, and creates all the crosswalks for this road (not just one), I think there's no need for a helper function. I would just inline it.

src/scenic/formats/opendrive/xodr_parser.py
```diff
@@ -934,7 +996,7 @@ def makeSidewalk(sections, backward=False):
                 leftEdge=leftEdge,
                 rightEdge=rightEdge,
                 road=None,
-                crossings=(),  # TODO add crosswalks
+                crossings=(pedestrian_crossings),
```
dfremont: This should be a `tuple`.

Also, I think it's theoretically possible that you could have a crosswalk that connects to one of the two sidewalks but not the other, or a crosswalk connecting to no sidewalks. So it would be safer here to check that at least one of the endpoints of the crosswalk is close to the sidewalk (within the tolerance) before adding it.

src/scenic/formats/opendrive/xodr_parser.py
```diff
@@ -1191,7 +1253,7 @@ def getEdges(forward):
             backwardLanes=backwardGroup,
             sections=roadSections,
             signals=tuple(roadSignals),
-            crossings=(),  # TODO add these!
+            crossings=tuple(pedestrian_crossings),
```
dfremont, line 1257: The documentation for `Road.crossings` says they are ordered from the start to the end of the road. If I understand correctly they aren't in any particular order in OpenDRIVE, so I think you'll need to sort them by `s`-coordinate (probably good to do the sorting right after the `makeCrosswalk` loop, so that they'll be sorted when assigned to the sidewalks too).

src/scenic/formats/opendrive/xodr_parser.py
```diff
@@ -1226,9 +1288,214 @@ def getEdges(forward):
...
+        y = y_ref + t * math.cos(heading)
+        z = z_ref + zOffset
+        return (x, y, 0)
+
+    def uv_to_xyz(self, s, t, zOffset, u, v, z_local, hdg, pitch, roll):
+        (x0, y0, z0) = self.st_to_xyz(s, t, zOffset)
+        (_, _, _), road_heading = self.xyz_heading_at_s(s)
+        yaw = hdg + road_heading
+
+        r = R.from_euler('zyx', [yaw, pitch, roll], degrees=False)
```
dfremont: If I'm reading [the spec](https://publications.pages.asam.net/standards/ASAM_OpenDRIVE/ASAM_OpenDRIVE_Specification/v1.8.1/specification/08_coordinate_systems/08_02_inertial_coordinate_system.html) correctly, these are supposed to be *intrinsic* rotations, i.e. you apply the yaw rotation first, then the pitch rotation is around the *new* Y axis, and finally the roll rotation is around the *new* X axis (vs. *extrinsic* rotations where you always rotate around the global coordinate axes). So I think you need to use `ZYX` rather than `zyx` as the axis specification here (see the [from_euler docs](https://docs.scipy.org/doc/scipy/reference/generated/scipy.spatial.transform.Rotation.from_euler.html#scipy.spatial.transform.Rotation.from_euler)).

It would be good to add a unit test for this function to make sure the output is as expected even in the presence of nonzero heading, pitch, and roll.

src/scenic/formats/opendrive/xodr_parser.py
```diff
@@ -10,13 +10,23 @@
 import numpy as np
 from scipy.integrate import quad, solve_ivp
 from scipy.optimize import brentq
-from shapely.geometry import GeometryCollection, MultiPoint, MultiPolygon, Point, Polygon
-from shapely.ops import snap, unary_union
+from scipy.spatial.transform import Rotation as R
```
dfremont: I don't think it will be obvious to people what `R` is later on in the code. It's only used one time anyway, so I would just keep `Rotation`.

src/scenic/formats/opendrive/xodr_parser.py
```diff
@@ -1226,9 +1288,214 @@ def getEdges(forward):
...
+        r = R.from_euler('zyx', [yaw, pitch, roll], degrees=False)
+        local_vector = np.array([u, v, z_local])
+        rotated_vector = r.apply(local_vector)
+
+        x = x0 + rotated_vector[0]
+        y = y0 + rotated_vector[1]
+        z = z0 + rotated_vector[2]
+        return (x, y, z)
+
+    def create_center_line(self, leftEdge, rightEdge):
```
dfremont: Presumably you meant to use this in `combineSections`, where this code came from? Also since it doesn't use `self` at all, I would make it a helper function rather than a method.

src/scenic/formats/opendrive/xodr_parser.py
```diff
@@ -1226,9 +1288,214 @@ def getEdges(forward):
...
+                z_local = corner_local['z']
+
+                x, y, z = self.uv_to_xyz(
+                    float(cw.s), float(cw.t), float(cw.zOffset),
+                    u, v, z_local,
+                    float(cw.hdg), float(cw.pitch), float(cw.roll),
+                )
+                points.append((x, y))
+
+        crosswalk_polygon = cleanPolygon(Polygon(points), tolerance)
```
dfremont: We've had issues with `cleanPolygon` sometimes being too aggressive, so I'd suggest only applying it if the raw polygon is not valid (see how we use it now in `calc_geometry_for_type`).

src/scenic/formats/opendrive/xodr_parser.py
```diff
@@ -1226,9 +1288,214 @@ def getEdges(forward):
...
+        for (a, b) in edges:
+            dx = b[0] - a[0]
+            dy = b[1] - a[1]
+            distance = math.hypot(dx, dy)
+            lengths.append(distance)
+
+        get_mrr_shortest_indices = np.argsort(lengths)[:2]
+
+        mrr_shortest_edge = edges[get_mrr_shortest_indices[0]]
+        mrr_shortest_opp_edge = edges[get_mrr_shortest_indices[1]]
```
dfremont: This might be a non-opposite edge if the MRR is a square, I think. In that case we have a 50% chance of picking the wrong direction anyway, but we should still make sure we at least pick opposite edges. If `shortest_index` is the index of the shortest edge (breaking ties arbitrarily), I think you can just use `(shortest_index + 2) % 4` as the other index.

src/scenic/formats/opendrive/xodr_parser.py
```diff
@@ -1226,9 +1288,214 @@ def getEdges(forward):
...
+            num = max(len(leftPoints), len(rightPoints))
+            centerPoints = []
+            for d in np.linspace(0, 1, num):
+                l = leftEdge.lineString.interpolate(d, normalized=True)
+                r = rightEdge.lineString.interpolate(d, normalized=True)
+                centerPoints.append(averageVectors(l.coords[0], r.coords[0]))
+        centerline = PolylineRegion(cleanChain(centerPoints))
+        return centerline
+
+    def construct_crosswalk_polys(self, cw, tolerance):
```
dfremont, line 1334: I think it would help to add a comment about the overall approach here, since it won't be obvious for someone reading this code for the first time. It can just be a sentence or two explaining how you pick which edges are the "top" and "bottom" edges and split the chain at those two edges to yield left and right pieces.

src/scenic/formats/opendrive/xodr_parser.py
```diff
@@ -1226,9 +1288,214 @@ def getEdges(forward):
...
+        self.hdg = hdg
+        self.pitch = pitch
+        self.roll = roll
+        self.outlines = outlines
+        self.polygon = None
+        self.centerline = None
+        self.leftEdge = None
+        self.rightEdge = None
+
+    def is_valid(self):
```
dfremont: This doesn't seem to get used anywhere. Also I'm not sure it's the right check, since the spec says the `length` and `width` fields are optional, and if they're missing you use a default value of zero. I think we only need them if there's no `<outline>`, in which case the object is rectangular and `length` and `width` give you the dimensions (unless `radius` is set, but presumably that doesn't happen for crosswalks).

src/scenic/formats/opendrive/xodr_parser.py
```diff
@@ -1933,12 +2266,14 @@ def cyclicOrder(elements, contactStart=None):
                 outgoingLanes=cyclicOrder(allOutgoingLanes, contactStart=True),
                 maneuvers=tuple(allManeuvers),
                 signals=tuple(allSignals),
-                crossings=(),  # TODO add these
+                crossings=tuple(junctionCrossings),
```
dfremont: These are supposed to be ordered by adjacency too, so I think you need to use something like `cyclicOrder(junctionCrossings, contactStart=True)`.

src/scenic/formats/opendrive/xodr_parser.py
```diff
@@ -1970,7 +2305,7 @@ def cyclicOrder(elements, contactStart=None):
                 groups.append(road.backwardLanes)
         lanes = [lane for road in allRoads for lane in road.lanes]
         intersections = tuple(intersections.values())
-        crossings = ()  # TODO add these
+        crossings = [cr for road in allRoads for cr in road.crossings]
```
dfremont: I think this works, because you're keeping the crossings of connecting roads listed in their `crossings` attribute even though the `parent` of the crossings will be the intersection. But in case that changes later (as it might make sense to list crossings under individual roads only if they are *not* part of an intersection), it might be safer to gather crossings from intersections too (making sure there are no duplicates).

dfremont (changes requested):
Looks good overall, thanks Aarav! I've pointed out a bunch of mostly-minor things above. The only other issue I can see is that there are no tests, but if necessary we could leave adding those to the next PR.

src/scenic/formats/opendrive/xodr_parser.py
```diff
@@ -1454,6 +1727,45 @@ def __parse_link(self, link_elem, road, contact):
...
+            width=float(crosswalk_elem.get("width", 0.0)),
+            hdg=float(crosswalk_elem.get("hdg", 0.0)),
+            pitch=float(crosswalk_elem.get("pitch", 0.0)),
+            roll=float(crosswalk_elem.get("roll", 0.0)),
+            outlines=[],
+        )
+
+        for outline_elem in crosswalk_elem.iter("outline"):
+            corners = []
+            for corner_elem in outline_elem.iter("cornerRoad"):
```
dfremont, line 1748: Doesn't look like `construct_crosswalk_polys` handles these: it assumes you only have `cornerLocal` elements.

Eric-Vin:
Going to merge this into an on-repo branch (so it's easier for me to edit) for some final tweaks before we merge it in. Thanks again for your hard work on this @aaravshar!

PR #482, opened 2026-06-25
Minimum Distance Operator

Eric-Vin:
### Description
Adds an operator to compute the minimum distance between two objects, accounting for shape, size, etc... To vote on the name of this operator, please use [this](https://forum.scenic-lang.org/t/naming-of-minimum-distance-operator/95) poll.

### Issue Link
N/A

### Checklist

### Additional Notes
Please do not merge until poll results are finalized and I have updated the code to reflect them.

### DIFF
--- a/docs/reference/operators.rst
+++ b/docs/reference/operators.rst
@@ -35,6 +35,14 @@ distance [from *vector*] to *vector*
 -------------------------------------
 The distance to the given position from ego (or the position provided with the optional from vector)
 
+.. _minimum distance [from {Object}] to {Object}:
+.. _minimum distance from:
+
+minimum distance [from *Object*] to *Object*
+--------------------------------------------
+The minimum distance to the given Object from ego (or the Object provided with the optional from Object). Unlike :ref:`distance from`, this operator takes into account the Objects' shapes, sizes, etc...
+
+
 .. _angle [from {vector}] to {vector}:
 
 angle [from *vector* ] to *vector*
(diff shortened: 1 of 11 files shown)


docs/reference/operators.rst
```diff
@@ -35,6 +35,14 @@ distance [from *vector*] to *vector*
 -------------------------------------
 The distance to the given position from ego (or the position provided with the optional from vector)
 
+.. _minimum distance [from {Object}] to {Object}:
+.. _minimum distance from:
+
+minimum distance [from *Object*] to *Object*
+--------------------------------------------
+The minimum distance to the given Object from ego (or the Object provided with the optional from Object). Unlike :ref:`distance from`, this operator takes into account the Objects' shapes, sizes, etc...
```
dfremont:
```suggestion
The minimum distance to the given Object from ego (or the Object provided with the optional :scenic:`from {Object}`). Unlike :ref:`distance from`, which computes the distance between the :prop:`position` of the objects, this operator takes into account the objects' :prop:`shape` and dimensions.
```

src/scenic/core/object_types.py
```diff
@@ -1168,6 +1168,20 @@ def distanceTo(self, point):
         """The minimal distance from the space this object occupies to a given point"""
         return self.occupiedSpace.distanceTo(point)
 
+    @cached_method
+    def minimumDistanceTo(self, other):
+        """The minimal distance between this object and another."""
+        if not isinstance(other, Object):
+            raise RuntimeError(
```
dfremont:
```suggestion
            raise TypeError(
```

dfremont (changes requested):
Looks good (just one comment), but please add a `.. note` to the Operator Reference section for `distance from` explaining that when passed objects it just computes the distance between their positions (i.e. their centers), and that if you want to find the separation between two objects taking into account their shape and size you should use `minimum distance from`.

Also the `minimum distance from` entry needs a `.. version-added` directive. You could put version 3.2 for now I guess, or maybe something like "unreleased".

dfremont (approved):
Looks good, thanks Eric!

PR #483, opened 2026-06-30
Allow for string and fstring for record names

Kv139:
### Description

Modified grammar and compiler to allow for using fstring and strings as names in record statements.

### Checklist

### DIFF
--- a/src/scenic/syntax/compiler.py
+++ b/src/scenic/syntax/compiler.py
@@ -1364,7 +1364,12 @@ class ScenicToPythonTransformer(Transformer):
         if node.delay:
             elts = [self.visit(node.delay.value), ast.Constant(node.delay.unitStr)]
             kwargs["delay"] = ast.Tuple(elts, loadCtx)
-
+        if node.name:
+            if isinstance(node.name, ast.AST):
+                node.name = self.visit(node.name)
+            else:
+                node.name = ast.Constant(node.name)
+            
         return self.createRequirementLike(
             "record", node.value, node.lineno, node.name, kwargs
         )
@@ -1429,7 +1434,7 @@ class ScenicToPythonTransformer(Transformer):
                     ast.Constant(requirementId),  # requirement ID
                     newBody,  # body
                     ast.Constant(lineno),  # line number
-                    ast.Constant(name),  # requirement name
+                    name if isinstance(name, ast.AST) else ast.constant(name),  # requirement name
                 ],
                 keywords=keywords,
             )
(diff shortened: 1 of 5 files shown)


src/scenic/syntax/compiler.py
```diff
@@ -1422,14 +1422,19 @@ def createRequirementLike(
...
+
         return ast.Expr(
             value=ast.Call(
                 func=ast.Name(functionName, loadCtx),
                 args=[
                     ast.Constant(requirementId),  # requirement ID
                     newBody,  # body
                     ast.Constant(lineno),  # line number
-                    ast.Constant(name),  # requirement name
+                    (name),  # requirement name
```
dfremont, line 1437:
```suggestion
                    name,  # requirement name
```

dfremont (changes requested):
This looks good, but we need a test where the `record` statement is in a loop and the f-string's value varies across iterations (so that we can check whether multiple records with different names are properly created). It would also be nice to have a test with the `record` statement in a modular scenario invoked multiple times, like in your original case, but that will need to wait until #490 is in.

tests/syntax/test_dynamics.py
```diff
@@ -2250,6 +2250,46 @@ def test_record():
     )
 
 
+def test_record_keys():
```
dfremont, line 2253: Let's simplify this test: if you are just looking to see that the correct record name is used, there's no need to have the behavior definition or the termination condition. You could do something as simple as:
```scenic
i = 0
record 42 as f"record_{i}"
```
In general, the simpler the test, the easier it is for someone to read and see what is being tested.

tests/syntax/test_dynamics.py
```diff
@@ -2250,6 +2250,46 @@ def test_record():
...
+        i=0
+        record ego.position as f"ego_position_{i}"
+        terminate when ego.position.x >= 6
+        """
+    )
+    result = sampleResult(scenario, maxSteps=4)
+    assert "ego_position_0" in result.records
+
+
+def test_record_keys_multiple():
```
dfremont, line 2270: You can simplify this test along the lines above. I also think it would be good to have the two records have different values, so that we test the names and values line up properly. You could record the value of a simple expression depending on `i` like `i+1`, for example, then assert that the value is correct in the test.

dfremont (changes requested):
Nearly there, I just have a couple suggestions for the tests (also my first suggestion still hasn't been applied).

PR #484, opened 2026-07-03
Extracting semantic tags, road and lane types & speed limits

aprabou:
### Description
This extension allows for the extraction of semantic tags and speed limits/types from roads and lanes.
The changes are made to the following files:

**roads.py:**
- added distributionFunction speedLimitAt to get the speed limit at coordinate-s along a given range
- speedLimit element depreciated

**xodr_parser.py:**
- Speed parsing/among helpers: 
  - speed_to_mps
  - speed_limit_ranges_from_type_records
  - speed_limit_ranges_from_lane_records
  - effective_speed_limit_ranges
  - speed_limit_for_s_interval
  - assign_speed_limit_from_ranges. 
- When an interval straddles multiple limits, the minimum is used and an OpenDriveWarning is emitted.
- Tag helpers: 
  - merge_scenic_tags
  - lane_scenic_tags
  - _junction_tags
  - assign_semantic_tags
- Propagation helpers: 
  - propagate_speed_limit
  - apply_lane_speed_limits
  - propagate_tags
- Road, Lane, and Junction now capture:
  - type_records
  - lane speed_records
  - junction type_
  - extra_tags
- Road.toScenicRoad wires per-section and per-lane speed limits and tags into the Scenic objects.

- Expanded drivable_lane_types / sidewalk_lane_types / shoulder_lane_types defaults (e.g. slipLane, biking, walking, restricted, curb, none; parking classified as a shoulder).

**test_opendrive.py:**
- New unit tests covering speed-unit conversion, road/lane speed-range parsing, per-section and per-lane effective limits, multi-limit warnings, and tag propagation

### Checklist


src/scenic/domains/driving/roads.py
```diff
@@ -239,9 +239,21 @@ class NetworkElement(_ElementReferencer, PolygonalRegion):
...
     tags: FrozenSet[str] = frozenset()
 
+    @distributionFunction
+    def speedLimitAt(self, s: float) -> Union[float, None]:
+        """Get the speed limit at coordinate *s* along this element, in meters."""
+        if self.speedLimitRanges:
+            for range_start, range_end, speed in self.speedLimitRanges:
+                if range_start <= s < range_end:
+                    return speed
+            return None
```
dfremont: We shouldn't return `None` at the end of the last range. I think returning `speed` here would be fine (so any `s` value at or beyond the end gets clamped to the end), and you can likewise put `s = max(0, s)` at the top to clamp negative values to zero.

src/scenic/formats/opendrive/xodr_parser.py
```diff
@@ -34,6 +34,243 @@ def warn(message):
     warnings.warn(message, OpenDriveWarning, stacklevel=2)
 
 
+def speed_to_mps(speed_elem):
+    """Convert an OpenDRIVE ``<speed>`` element to m/s."""
+    raw = speed_elem.get("max")
+    if raw in ("no limit", "undefined"):
```
dfremont, line 40: Did you actually see these strings in some files? The spec [says](https://publications.pages.asam.net/standards/ASAM_OpenDRIVE/ASAM_OpenDRIVE_Specification/v1.8.1/specification/11_lanes/11_07_lane_properties.html#sec-866ad6d9-a026-4051-9a3a-5f94405a15f7) the value must be a number.

dfremont: Oops, sorry, I see those strings actually are allowed for road speed limits (just not for lanes).

src/scenic/formats/opendrive/xodr_parser.py
```diff
@@ -34,6 +34,243 @@ def warn(message):
...
+    if not records:
+        return []
+
+    sorted_records = sorted(records, key=lambda record: record[0])
+    ranges = []
+    current_speed = None
+
+    for i, (s, speed_mps) in enumerate(sorted_records):
+        if speed_mps is not None:
+            current_speed = speed_mps
```
dfremont: If `speed_mps` is `None` (which I assume could happen if the corresponding `type` element has no `speed` element or has one specifying "no limit"), wouldn't this code mean we inherit the speed limit from the previous range? That doesn't seem right to me: if there is no speed limit for part of the road, we should respect that (and keep `None` as the speed limit there).

src/scenic/formats/opendrive/xodr_parser.py
```diff
@@ -34,6 +34,243 @@ def warn(message):
...
+    """Build ``(s_start, s_end, speed_mps)`` ranges from OpenDRIVE ``<type>`` records."""
+    if not type_records:
+        return []
+
+    sorted_records = sorted(type_records, key=lambda record: record[0])
+    s_speed_records = []
+    current_speed = None
+    for s, _road_type, speed_mps in sorted_records:
+        if speed_mps is not None:
+            current_speed = speed_mps
```
dfremont: See above comment about respecting `None`.

src/scenic/formats/opendrive/xodr_parser.py
```diff
@@ -34,6 +34,243 @@ def warn(message):
...
+
+    return ranges
+
+
+def speed_limit_ranges_from_type_records(type_records, road_length):
+    """Build ``(s_start, s_end, speed_mps)`` ranges from OpenDRIVE ``<type>`` records."""
+    if not type_records:
+        return []
+
+    sorted_records = sorted(type_records, key=lambda record: record[0])
```
dfremont, line 85: This list will get sorted again in the helper function -- I'd just sort it here and make the helper function assume the input is already sorted.

src/scenic/formats/opendrive/xodr_parser.py
```diff
@@ -34,6 +34,243 @@ def warn(message):
...
+
+    section_end = section_s0 + section_length
+    road_in_section = _clip_ranges_to_interval(road_ranges, section_s0, section_end)
+    return [
+        (range_start - section_s0, range_end - section_s0, speed)
+        for range_start, range_end, speed in road_in_section
+    ]
+
+
+def speed_limit_for_s_interval(ranges, s_start, s_end):
```
dfremont: I think the code would be simpler if you had this function just return the set of speed limits (you can call it `speed_limits_for_s_interval`) and move the `min` to the one place where you call this function.

src/scenic/formats/opendrive/xodr_parser.py
```diff
@@ -34,6 +34,243 @@ def warn(message):
...
+
+    if not overlapping:
+        return None, frozenset()
+    return min(overlapping), frozenset(overlapping)
+
+
+def assign_speed_limit_from_ranges(element, ranges, warn_context=None):
+    """Set ``speedLimit`` and ``speedLimitRanges`` from effective speed ranges."""
+    if not ranges:
+        return
```
dfremont, line 155: Unnecessary.

src/scenic/formats/opendrive/xodr_parser.py
```diff
@@ -34,6 +34,243 @@ def warn(message):
...
+def speed_limit_for_s_interval(ranges, s_start, s_end):
+    """Speed limit applying to the interval ``[s_start, s_end)``.
+
+    Returns ``(speed_limit, overlapping_speeds)``, where ``overlapping_speeds`` is
+    the set of every limit in ``ranges`` that overlaps the interval and
+    ``speed_limit`` is their minimum -- the conservative choice when an interval
+    straddles more than one limit. Both are empty/``None`` when nothing overlaps.
+    """
+    if not ranges or s_end <= s_start:
+        return None, frozenset()
```
dfremont: Can this actually happen?

src/scenic/formats/opendrive/xodr_parser.py
```diff
@@ -34,6 +34,243 @@ def warn(message):
...
+    if not ranges:
+        return
+
+    speeds = {speed for _, _, speed in ranges if speed is not None}
+    if not speeds:
+        return
+
+    element.speedLimit = min(speeds)
+    if len(speeds) > 1:
+        element.speedLimitRanges = tuple(ranges)
```
dfremont: I think we should set this attribute unconditionally, not only when there are multiple ranges. (That way consumer code can assume `speedLimitRanges` always has a meaningful value.)

src/scenic/formats/opendrive/xodr_parser.py
```diff
@@ -34,6 +34,243 @@ def warn(message):
...
+            speeds_text = ", ".join(f"{speed:.4g} m/s" for speed in sorted(speeds))
+            warn(
+                f"{warn_context}: spans multiple speed limits {{{speeds_text}}};"
+                f" using minimum {element.speedLimit:.4g} m/s"
+            )
+    else:
+        element.speedLimitRanges = ()
+
+
+def merge_scenic_tags(*tag_sets):
```
dfremont: This is just a union -- I don't think it needs a special helper function and unit tests. You can just write `frozenset(type_tags | self.extra_tags)`.

src/scenic/formats/opendrive/xodr_parser.py
```diff
@@ -34,6 +34,243 @@ def warn(message):
...
+    return frozenset({lane_type})
+
+
+def _junction_tags(junction):
+    """Tags taken directly from a junction's OpenDRIVE ``type`` and ``name``."""
+    tags = set()
+    if junction.type_ and junction.type_ != "default":
+        tags.add(junction.type_)
+    if junction.name:
+        tags.add(junction.name)
```
dfremont: This should go into the `name` attribute, not the `tags`.

src/scenic/formats/opendrive/xodr_parser.py
```diff
@@ -34,6 +34,243 @@ def warn(message):
...
+    A road belonging to a junction (``road.junction`` equals the junction id)
+    inherits that junction's semantic tags; all other roads get no extra tags.
+    """
+    extra = {road_id: set() for road_id in road_map.roads}
+
+    for jid, junction in road_map.junctions.items():
+        tags = _junction_tags(junction)
+        if not tags:
+            continue
+        for road_id, road in road_map.roads.items():
```
dfremont: This nested loop will take quadratic time. I'd compute the tags for each junction first (might as well just do it in the `Junction` initializer), then loop over roads and pull out the tags from the appropriate junction.

src/scenic/formats/opendrive/xodr_parser.py
```diff
@@ -34,6 +34,243 @@ def warn(message):
...
+)
+_ROAD_PROPAGATION_TYPES = _ROAD_LEVEL_PROPAGATION_TYPES + _LANE_PROPAGATION_TYPES
+
+
+def propagate_speed_limit(speed_limit, elements):
+    """Copy a road speed limit down to lane groups, sections, and lanes."""
+    if speed_limit is None:
+        return
+    for element in elements:
+        if isinstance(element, _ROAD_PROPAGATION_TYPES):
```
dfremont: I think iterating over a bunch of elements and only taking the ones of the correct type is roundabout and confusing. Where this function is used, you know exactly which sub-elements to use (e.g. `road.laneGroups` and `road.sections`), so I would propagate to those directly.

src/scenic/formats/opendrive/xodr_parser.py
```diff
@@ -34,6 +34,243 @@ def warn(message):
...
+            if element.speedLimit is None:
+                element.speedLimit = speed_limit
+
+
+def apply_lane_speed_limits(elements):
```
dfremont: It should be the `min` over the sections. Also I don't think we need a helper function for this: just do it directly when creating the `Lane`.

src/scenic/formats/opendrive/xodr_parser.py
```diff
@@ -34,6 +34,243 @@ def warn(message):
...
+        section_limits = [
+            section.speedLimit
+            for section in element.sections
+            if section.speedLimit is not None
+        ]
+        if section_limits:
+            element.speedLimit = max(section_limits)
+
+
+def propagate_tags(tags, elements):
```
dfremont: As above, let's do this directly with the correct elements and not use a helper function.

src/scenic/formats/opendrive/xodr_parser.py
```diff
@@ -286,6 +523,8 @@ class Lane:
     def __init__(self, id_, type_, pred=None, succ=None):
         self.id_ = id_
         self.width = []  # List of tuples (Poly3, int) for width and s-offset.
+        self.speed_records = []  # [(sOffset, speed_mps), ...] from <speed> elements.
+        self.speed_limit = None  # First record's speed; kept for backward compatibility.
```
dfremont: There are no backwards-compatibility concerns since this is a new attribute. I think you can drop it. (Unlike `NetworkElement.speedLimit`, which we could deprecate but shouldn't remove immediately.)

src/scenic/formats/opendrive/xodr_parser.py
```diff
@@ -610,6 +853,8 @@ def calc_geometry_for_type(self, lane_types, num, tolerance, calc_gap=False):
                                 prev_id = id_ - 1
                             else:
                                 prev_id = id_ + 1
+                            if prev_id not in offsets:
+                                continue
```
dfremont: What's the reason for this change? This condition should never be true.

src/scenic/formats/opendrive/xodr_parser.py
```diff
@@ -767,15 +1012,50 @@ def calculate_geometry(
...
+
+        # collect speed ranges from type records and total road length
+        speed_ranges = speed_limit_ranges_from_type_records(
+            self.type_records, self.length
+        )
+
+        type_tags = frozenset(
+            open_drive_type
+            for _, open_drive_type, _ in self.type_records
+            if open_drive_type
```
dfremont: How can `open_drive_type` be falsy?

src/scenic/formats/opendrive/xodr_parser.py
```diff
@@ -826,6 +1107,28 @@ def toScenicRoad(self, tolerance):
...
+                )
+                if effective_ranges:
+                    assign_speed_limit_from_ranges(
+                        lane_section,
+                        effective_ranges,
+                        warn_context=(
+                            f"road {self.id_} lane {id_} section s=[{s_start},{s_end})"
+                        ),
+                    )
+                else:
```
dfremont: It's not obvious to me how we get into this case. Do we really need two cases, or is there a way to make it uniform?

src/scenic/formats/opendrive/xodr_parser.py
```diff
@@ -1264,14 +1581,24 @@ def __init__(
         fill_intersections=True,
         drivable_lane_types=(
             "driving",
+            "stop",
```
dfremont: Why change this one from shoulder to drivable? The spec says it's a shoulder.

src/scenic/formats/opendrive/xodr_parser.py
```diff
@@ -1264,14 +1581,24 @@ def __init__(
...
             "entry",
             "exit",
-            "offRamp",
             "onRamp",
+            "offRamp",
             "connectingRamp",
+            "slipLane",
+            "biking",
```
dfremont: This is a bike lane, not a driving lane.

src/scenic/formats/opendrive/xodr_parser.py
```diff
@@ -1264,14 +1581,24 @@ def __init__(
...
+        ),
+        sidewalk_lane_types=("walking", "sidewalk"),  # sidewalk deprecated
+        shoulder_lane_types=(
+            "shoulder",
+            "border",
+            "restricted",
+            "curb",
```
dfremont: We should not include `curb`, which is an actual (elevated) curb, not a shoulder.

src/scenic/formats/opendrive/xodr_parser.py
```diff
@@ -1264,14 +1581,24 @@ def __init__(
...
+            "parking",
+            "none",
```
dfremont: What's the motivation for including `none`? It isn't necessarily a shoulder.

src/scenic/formats/opendrive/xodr_parser.py
```diff
@@ -1530,6 +1864,14 @@ def parse(self, path):
             else:
                 pred_link = succ_link = None
 
+            # 2.3 Handle type and speed parsing
```
dfremont: It won't be clear what `2.3` refers to. (I'm not sure the comment adds anything anyway.)

tests/formats/opendrive/test_opendrive.py
```diff
@@ -178,3 +192,541 @@ def test_make_curve_rejects_unknown_geometry_type():
...
+          </lane>
+          <lane id="-2" type="onRamp" level="false">
+            <width sOffset="0" a="3.5" b="0" c="0" d="0"/>
+          </lane>"""
+    road_extras = '<type s="0" type="motorway"><speed max="120" unit="km/h"/></type>'
+    path = write_xodr_multi_lane(tmp_path, lanes_xml, road_extras=road_extras)
+    road_map = RoadMap()
+    road_map.parse(path)
+    road_map.calculate_geometry(num=5, calc_intersect=True)
+    road = road_map.toScenicNetwork().roads[0]
```
dfremont: Can you use `parse_scenic_network` (or a variant) to simplify this kind of test?

tests/formats/opendrive/test_opendrive.py
```diff
@@ -178,3 +192,541 @@ def test_make_curve_rejects_unknown_geometry_type():
...
+            <width sOffset="0" a="3.5" b="0" c="0" d="0"/>
+          </lane>"""
+    road_extras = '<type s="0" type="motorway"><speed max="120" unit="km/h"/></type>'
+    path = write_xodr_multi_lane(tmp_path, lanes_xml, road_extras=road_extras)
+    road_map = RoadMap()
+    road_map.parse(path)
+    road_map.calculate_geometry(num=5, calc_intersect=True)
+    internal_road = road_map.roads[7]
+    internal_road.extra_tags = frozenset({"onramp"})
+    road, _ = internal_road.toScenicRoad(road_map.tolerance)
```
dfremont: Rather than modifying internal attributes like this, I think it's clearer to generate the whole `Network` object and then assert things about it.

tests/formats/opendrive/test_opendrive.py
```diff
@@ -178,3 +192,541 @@ def test_make_curve_rejects_unknown_geometry_type():
...
+    )
+    return path
+
+
+def parse_scenic_network(tmp_path, road_extras="", plan_view=DEFAULT_PLAN_VIEW):
+    path = write_xodr_with_type(tmp_path, plan_view, road_extras=road_extras)
+    road_map = RoadMap()
+    road_map.parse(path)
+    road_map.calculate_geometry(num=5, calc_intersect=True)
+    return road_map.toScenicNetwork()
```
dfremont: Until now no tests in this folder used `Network`: tests that do are in `tests/domains/driving` instead. Since you're checking OpenDRIVE-specific things you could keep them in this folder, but maybe make a new file to distinguish them (e.g. `test_speeds.py` or `test_tags.py`).

dfremont (changes requested):
This is definitely headed in the right direction, thanks Ashwin! I think we can simplify the code a lot, though: see the comments above. Here are a few other suggestions:

* Let's add a `Network.speedLimitAt` method to get the speed limit (if any) at a given point. I think it can just call `Network.laneSectionAt` to get the lane section, compute `s` in the same way that the first line of `PolylineRegion.nearestSegmentTo` does (only necessary if there's more than 1 speed limit range), and then look up the right speed limit range.
* I haven't finished reading all the tests yet (I'll wait until you refactor things), but they're pretty verbose: please see if you can simplify the code. One note is that I don't think we need unit tests for things like `propagate_speed_limit`: that's an internal function which could easily get refactored away. I'd rather have integration tests which load a simple map and then assert things about the resulting `Network` object. (Of course, for functions that will be used in many places and are easily testable on their own, unit tests make sense.)

src/scenic/domains/driving/roads.py
aprabou: Modified, addressed in the next push.

src/scenic/formats/opendrive/xodr_parser.py
aprabou: That does make sense. I made the following changes and would like to know if this aligns with dealing with `None` values:
- `_speed_limit_ranges_from_s_speed_records`: removed `current_speed` inheritance. Each record now has its own `speed_mps` which can be `None`. Also added a prepend record starting at 0.0, so coverage always starts at s = 0, if the first record has s > 0.
- `speed_limit_ranges_from_type_records`: no more inheritance across types, allows for `None` speeds.
- `assign_speed_limit_from_ranges`: if `None` gaps exist, it is kept track of rather than discarded.
- `speed_limit_for_s_interval`: interval overlapping only undefined stretches returns `None`.

src/scenic/formats/opendrive/xodr_parser.py
aprabou: 'ranges' can be empty, but 's_end <= s_start' is practically covered in the rest of the code and is just another guardrail. I will take that out.

src/scenic/formats/opendrive/xodr_parser.py
aprabou: The [spec](https://publications.pages.asam.net/standards/ASAM_OpenDRIVE/ASAM_OpenDRIVE_Specification/v1.8.1/specification/16_annexes/enumerations/map_uml_enumerations.html#top-EAID_9692E2F3_4895_4ce6_A84E_FB1297B0B58E:~:text=use%20exit%20instead-,none,is%20still%20present%20in%20case%20the%20(human)%20driver%20leaves%20the%20road.,-offRamp) mentions that while it does not have actual content, it serves for applications to register that OpenDRIVE is still present in case the driver leaves the road. I'm not sure if this would be relevant to Scenic but I wanted to add that in case it was.

src/scenic/formats/opendrive/xodr_parser.py
aprabou: You're right, thanks for catching that.

src/scenic/formats/opendrive/xodr_parser.py
aprabou: In that case, similar to how the code deals with sidewalk and shoulder lane types, should I model biking lanes as a separate `LaneGroup.bikeLane`?

aprabou: note: just ignore for now, this can be a separate PR

src/scenic/formats/opendrive/xodr_parser.py
aprabou: update- no longer including none

src/scenic/formats/opendrive/xodr_parser.py
aprabou: A guard for bad maps, but like some other conditions it is practically unreachable. Removed.

src/scenic/formats/opendrive/xodr_parser.py
aprabou: It can be falsy if the attribute is missing or empty but type is still defined. Only relevant for bad maps.

aprabou: Check is removed in the latest changes.

aprabou:
Thank you again for the detailed comments and proposed changes. A lot of what was done was refactoring / removing unnecessary checks, but here are the major changes:
- Changes dealing with `None` values:
  - `_speed_limit_ranges_from_s_speed_records`: removed `current_speed` inheritance. Each record now has its own `speed_mps` which can be `None`. Also added a prepend record starting at 0.0, so coverage always starts at s = 0, if the first record has s > 0.
  - `speed_limit_ranges_from_type_records`: no more inheritance across types, allows for `None` speeds.
  - `assign_speed_limit_from_ranges`: if `None` gaps exist, it is kept track of rather than discarded.
  - `speed_limit_for_s_interval`: interval overlapping only undefined stretches returns `None`.
- Add `NetworkElement.speedLimitAt(s)` and `Network.speedLimitAt(point)` to look up the limit at an arc-length or world point (via the containing lane section).
- Propagate semantic tags from road types and junction types onto roads / road sections / lane groups; lanes and lane sections get their OpenDRIVE lane `type` as tags.
- Split OpenDRIVE coverage into `test_speeds.py` and `test_tags.py` (shared helpers in `conftest.py`).
- Cover unit conversion, road/lane range parsing, lane override, multi-limit warnings, `None`/no-limit ranges, `Network.speedLimitAt`, and tag propagation on synthetic maps.

src/scenic/domains/driving/roads.py
```diff
@@ -237,11 +237,25 @@ class NetworkElement(_ElementReferencer, PolygonalRegion):
 
     #: Which types of vehicles (car, bicycle, etc.) can be here.
     vehicleTypes: FrozenSet[VehicleType] = frozenset([VehicleType.CAR])
-    #: Optional speed limit, which may be inherited from parent.
+    #: Optional speed limit, which may be inherited from parent. (depreciated)
```
Eric-Vin, line 240:
```suggestion
    #: Optional speed limit, which may be inherited from parent. (deprecated)
```

src/scenic/domains/driving/roads.py
```diff
@@ -237,11 +237,25 @@ class NetworkElement(_ElementReferencer, PolygonalRegion):
...
     speedLimit: Union[float, None] = None
+    #: Optional ``(s_start, s_end, speed_mps)`` ranges along this element's centerline.
+    speedLimitRanges: Tuple[Tuple[float, float, float], ...] = ()
```
Eric-Vin, line 243: Do we need to have `s_end` included in this? It seems redundant if we just require that the first `s_start` is 0.

src/scenic/domains/driving/roads.py
```diff
@@ -237,11 +237,25 @@ class NetworkElement(_ElementReferencer, PolygonalRegion):
...
+    @distributionFunction
+    def speedLimitAt(self, s: float) -> Union[float, None]:
+        """Get the speed limit at coordinate *s* along this element, in meters."""
+        s = max(0, s)
+        if self.speedLimitRanges:
+            for range_start, range_end, speed in self.speedLimitRanges:
+                if range_start <= s < range_end:
+                    return speed  # may be None -> genuine "no limit" stretch
+            # s is at/beyond the final (half-open) range's end: clamp to it.
+            return speed
```
Eric-Vin, line 256: I think this could be simplified if we make the above changes getting rid of `s_end`, and the logic would be a little clearer and more robust to weird edge cases.

src/scenic/domains/driving/roads.py
```diff
@@ -237,11 +237,25 @@ class NetworkElement(_ElementReferencer, PolygonalRegion):
...
+        return self.speedLimit
```
Eric-Vin, line 257: We could also have this be a deprecated API, that is equivalent to `speedLimitRanges = (0, self.speedLimit)`.

src/scenic/domains/driving/roads.py
```diff
@@ -1293,6 +1307,24 @@ def laneSectionAt(self, point: Vectorlike, reject=False) -> Union[LaneSection, N
...
+        Looks up the `LaneSection` containing the point and returns its speed
+        limit there. When the section has multiple ``speedLimitRanges``, the
+        limit is taken from the range covering the point's projected distance
+        along the section centerline.
+        """
+        point = _toVector(point)
+        section = self.laneSectionAt(point, reject=reject)
+        if section is None:
+            return None
+        if len(section.speedLimitRanges) <= 1:
```
Eric-Vin, line 1323: Why not just always project and defer to the section's `speedLimitAt` function? I don't think the computational cost is that significant and it abstracts things nicely.

src/scenic/formats/opendrive/xodr_parser.py
```diff
@@ -767,15 +920,49 @@ def calculate_geometry(
...
+
+        # collect speed ranges from type records and total road length
+        speed_ranges = speed_limit_ranges_from_type_records(
+            self.type_records, self.length
+        )
+
+        type_tags = frozenset(
+            open_drive_type
+            for _, open_drive_type, _ in self.type_records
+            # if open_drive_type
```
dfremont, line 932: ?

src/scenic/formats/opendrive/xodr_parser.py
```diff
@@ -1017,6 +1227,20 @@ def makeShoulder(sections, backward=False):
...
+            section.speedLimit
+            for section in roadSections
+            if section.speedLimit is not None
+        ]
+        road_speed_limit = min(section_speed_limits) if section_speed_limits else None
+        if road_speed_limit is not None:
+            for section in roadSections:
+                if section.speedLimit is None:
+                    section.speedLimit = road_speed_limit
+                for lane_section in section.lanesByOpenDriveID.values():
```
dfremont, line 1240:
```suggestion
                for lane_section in section.lanes:
```

src/scenic/formats/opendrive/xodr_parser.py
```diff
@@ -767,15 +920,49 @@ def calculate_geometry(
...
+            )
         ):
+            s_start = sec.s0
+            s_end = (
+                self.lane_secs[sec_index + 1].s0
+                if sec_index + 1 < len(self.lane_secs)
+                else self.length
+            )
+            overlapping_speeds = speed_limits_for_s_interval(speed_ranges, s_start, s_end)
+            section_speed_limit = min(overlapping_speeds) if overlapping_speeds else None
```
dfremont, line 956:
```suggestion
            section_speed_limit = min(overlapping_speeds, default=None)
```

src/scenic/formats/opendrive/xodr_parser.py
```diff
@@ -1017,6 +1227,20 @@ def makeShoulder(sections, backward=False):
                     adj.append(lane._laneToRight)
                 lane.adjacentLanes = tuple(adj)
 
+        section_speed_limits = [
+            section.speedLimit
+            for section in roadSections
+            if section.speedLimit is not None
+        ]
+        road_speed_limit = min(section_speed_limits) if section_speed_limits else None
```
dfremont, line 1235:
```suggestion
        road_speed_limit = min(section_speed_limits, default=None)
```

src/scenic/formats/opendrive/xodr_parser.py
```diff
@@ -1068,6 +1297,12 @@ def makeShoulder(sections, backward=False):
                         road=None,
                         sections=tuple(sections),
                         successor=successorLane,  # will correct inter-road links later
+                        tags=frozenset().union(*(sec.tags for sec in sections)),
+                        speedLimit=(
+                            min(lane_section_speed_limits)
+                            if lane_section_speed_limits
+                            else None
+                        ),
```
dfremont, line 1305:
```suggestion
                        speedLimit=min(lane_section_speed_limits, default=None),
```

src/scenic/formats/opendrive/xodr_parser.py
```diff
@@ -1192,8 +1430,17 @@ def getEdges(forward):
             sections=roadSections,
             signals=tuple(roadSignals),
             crossings=(),  # TODO add these!
+            speedLimit=road_speed_limit,
+            tags=road_level_tags,
         )
         allElements.append(road)
+        if road_speed_limit is not None:
+            for group in (forwardGroup, backwardGroup):
+                if group is not None and group.speedLimit is None:
```
dfremont, line 1439:
```suggestion
            for group in road.laneGroups:
                if group.speedLimit is None:
```

src/scenic/formats/opendrive/xodr_parser.py
```diff
@@ -1266,12 +1513,19 @@ def __init__(
...
             "onRamp",
+            "offRamp",
             "connectingRamp",
+            "slipLane",
+        ),
+        sidewalk_lane_types=("walking", "sidewalk"),  # sidewalk deprecated
+        shoulder_lane_types=(
+            "shoulder",
+            "border",
+            "restricted",
```
dfremont, line 1525: I don't think we should include `restricted`, since the spec says cars are not supposed to drive there (unlike the other types).

tests/formats/opendrive/conftest.py
```diff
@@ -0,0 +1,145 @@
...
+        )
+    road_map = RoadMap()
+    road_map.parse(path)
+    road_map.calculate_geometry(num=5, calc_intersect=True)
+    if junction_id != "-1":
+        jid = int(junction_id)
+        if jid not in road_map.junctions:
+            road_map.junctions[jid] = Junction(
+                jid, junction_name, junction_type or "default"
+            )
```
dfremont, line 124: I'm not sure manually injecting elements like this (after parsing) is a good idea. Why can't you put the junction you need into the XML?

src/scenic/formats/opendrive/xodr_parser.py
```diff
@@ -34,6 +34,153 @@ def warn(message):
...
+    """Get the speed active at *s* from nonempty, ordered speed ranges."""
+    speed = ranges[0][1]
+    for range_start, range_speed in ranges[1:]:
+        if s < range_start:
+            break
+        speed = range_speed
+    return speed
+
+
+def effective_speed_limit_ranges(road_ranges, lane_ranges, section_s0, section_length):
```
dfremont, line 115: I think the logic will be clearer if this function is inlined (it's only used in one place).

src/scenic/formats/opendrive/xodr_parser.py
```diff
@@ -34,6 +34,153 @@ def warn(message):
...
+    if not type_records:
+        return []
+
+    sorted_records = sorted(type_records, key=lambda record: record[0])
+    # Each <type> keeps its own speed (may be None); no inheritance across types.
+    s_speed_records = [(s, speed_mps) for s, _road_type, speed_mps in sorted_records]
+    return _speed_limit_ranges_from_s_speed_records(s_speed_records, road_length)
+
+
+def speed_limit_ranges_from_lane_records(speed_records, section_length):
```
dfremont, line 93: Ditto for this function.

src/scenic/formats/opendrive/xodr_parser.py
```diff
@@ -34,6 +34,153 @@ def warn(message):
...
+
+    speeds = {_speed_limit_at(ranges, s_start)}
+    speeds.update(
+        speed for range_start, speed in ranges[1:] if s_start < range_start < s_end
+    )
+    speeds.discard(None)
+    return frozenset(speeds)
+
+
+def assign_speed_limit_from_ranges(element, ranges, warn_context=None):
```
dfremont, line 152: This function too.

src/scenic/formats/opendrive/xodr_parser.py
```diff
@@ -34,6 +34,153 @@ def warn(message):
...
+    element.speedLimitRanges = tuple(ranges)
+    if len(speeds) > 1 and warn_context is not None:
+        speeds_text = ", ".join(f"{speed:.4g} m/s" for speed in sorted(speeds))
+        warn(
+            f"{warn_context}: spans multiple speed limits {{{speeds_text}}};"
+            f" using minimum {element.speedLimit:.4g} m/s"
+        )
+
+
+def assign_semantic_tags(road_map):
```
dfremont, line 169: I'd inline this too: it's a simple function, currently separated by >1000 lines from the single place where it's used. After inlining you can preface it with a 1-2 line comment saying what you're doing.

tests/formats/opendrive/conftest.py
```diff
@@ -0,0 +1,205 @@
...
+    {plan_view}
+    <lanes>
+      <laneOffset s="0.0" a="0.0" b="0.0" c="0.0" d="0.0"/>
+{lanes}
+    </lanes>
+  </road>""",
+    )
+
+
+def write_xodr_junction(tmp_path, connecting_lanes_xml):
```
dfremont, line 98: This big, hard-to-follow function is only used for a single test. I'd replace it with a single XODR file for that specific test.

tests/formats/opendrive/conftest.py
```diff
@@ -0,0 +1,205 @@
...
+            lane_sections_xml=lane_sections_xml,
+        )
+    road_map = RoadMap()
+    road_map.parse(path)
+    road_map.calculate_geometry(num=5, calc_intersect=True)
+    return road_map.toScenicNetwork()
+
+
+def scenic_road(network, road_id=7):
+    return next(road for road in network.allRoads if road.id == road_id)
```
dfremont, line 191: I think you can replace this with just `network.elements[f"road{road_id}"]`. In individual tests it's not so clear what `scenic_road(network)` does, so it might be clearer to just write `network.elements["road7"]`.

tests/formats/opendrive/test_speeds.py
```diff
@@ -0,0 +1,194 @@
...
+        tmp_path,
+        lanes_xml=lane_xml(-1, speeds=((5, 30, "m/s"),)),
+        road_extras='<type s="0" type="town"><speed max="20" unit="m/s"/></type>',
+    )
+    lane_section = scenic_road(network).sections[0].lanes[0]
+    assert lane_section.speedLimitRanges == (
+        (0.0, None),
+        (5.0, 30.0),
+    )
+    assert lane_section.speedLimitAt(2.0) is None
```
dfremont, line 54: Is this the expected behavior according to the spec? I would have thought if no speed limit is specified for part of a lane then the road's speed limit applies. Please double-check.

tests/formats/opendrive/test_tags.py
```diff
@@ -0,0 +1,89 @@
...
+    parse_scenic_network,
+    scenic_road,
+    type_tags_from_road_extras,
+)
+
+
+def test_map_tags_propagate_to_lane_hierarchy(tmp_path):
+    road_extras = '<type s="0" type="motorway"><speed max="120" unit="km/h"/></type>'
+    road = parse_scenic_network(tmp_path, road_extras=road_extras).roads[0]
+    assert road.tags == type_tags_from_road_extras(road_extras)
```
dfremont, line 16: I think using a function here obscures what the expected value actually is. I think writing `frozenset({"motorway"})` would be clearer.

tests/formats/opendrive/test_tags.py
```diff
@@ -0,0 +1,89 @@
...
+    assert_road_tags_propagated_to_groups(road)
+    assert road.sections[0].tags == road.tags
+    assert road.lanes[0].tags == frozenset({"driving"})
```
dfremont, line 19: Shouldn't this be `"motorway"`? Or is there a reason for having lane tags and section tags differ?

dfremont (changes requested):
Looking good; just a few minor questions and comments above.

PR #491, opened 2026-07-19
Link OpenDRIVE signals to intersection maneuvers

aprabou:
### Description
- Add `Maneuver.signal` and `Signal.controlledManeuvers` so each junction movement can reference its controlling signal (and vice versa).
- Wire the link during OpenDRIVE conversion using connecting-road signals first (by lane validity), then incoming-road approach signals.
- Add self-contained integration tests (inline `.xodr` fixtures) covering per-lane validity, approach fallback, T-junction distinct signals, and unsignalized junctions.

### Checklist

### DIFF
--- a/src/scenic/domains/driving/roads.py
+++ b/src/scenic/domains/driving/roads.py
@@ -165,6 +165,8 @@ class Maneuver(_ElementReferencer):
     connectingLane: Union[Lane, None] = None
     #: intersection where the maneuver takes place, if any (`None` for lane mergers)
     intersection: Union[Intersection, None] = None
+    #: traffic signal controlling this maneuver, if any (`None` if uncontrolled or unknown)
+    signal: Optional[Signal] = None
 
     def __attrs_post_init__(self):
         assert self.type is ManeuverType.STRAIGHT or self.connectingLane is not None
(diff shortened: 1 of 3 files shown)


aprabou:
Parse OpenDRIVE 1.8+ `<semantics><priority type="…"/>` and expose it on domain `Signal` via an optional `SignalPriorityType` enum and `priorities` tuple. Classification prefers priorities when present; otherwise falls back to CARLA/OpenDRIVE type codes (`1000001` traffic light, `206` stop, `205` yield). Also pass through `subtype`, propagate priorities through `signalReference`, and warn once when a known legacy type disagrees with `<priority>`.

src/scenic/domains/driving/roads.py
```diff
@@ -147,6 +147,42 @@ def guessTypeFromLanes(
...
 
+@enum.unique
+class SignalPriorityType(enum.Enum):
+    """OpenDRIVE ``e_signals_semantics_priority`` literals (Annex A.7.3).
+
+    Values match the ``type`` attribute of ``<semantics><priority …/>``.
+    Introduced in OpenDRIVE 1.8.0 as a country-agnostic behavior tag.
+    """
+
+    FOUR_WAY = "4way"
```
Eric-Vin, line 158: It would be great to document the meanings of these tags independent of the XODR spec, just in case we need to unify them with another spec or diverge in some way.

aprabou: Will be reflected and resolved in #498 – migrated everything from #491 into #498, now closing #491.

aprabou:
Successful migration of changes from this issue into #498 – closing for now.

PR #505, opened 2026-09-10
Fix paramPoly3 pRange="arcLength" support in OpenDRIVE parser

### LINKED ISSUE #502: paramPoly3 pRange="arcLength" is unsupported
**Title:** paramPoly3 pRange="arcLength" is unsupported

### System Details

1. Python 3.13
2. Scenic 3.1.1 (also present on `main`)
3. Windows 11
4. Simulator: none, parsing only

### Detailed Description

`Network.fromFile()` cannot load a publicly available CC BY 4.0 HD map, so this
should be fully reproducible:

> Scholz, M., Böttcher, O., Schäfer, J. P., Andert, F., iMAR Navigation GmbH (DLR).
> *OpenDRIVE road network dataset of Schwarzer Berg in Brunswick*, v1.0.0, 2025.
> https://doi.org/10.5281/zenodo.15395840 — file `bs-schwarzer-berg.xodr`

The map loads in esmini. 

 `pRange="arcLength"` is unsupported.** All 1,799 `paramPoly3` elements in
this map use it, so the parse stops on the first one:

```python
if p_range and p_range != "normalized":
    # TODO support arcLength
    raise NotImplementedError("unsupported pRange for paramPoly3")
```

This looks cheap to fix exactly rather than approximately. Under `arcLength` the
parameter runs `0..@length` instead of `0..1`, so substituting `p = t*L` yields
the same curve with each coefficient scaled by `L**k`:

```
aU' = aU,  bU' = bU*L,  cU' = cU*L**2,  dU' = dU*L**3
```

I applied that as a preprocessing step and checked the result at five points on
each of the 1,799 elements — the reparameterisation moves no point.

### Steps To Reproduce

1. Download `bs-schwarzer-berg.xodr` from the Zenodo record above.
2. `Network.fromFile("bs-schwarzer-berg.xodr")`
3. NotImplementedError raises immediately.

Souravrajvi0:
## Summary

Fixes #502 by supporting OpenDRIVE `paramPoly3` geometries with `pRange="arcLength"`.

When `pRange` is `arcLength`, the parametric parameter runs from `0` to the geometry `length` (in meters) instead of `0` to `1`. `ParamCubic` already supports arbitrary parameter ranges via `p_range`, so the fix passes the segment length directly instead of raising `NotImplementedError`.

## Changes

- `makeCurve()` in `xodr_parser.py`: handle `pRange="arcLength"` by setting `p_range = length`
- Replace the rejection test with success tests for arc-length curves and equivalence with normalized (scaled-coefficient) form

## Test plan

- [x] `test_make_curve_param_poly3` (existing normalized case)
- [x] `test_make_curve_param_poly3_arc_length` — verifies `p_range == length`
- [x] `test_make_curve_param_poly3_arc_length_matches_normalized` — arc-length and normalized forms produce matching points
- [x] Manually verified all 1,799 `paramPoly3` elements in `bs-schwarzer-berg.xodr` parse successfully via `makeCurve`


src/scenic/formats/opendrive/xodr_parser.py
```diff
@@ -268,9 +268,10 @@ def makeCurve(x0, y0, hdg, length, curve_elem):
             float(curve_elem.get("dV")),
         )
         p_range = curve_elem.get("pRange")
-        if p_range and p_range != "normalized":
-            # TODO support arcLength
-            raise NotImplementedError("unsupported pRange for paramPoly3")
+        if p_range == "arcLength":
```
Eric-Vin: Minor point, but it might be clearer to handle each valid case of `p_range` and then raise the `NotImplementedError` in the else block.

src/scenic/formats/opendrive/xodr_parser.py
Eric-Vin:
```suggestion
        self.p_range = p_range
```

Eric-Vin (changes requested):
Thanks for the fix! Just a few minor comments looking at the code.

It would also be great if you could add the map from the issue to the `assets` folder under `misc` (with the appropriate license and attribution as required by CC BY 4.0), and make sure that it visually looks OK. That way we can have parsing it be part of our test suite, to ensure we maintain compatibility with it.

Souravrajvi0:
Visual check  `bs-schwarzer-berg.xodr` reference lines after `RoadMap.parse()` (all 257 roads, paramPoly3 arcLength):

bs-schwarzer-berg reference lines

Souravrajvi0:
> Thanks for the fix! Just a few minor comments looking at the code.
> 
> It would also be great if you could add the map from the issue to the `assets` folder under `misc` (with the appropriate license and attribution as required by CC BY 4.0), and make sure that it visually looks OK. That way we can have parsing it be part of our test suite, to ensure we maintain compatibility with it.

@Eric-Vin I've pushed the review fixes,  explicit pRange branches with the error in else, and self.p_range = p_range on ParamCubic.

Added [bs-schwarzer-berg.xodr](https://github.com/BerkeleyLearnVerify/Scenic/pull/505#) under assets/maps/misc/ with CC BY 4.0 attribution in the README. Tests cover RoadMap.parse() and all 1,799 paramPoly3 elements through makeCurve.

Dropped a reference-line plot in the comment above  looks like a normal road network to me after parse but you guide me on this better

One thing: Network.fromFile() still blows up later in shoulder triangulation, so I left this map out of the slow full-scenario test_opendrive for now. That's separate from the arcLength fix which  should i dig into it here or follow up if you want? Happy to do in any way you prefer

Eric-Vin:
Hmm, it's a little concerning that the triangulation would fail. Can you share the error message?

Souravrajvi0:
> Hmm, it's a little concerning that the triangulation would fail. Can you share the error message?

@Eric-Vin  the actual error from Network.fromFile("bs-schwarzer-berg.xodr"). It is not triangulation.

```
AssertionError
  File ".../xodr_parser.py", line 767, in calculate_geometry
    _, _, _, _, self.shoulder_region = self.calc_geometry_for_type(
        shoulder_lane_types, num, tolerance, calc_gap=calc_gap
    )
  File ".../xodr_parser.py", line 699, in calc_geometry_for_type
    lane_polys[i], lane_polys[j] = separate(lane_polys[i], lane_polys[j])
  File ".../xodr_parser.py", line 51, in separate
    assert not polyA.overlaps(polyB)
```
What I checked locally:

RoadMap.parse() succeeds (257 roads, 31 junctions). All 1,799 paramPoly3 elements use pRange="arcLength" and go through makeCurve without error.

The new parser tests pass.

The crash is later, While building shoulder polygons (border/parking). First failure is road : 1039600. On that road, driving and sidewalk geometry both succeed; only shoulders hit the assert.

After separate() subtracts the overlap, the leftover is about (4.5e-7)m², so shapely still reports .overlaps() == True and the hard assert fires.
On main (without this PR) the same file never gets that far: it raises NotImplementedError: unsupported pRange for paramPoly3 in makeCurve, so this geometry assert was hidden behind the missing arcLength support.
I left the map out of the slow full-scenario test_opendrive list for that reason. The arcLength parser fix looks separate from this overlap-cleanup brittleness.

Happy to file a follow ,up for separate() if you’d rather not block this pr on it.  you could direct me on this should I dig into that here, or merge the parser fix first?

Eric-Vin:
Assuming that this is something we explicitly call `separate` on, this is a bit worrying. Can you plot the polygons (before/after separation) so we can see what they look like? You can do that using `show()` or `plotPolygon`. Thanks for digging into this!

Souravrajvi0:
Hi , @Eric-Vin here are the `separate()` plots for road `1039600` (shoulders / `lane_polys`).

This is the **first** pair that trips `assert not polyA.overlaps(polyB)` during `Network.fromFile()` on `bs-schwarzer-berg.xodr`. Driving and sidewalk geometry on this road succeed; only the shoulder pass hits it.

**Pair:** `lane_polys[50]` (blue, ~28.25 m²) vs `lane_polys[24]` (orange, ~0.30 m²), after earlier `separate()` steps on the same list.

| | |
|---|---|
| Overlap before | ~0.0117 m² |
| After `difference` + `buffer(-1e-6)` on the larger poly | residual **~4.49×10⁻⁷ m²**, `overlaps()` still `True` |

Zoom first : this is the useful view (`plotPolygon` outlines). The orange polygon crosses the blue edge; the red segment is the leftover sliver after shrink:

Zoom: overlap + residual

Same pair at full lane scale. The orange poly is a small tick on the long blue arc; the residual is a single red pixel (expected at 10⁻⁷ m²):

Before / after / residual

`RoadMap.parse()` still succeeds (257 roads, 31 junctions), this  map still is not in the slow full-scenario `test_opendrive` list for that reason only 

if it's okay I'd like to treat this `separate()` brittleness as a follow up if you would rather not block the `pRange="arcLength"` parser fix on it.

